PR #50, opened 2025-12-25
fix(homarr): add auth subdomain to extra_hosts for OIDC

mairas:
## Summary

- Add `auth.${HALOS_DOMAIN}` to Homarr's `extra_hosts` for container DNS resolution
- Add `NEXTAUTH_URL=https://${HALOS_DOMAIN}` to tell NextAuth the canonical HTTPS URL
- Add `NODE_TLS_REJECT_UNAUTHORIZED=0` for self-signed cert communication

## Problem

OIDC login from Homarr to Authelia was failing due to multiple issues:

1. **Container DNS**: Docker containers cannot resolve mDNS names like `auth.halos.local`
2. **HTTP vs HTTPS**: NextAuth didn't know it was behind an HTTPS proxy, generating `http://` redirect URIs
3. **Self-signed certs**: Node.js rejected Traefik's self-signed certificate

## Solution

1. Add auth subdomain to `extra_hosts` pointing to `host-gateway`
2. Set `NEXTAUTH_URL` to the canonical HTTPS URL
3. Disable TLS verification for internal OIDC communication

## Test plan

- [x] Verified fix on test device (halos.hal)
- [x] Homarr can resolve `auth.halos.local` from within container
- [x] OIDC discovery endpoint accessible
- [x] HTTP to HTTPS redirect works
- [x] OIDC redirect_uri uses HTTPS
- [x] Login flow reaches Authelia login page

## Related Issues

- hatlabs/halos-distro#60 - Device local mDNS resolution (separate issue for /etc/hosts)
- hatlabs/halos-distro#57 - SSO Epic

### DIFF
--- a/apps/homarr/docker-compose.yml
+++ b/apps/homarr/docker-compose.yml
@@ -8,8 +8,15 @@ services:
     command: ["sh", "run.sh"]
     extra_hosts:
       - "host.docker.internal:host-gateway"
+      # Allow Homarr to resolve auth subdomain for OIDC (mDNS doesn't work in containers)
+      - "auth.${HALOS_DOMAIN}:host-gateway"
     environment:
       - SECRET_ENCRYPTION_KEY=${SECRET_ENCRYPTION_KEY}
+      # Tell NextAuth the canonical HTTPS URL (behind Traefik proxy)
+      - NEXTAUTH_URL=https://${HALOS_DOMAIN}
+      # SECURITY: Disables TLS verification for ALL Node.js HTTPS requests.
+      # Required for self-signed Traefik certs. Acceptable for local-only device.
+      - NODE_TLS_REJECT_UNAUTHORIZED=0
       # SSO/OIDC configuration (optional, set AUTH_PROVIDERS=oidc to enable)
       - AUTH_PROVIDERS=${AUTH_PROVIDERS:-}
       - AUTH_OIDC_ISSUER=${AUTH_OIDC_ISSUER}


mairas (commented):
## Review Comments

Thanks for the detailed PR description explaining the problem and solution.

### Issues Found

**1. Inconsistent default value handling (Bug)**

Line 12 uses a default value:
```yaml
- "auth.${HALOS_DOMAIN:-halos.local}:host-gateway"
```

But line 16 doesn't:
```yaml
- NEXTAUTH_URL=https://${HALOS_DOMAIN}
```

If `HALOS_DOMAIN` isn't set, `extra_hosts` will use `auth.halos.local` but `NEXTAUTH_URL` will be `https://` (empty domain). This inconsistency will cause failures.

**Suggested fix**: Either add the default to `NEXTAUTH_URL`:
```yaml
- NEXTAUTH_URL=https://${HALOS_DOMAIN:-halos.local}
```
Or document that `HALOS_DOMAIN` must always be set (though the other line suggests it's optional).

---

**2. `NODE_TLS_REJECT_UNAUTHORIZED=0` affects all HTTPS requests (Security consideration)**

The comment says "Allow self-signed certificates for internal OIDC communication" but this environment variable disables TLS verification for ALL Node.js HTTPS requests in the container, not just OIDC calls.

If Homarr makes any external HTTPS requests (fetching RSS feeds, checking for updates, integrations with external services), they would all bypass certificate verification, making them vulnerable to MITM attacks.

Questions to consider:
- Does Homarr make any external HTTPS requests beyond the local OIDC auth?
- Is there a more targeted way to trust only the local self-signed cert (e.g., NODE_EXTRA_CA_CERTS)?
- Is this an acceptable trade-off for a local-only device?

For a device that's primarily on a local network, this might be an acceptable pragmatic choice, but it should be a conscious decision.

---

**3. Minor: Test plan mentions `halos.hal`**

The test plan says "Verified fix on test device (halos.hal)" - just confirming this is the actual test domain and not a typo for `halos.local`?

---

### What looks good

- The `extra_hosts` approach for mDNS resolution is the right pattern for Docker containers
- `NEXTAUTH_URL` correctly tells NextAuth.js about the HTTPS proxy
- Clear commit message with good context
- Comprehensive test plan with actual verification steps

Let me know your thoughts on the default value inconsistency - that one looks like it could cause real issues if HALOS_DOMAIN isn't set.

mairas:
> This is wrong. `halos.hal` is the local network DNS name while `halos.local` is the mDNS hostname.

Thanks for the correction. So:
- `halos.hal` - resolved via local network DNS
- `halos.local` - resolved via mDNS (Avahi)

The SSO configuration uses `*.halos.local` (mDNS) for all the subdomains.

PR #51, opened 2025-12-25
fix(homarr): add missing OIDC environment variables

mairas:
## Summary

- Add `AUTH_SECRET` env var required by NextAuth.js for OIDC state encryption
- Add `AUTH_OIDC_FORCE_USERINFO` for Authelia v4.39+ compatibility
- Add `AUTH_OIDC_ENABLE_DANGEROUS_ACCOUNT_LINKING` for account linking
- Generate AUTH_SECRET in prestart.sh on first boot

## Problem

OIDC authentication was failing with `error=Configuration` after successful login at Authelia. The Homarr logs showed:
- `OAuthProfileParseError` 
- `InvalidCheck: State could not be decoded`

The root cause was that `AUTH_SECRET` was being generated by prestart.sh and saved to the env file, but was not being passed to the Docker container via docker-compose.yml. Without `AUTH_SECRET`, NextAuth.js cannot encrypt/decrypt the OAuth state parameter, causing the OIDC callback to fail.

## Test plan

- [x] Tested OIDC login flow on halos.hal - now works correctly
- [ ] Verify new installations generate AUTH_SECRET properly
- [ ] Verify existing installations pick up the new env vars after restart


mairas:
## Re-review after refactoring commit

The refactoring addresses the original review feedback well. The `has_nonempty_env_var()` helper is clean and the context comment for `DANGEROUS_ACCOUNT_LINKING` is helpful. CI passes.

### ⚠️ Potential regression for `SECRET_ENCRYPTION_KEY`

The refactoring changed `SECRET_ENCRYPTION_KEY` from:
```bash
if [ -z "$SECRET_ENCRYPTION_KEY" ]; then
```
to:
```bash
if ! has_nonempty_env_var "SECRET_ENCRYPTION_KEY"; then
```

This is a **behavior change**. The new check requires the value to be double-quoted in the file (`VAR="value"`). If someone has an unquoted value (`VAR=value`), the old check would accept it (after sourcing), but the new check would fail and **regenerate the key**.

For `SECRET_ENCRYPTION_KEY`, regeneration causes **data loss** - Homarr uses this to encrypt sensitive data in the database. If the key changes, previously encrypted secrets become unreadable.

**Test case:**
```bash
# File has: SECRET_ENCRYPTION_KEY=abc123  (no quotes)
# OLD: Variable is set after sourcing → no regeneration ✓
# NEW: Regex doesn't match → regenerates → encrypted data lost ✗
```

**Risk assessment:**
- Probability: Low - the script always writes double-quoted values
- Impact: High - encrypted secrets become unreadable
- Scenario: Manual file editing with different quoting

### Recommendation

For `SECRET_ENCRYPTION_KEY` specifically, consider keeping the original `[ -z "$SECRET_ENCRYPTION_KEY" ]` check, or using a combined approach:

```bash
# Regenerate only if truly empty (after sourcing AND in file)
if [ -z "$SECRET_ENCRYPTION_KEY" ]; then
    echo "Generating SECRET_ENCRYPTION_KEY..."
    SECRET_ENCRYPTION_KEY=$(openssl rand -hex 32)
    echo "SECRET_ENCRYPTION_KEY=\"${SECRET_ENCRYPTION_KEY}\"" >> "${ENV_FILE}"
fi
```

For `AUTH_SECRET`, the stricter check is fine - regeneration just requires re-login, not data loss.

### Everything else looks good

- Helper function is well-designed
- Regex pattern handles expected cases correctly
- Context comment for `DANGEROUS_ACCOUNT_LINKING` is clear
- CI passes

---
*Re-review by Claude Code*

PR #53, opened 2025-12-27
feat: add hostname lint check

mairas:
Add pre-commit hook to detect hard-coded hostname references.

Related: hatlabs/halos-distro#61, #52

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #55, opened 2025-12-27
fix(ci): add HALOS_HOSTNAME env var for package builds

mairas:
## Summary

Container-packaging-tools now requires `HALOS_HOSTNAME` for registry URL generation. Set a placeholder value that will be replaced at runtime by homarr-container-adapter.

## Changes

- Added `HALOS_HOSTNAME: halos.local` env var to build action

## Related

Fixes the CI failure from PR #53/#54 merge.

### DIFF
--- a/.github/actions/build-deb/action.yml
+++ b/.github/actions/build-deb/action.yml
@@ -27,4 +27,8 @@ runs:
 
     - name: Build packages
       shell: bash
+      env:
+        # Placeholder hostname for registry URL generation during build.
+        # The actual URL is determined at runtime by homarr-container-adapter.
+        HALOS_HOSTNAME: halos.local
       run: ./tools/build-all.sh


mairas:
## Review: Concern ⚠️

This fix gets CI working but **violates the hostname policy** documented in `halos-distro/docs/HOSTNAME_POLICY.md`.

The policy states:
> Hard-coded references to `halos.local` are NOT allowed except in documentation and halos-pi-gen

Configuration files with hard-coded hostnames are explicitly listed as NOT allowed.

### Suggested Fix

The build script should require `HALOS_HOSTNAME` to be passed explicitly rather than defaulting to a hard-coded value. In the workflow that calls this action, pass it from a repository variable or workflow input:

```yaml
# In the workflow calling this action
env:
  HALOS_HOSTNAME: ${{ vars.HALOS_HOSTNAME }}
```

Then set `HALOS_HOSTNAME` as a repository variable (Settings → Secrets and variables → Actions → Variables).

This aligns with the policy: explicit configuration, no hard-coded defaults.

**Issue created**: Please track follow-up to fix this properly.

PR #57, opened 2025-12-28
feat(traefik): add localhost to mDNS redirect

mairas:
## Summary
- Generate Traefik dynamic configuration that redirects `http://*.localhost/` to `https://*.${HALOS_DOMAIN}/`
- Enables desktop launchers to use simple localhost URLs that work on any HaLOS installation
- Requests go through Traefik for proper TLS and authentication

## How it works
1. Desktop icon opens `http://homarr.localhost/`
2. Browser resolves `*.localhost` to 127.0.0.1 (per RFC 6761)
3. Traefik receives request, redirects to `https://homarr.halos.local/`
4. Regular Traefik routing handles the request with TLS

## Test plan
- [ ] Verify localhost-redirect.yml is generated in /etc/halos/traefik-dynamic.d/
- [ ] Test that http://cockpit.localhost/ redirects to https://cockpit.halos.local/
- [ ] Test that http://homarr.localhost/ redirects to https://homarr.halos.local/


mairas (commented):
## Review: feat(traefik): add localhost to mDNS redirect

### Overview

The implementation looks solid overall. The use of `${HALOS_DOMAIN}` follows the hostname policy correctly, and the redirect logic appears correct.

### Questions/Concerns

**1. `noop@internal` service validity**

```yaml
service: noop@internal
```

I couldn't find `noop@internal` documented as a valid Traefik internal service. The documented internal services are:
- `api@internal`
- `dashboard@internal` 
- `ping@internal`

While the redirect middleware should intercept requests before they reach the service, Traefik may still validate that the service exists when loading the configuration. If `noop@internal` doesn't exist, this could cause the router to fail silently or log errors.

**Question**: Have you verified that `noop@internal` works in Traefik v2/v3? If not, consider using `api@internal` as a fallback target.

**2. Manual testing only**

The test plan is manual-only. Given this script has no automated tests, there's no way to catch regressions. Not blocking for this PR, but worth noting.

### Minor observations

- The regex `^https?://` is slightly broader than needed (only `http://` is possible on the `web` entrypoint), but this is fine for robustness.
- The HostRegexp pattern `[a-z0-9-]+` only matches lowercase. Browsers typically lowercase hostnames before resolving, so this should be fine.

### Suggestions for testing

When testing, verify:
1. Traefik starts without errors after generating the config
2. `curl -v http://homarr.localhost/` returns a 302/307 redirect to `https://homarr.<domain>/`
3. Check Traefik logs for any warnings about the `noop@internal` service

mairas:
## Response to Review

### Regarding `noop@internal`

I've researched this and confirmed that `noop@internal` **is a valid Traefik internal service**. While it's not in the official docs (there's actually an open issue requesting documentation: [traefik/traefik#7291](https://github.com/traefik/traefik/issues/7291)), it's a legitimate built-in service:

- It's commonly used for exactly this pattern: routers with redirect middleware that intercept requests before reaching any backend
- When requests reach it without middleware interception, it returns 418 (I'm a teapot) - but our redirect middleware will always intercept first
- It's the recommended approach to avoid creating dummy services when using redirect middleware

I'll add a code comment explaining this to prevent future confusion.

### Regarding manual testing

Acknowledged. Automated testing for this script would require either:
- A mock Traefik setup in CI
- Integration tests on an actual HaLOS system

Both are beyond the scope of this PR but worth considering for the future.

PR #63, opened 2025-12-30
fix(traefik): add insecure transport for HTTPS backends, remove HALOS_HOSTNAME

mairas:
## Summary

Two related fixes:

### 1. Insecure transport for HTTPS backends
When `routing.backend.scheme` is `https`, Traefik needs to skip TLS certificate verification for backends with self-signed certificates (e.g., OpenCPN with selkies-gstreamer).

- Add `insecure-transport.yml` to dynamic config with `insecureSkipVerify: true`
- Update `generate-routing-labels.sh` to add `serversTransport` label for HTTPS backends
- Update `prestart.sh` to install all dynamic config files from package

### 2. Remove HALOS_HOSTNAME placeholder
- Remove bogus `HALOS_HOSTNAME=build.example.local` from CI
- Add `CONTAINER_TOOLS_PATH` support for local development builds

This is a companion PR to:
- hatlabs/container-packaging-tools#174 (template variables for URLs)
- hatlabs/halos-marine-containers#94 (remove placeholder)

## Test plan

- [x] Built and deployed to test device
- [x] Verified `insecure-transport.yml` is installed by package
- [x] Verified OpenCPN routing labels include `serversTransport: "insecure@file"`
- [x] Verified `https://opencpn.halos.local/` returns HTTP 200

### DIFF
--- a/.github/actions/build-deb/action.yml
+++ b/.github/actions/build-deb/action.yml
@@ -27,9 +27,4 @@ runs:
 
     - name: Build packages
       shell: bash
-      env:
-        # Build-time placeholder for registry URL generation.
-        # Uses RFC 2606 reserved domain to clearly indicate this is not a real hostname.
-        # The actual URL is determined at runtime by homarr-container-adapter.
-        HALOS_HOSTNAME: build.example.local
       run: ./tools/build-all.sh
--- a/apps/traefik/assets/dynamic/insecure-transport.yml
+++ b/apps/traefik/assets/dynamic/insecure-transport.yml
@@ -1,0 +1,11 @@
+# ServersTransport for backends with self-signed certificates
+# Used when routing.backend.scheme is "https"
+#
+# This transport skips TLS certificate verification, which is necessary
+# for backends that use self-signed certificates (e.g., OpenCPN with
+# selkies-gstreamer).
+
+http:
+  serversTransports:
+    insecure:
+      insecureSkipVerify: true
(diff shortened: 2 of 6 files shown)


mairas (commented):
## Review: LGTM ✅

The changes look good. The insecure transport configuration is the correct solution for HTTPS backends with self-signed certificates.

### Strengths

1. **`insecure-transport.yml`** - Well-documented config file. The serversTransport with `insecureSkipVerify: true` is the standard Traefik approach for self-signed backends.

2. **Dynamic config loop in `prestart.sh`** - Good refactoring from hardcoded file handling to a loop over all `*.yml` files. This makes adding future dynamic configs easier and cleaner.

3. **`generate-routing-labels.sh`** - Correctly adds both the scheme and transport labels only for HTTPS backends.

4. **`CONTAINER_TOOLS_PATH` support** - Same useful addition for local development as in the marine-containers PR.

### Minor suggestions

1. In `generate-routing-labels.sh:230-231`, when `scheme_label` or `transport_label` are empty, they produce blank lines in the output. This is cosmetic but could be avoided with conditional output:
```bash
${scheme_label:+${scheme_label}$'\n'}${transport_label:+${transport_label}}
```

2. Same note about unquoted `$mount_opts` in the `run` script as mentioned in halos-marine-containers#94.

### Dependency note

Like the marine-containers PR, this technically depends on container-packaging-tools#174 being merged for the template variable changes. However, CI passes here because the build-all.sh only builds traefik which may not need registry generation.

---

**Merge order**: Safe to merge after container-packaging-tools#174, or could merge now if CI is green.

PR #68, opened 2025-12-30
fix(authelia): override healthcheck for faster Traefik routing

mairas:
## Summary

- Override Authelia image's 60-second healthcheck start_period with 1s
- Reduces routing availability delay from 24+ seconds to ~3 seconds after reboot

## Problem

Traefik v3 won't route to containers until their Docker healthcheck status is "healthy". Authelia's Docker image has a 60-second `start_period` in its healthcheck definition, causing the container to remain in "starting" state for up to a minute after startup.

This resulted in 404 errors when accessing auth.halos.local immediately after system reboot, because Homarr redirects to Authelia for OIDC authentication before Authelia is ready.

## Solution

Override the image's healthcheck in docker-compose.yml with:
- `start_period: 1s` (down from 60s)
- `interval: 2s` (more frequent checks)
- `retries: 3` (reasonable retry count)

## Test plan

- [x] Reboot test device
- [x] Verify auth.halos.local responds within 5 seconds of Traefik starting
- [x] Verify Authelia healthcheck transitions from "starting" to "healthy" quickly

### DIFF
--- a/apps/authelia/docker-compose.yml
+++ b/apps/authelia/docker-compose.yml
@@ -9,6 +9,13 @@ services:
       - TZ=${TZ:-UTC}
       # Load base configuration and OIDC clients (merged by Authelia)
       - X_AUTHELIA_CONFIG=/config/configuration.yml,/config/oidc-clients.yml
+    # Override image's 60s start_period - Traefik won't route until healthy
+    healthcheck:
+      test: ["CMD-SHELL", "/app/healthcheck.sh"]
+      interval: 2s
+      timeout: 3s
+      start_period: 1s
+      retries: 3
     networks:
       - proxy
     labels:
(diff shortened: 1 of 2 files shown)


mairas:
## Code Review

Clean fix for the Traefik routing delay issue.

**Review notes:**

1. **Timing looks reasonable**: With `start_period: 1s`, `interval: 2s`, `retries: 3`, Authelia has about 7 seconds total before being marked unhealthy. Since Authelia typically starts quickly, this should be sufficient.

2. **Uses upstream healthcheck script**: Good that you're using the same `/app/healthcheck.sh` from the image rather than rolling a custom check.

3. **Consider documenting**: This is a good pattern for other slow-starting containers behind Traefik. Might be worth noting in the repo's general documentation.

CI passes. LGTM ✓

Issue #21, opened 2025-12-19
feat(sso): Update Homarr for Traefik and OIDC


mairas:
## Description

Modify `apps/homarr/` to work with Traefik reverse proxy and Authelia OIDC authentication.

## Implementation Requirements

**MANDATORY**: Follow `docs/IMPLEMENTATION_CHECKLIST.md` during implementation.

This includes:
- Exploration phase WITHOUT coding
- Planning with "think hard"
- Test-Driven Development (write tests first)
- Verification with subagents
- References to SSO_SPEC.md and SSO_ARCHITECTURE.md

## Implementation Details

Modify existing files:

**docker-compose.yml changes:**
- Remove `ports: - "80:7575"` (Traefik handles routing)
- Add Traefik labels:
  - `traefik.enable=true`
  - `traefik.http.routers.homarr.rule=Host(\`halos.local\`)`
  - `traefik.http.routers.homarr.entrypoints=web`
  - `traefik.http.services.homarr.loadbalancer.server.port=7575`
- Add mDNS label: `halos.subdomain=` (empty for root domain, or omit)
- Add network configuration:
  - Join `halos-proxy-network` with `external: true`

**config.yml changes:**
Add SSO configuration group with fields:
- `AUTH_PROVIDERS` - Set to "oidc" to enable SSO
- `AUTH_OIDC_ISSUER` - Authelia URL (default: http://auth.halos.local)
- `AUTH_OIDC_CLIENT_ID` - OIDC client ID (default: homarr)
- `AUTH_OIDC_CLIENT_SECRET` - OIDC client secret
- `AUTH_OIDC_CLIENT_NAME` - Display name (default: HaLOS)
- `AUTH_OIDC_SCOPE_OVERWRITE` - Scopes (default: openid profile email groups)
- `AUTH_LOGOUT_REDIRECT_URL` - Post-logout URL

**metadata.yaml changes:**
- Add `recommends: [@authelia]` for optional SSO

## Test Scenarios

- Homarr accessible via Traefik at halos.local
- Direct port access (7575) no longer works
- OIDC login redirects to Authelia when configured
- Callback completes and session established
- Dashboard shows logged-in user information

## Acceptance Criteria

- [ ] Homarr accessible only through Traefik
- [ ] OIDC flow works end-to-end with Authelia
- [ ] Dashboard shows logged-in user
- [ ] Package still builds with container-packaging-tools
- [ ] Matches SSO_SPEC.md requirements
[...]

PR #77, opened 2026-01-02
refactor: consolidate into single unified Debian package

mairas:
## Summary

- Refactors halos-core-containers from 3 separate container packages (traefik, authelia, homarr) into a single unified package with manual Debian build system
- Moves from apps/* multi-package structure to single-package root layout
- Adds debian/ packaging with proper systemd service integration
- Uses docker-compose depends_on for container orchestration instead of separate services
- Adds unified prestart.sh that configures all services (TLS certs, OIDC, database seeding)
- Generates Traefik routing dynamically for auth.halos.local (critical for OIDC)
- Proper OIDC client setup with secret hashing via Authelia CLI
- Database seeding from halos-homarr-branding seed db

## Test plan

- [ ] Build package with `./run build-deb`
- [ ] Install on test device
- [ ] Verify all containers start (traefik, authelia, homarr)
- [ ] Test OIDC login flow (click "Login with HaLOS", authenticate with Authelia)
- [ ] Verify homarr-container-adapter can access Homarr API

## Dependencies

This PR requires corresponding changes in:
- halos-metapackages PR #21 (depend on halos-core-containers)
- halos-homarr-branding PR #21 (depend on halos-core-containers)


mairas:
## PR Review: refactor: consolidate into single unified Debian package

### Summary

This is a significant architectural change that consolidates three separate container packages (traefik, authelia, homarr) into a single unified package with proper Docker Compose orchestration and dependency management. The implementation is well-structured and addresses the tight coupling between these services. However, I've identified a few issues that should be addressed.

### Issues Found

#### Major: Inconsistency between metadata.yaml and debian/control for halos-homarr-branding

**Location:** `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/metadata.yaml:32-34` and `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/debian/control:11`

**Issue:** The PR description states that halos-homarr-branding was changed from `Recommends` to `Depends`, and `debian/control` correctly shows it in the `Depends` line. However, `metadata.yaml` still lists it under `recommends:` (lines 32-34). This inconsistency could cause confusion if tooling ever uses metadata.yaml for package generation.

**Suggestion:** Update `metadata.yaml` to move `halos-homarr-branding` from `recommends` to `depends` for consistency, or add a comment explaining that `debian/control` is the source of truth.

#### Minor: debian/rules misses installing traefik/dynamic subdirectory files

**Location:** `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/debian/rules:30-39`

**Issue:** The `for dir in` loop installs files from `assets/$dir/*` but does not handle subdirectories like `assets/traefik/dynamic/*.yml`. The three files (`authelia.yml`, `insecure-transport.yml`, `redirects.yml`) in `assets/traefik/dynamic/` appear to NOT be installed by this rule.

**Suggestion:** The loop should either recursively copy subdirectories or explicitly handle the `traefik/dynamic` subdirectory:
```makefile
# Option 1: Use cp -r for directory preservation
cp -r assets/traefik debian/$(PKG_NAME)$(LIB_DIR)/assets/

# Option 2: Explicit install for dynamic configs
for f in assets/traefik/dynamic/*.yml; do \
    install -D -m 644 "$$f" "debian/$(PKG_NAME)$(LIB_DIR)/$$f"; \
done
```

This is actually a **critical** issue because `prestart.sh` relies on these files at lines 152-165 and copies them to `/etc/halos/traefik-dynamic.d/`. Without them, the authelia middleware and other dynamic configs won't be available.

#### Minor: Potential race condition in OIDC client secret sync

**Location:** `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/prestart.sh:251-257`

**Issue:** The logic to always update `AUTH_OIDC_CLIENT_SECRET` when it already exists uses `sed -i`, but if the env file doesn't exist, the grep will fail silently and the sed will never run. The file creation only happens on first writes, so if the file is missing entirely, secrets won't be persisted.

**Suggestion:** Ensure `${ENV_FILE}` exists before attempting to update it, or touch it early in the setup phase. The current code creates it implicitly by appending to it, but this depends on order of operations.

#### Minor: Code duplication between prestart.sh and reload-oidc-clients

**Location:** `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/prestart.sh:297-408` and `/Users/mairas/w/hatlabs/halos/halos-distro/halos-core-containers/assets/reload-oidc-clients:31-139`

**Issue:** The `hash_client_secret()` function and `merge_oidc_clients` logic are duplicated between these two files. This could lead to drift if one is updated but not the other.

**Suggestion:** Consider refactoring to have `prestart.sh` call `reload-oidc-clients` (or a shared library) rather than duplicating the logic. Alternatively, add a comment noting the intentional duplication and the need to keep them in sync.

### Questions

1. The docker-compose.yml has Traefik depending on Authelia being healthy (`condition: service_healthy`), which makes sense for ForwardAuth. But the old code in `apps/traefik/docker-compose.yml` had a `wait-for-authelia.sh` script as entrypoint. Has this been removed? I see the volume mount is gone in the new docker-compose.yml, but the wait script still exists in assets. Is this intentional?

2. The network changed from `external: true` to `driver: bridge`. Does this mean the network is now created by this compose file? If so, what happens if other container apps that depend on this network start before halos-core-containers?

### Positive Notes

- Excellent use of `depends_on` with health checks to ensure proper startup ordering
- Clean separation of concerns with dedicated prestart script sections for each service
- Good documentation comments throughout the scripts explaining the "why"
- Smart use of Docker labels for Traefik routing instead of generating dynamic config files
- The OIDC client secret sync fix (always updating to match the file) addresses a real idempotency issue
- Proper file permissions (chmod 600 for secrets) throughout

### Verdict

**Request changes**

The missing installation of `assets/traefik/dynamic/*.yml` files in `debian/rules` is a critical issue that will break the ForwardAuth middleware and other Traefik dynamic configurations. This must be fixed before the PR can be merged.

**CI Status:** All passing (test check succeeded)

---
*Review by Claude Code*

PR #79, opened 2026-01-02
feat(homarr): use OIDC auto-login instead of ForwardAuth

mairas:
## Summary

- Replace Traefik ForwardAuth middleware with Homarr's native OIDC auto-login
- Add `AUTH_OIDC_AUTO_LOGIN=true` to auto-redirect users to Authelia on the login page
- Remove `authelia@file` ForwardAuth middleware from Homarr router
- Fix hard-coded hostname references in comments to use `${HALOS_DOMAIN}` variable

## Why

Previously, users experienced a two-step authentication:
1. ForwardAuth redirects to Authelia → user logs in → returns to Homarr
2. User sees Homarr but is NOT logged in → must click "Login with HaLOS" button

Now with OIDC auto-login:
1. User visits halos.local → Homarr auto-redirects to Authelia
2. User logs in → returns to Homarr **already logged in**

This provides a seamless single sign-on experience.

## Test plan

- [ ] Open incognito browser and navigate to https://halos.local
- [ ] Verify auto-redirect to auth.halos.local
- [ ] Log in with Authelia credentials
- [ ] Verify return to Homarr **already logged in** (no extra login step needed)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,7 +58,7 @@ services:
       - traefik.http.routers.authelia-secure.entrypoints=websecure
       - traefik.http.routers.authelia-secure.tls=true
       - traefik.http.services.authelia.loadbalancer.server.port=9091
-      # mDNS subdomain - halos-mdns-publisher uses this to publish auth.halos.local
+      # mDNS subdomain - halos-mdns-publisher publishes auth.${HALOS_DOMAIN}
       - halos.subdomain=auth
     logging:
       driver: journald
@@ -96,6 +96,8 @@ services:
       - AUTH_OIDC_FORCE_USERINFO=${AUTH_OIDC_FORCE_USERINFO:-}
       - AUTH_OIDC_ENABLE_DANGEROUS_ACCOUNT_LINKING=${AUTH_OIDC_ENABLE_DANGEROUS_ACCOUNT_LINKING:-}
       - AUTH_LOGOUT_REDIRECT_URL=${AUTH_LOGOUT_REDIRECT_URL}
+      # Auto-redirect to OIDC provider on login page for seamless SSO
+      - AUTH_OIDC_AUTO_LOGIN=true
     ports:
       # Expose to localhost only for homarr-container-adapter
       - "127.0.0.1:7575:7575"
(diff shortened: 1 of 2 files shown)


mairas:
## PR Review: feat(homarr): use OIDC auto-login instead of ForwardAuth

### Summary

This is a well-reasoned change that improves UX by eliminating the double-authentication problem. The implementation is clean and the security model remains sound.

### Security Analysis

**Regarding ForwardAuth removal**: This change is security-neutral, not a reduction in security:

1. **ForwardAuth pattern**: Routes through Authelia first, then if authenticated, passes through to app. App sees headers like `Remote-User`.

2. **OIDC pattern**: App initiates OAuth flow to Authelia. User authenticates with Authelia. App receives tokens and establishes session.

Both patterns:
- Require authentication via Authelia
- Result in the user being authenticated

The key difference is **who initiates the auth check**:
- ForwardAuth: Traefik intercepts and redirects before the request reaches Homarr
- OIDC: Homarr redirects the user to Authelia itself

In this case, `AUTH_OIDC_AUTO_LOGIN=true` makes Homarr immediately redirect unauthenticated users to Authelia, achieving the same security posture as ForwardAuth but with proper OIDC session establishment.

**Access control**: Authelia still enforces `one_factor` policy (configured in `configuration.yml.template` line 47) and the OIDC client uses `authorization_policy: one_factor` (generated by prestart.sh).

### Code Review

#### Minor: Comment clarity in `docker-compose.yml:121`

**Location:** `docker-compose.yml:121`
**Observation:** The comment says "no ForwardAuth needed" which is accurate, but could be clearer about why this is still secure.
**Suggestion:** Consider: `# HTTPS router - Homarr handles auth via OIDC auto-login (equivalent security to ForwardAuth)`

This is optional - the current comment is acceptable.

### Questions

1. **Unauthenticated access to Homarr**: With ForwardAuth removed, is there any path where Homarr can be accessed without authentication? The auto-login should immediately redirect, but are there any API endpoints or assets that might be exposed?

   Looking at the code, Homarr's `AUTH_OIDC_AUTO_LOGIN=true` should handle the login page redirect, but I don't have visibility into Homarr's internal behavior for other routes. Worth verifying in manual testing that no endpoints are accessible without auth.

2. **Local-only port binding**: I see `127.0.0.1:7575:7575` for homarr-container-adapter access. This is good - it prevents direct access bypassing Traefik.

### Positive Notes

- Clean, minimal change that solves a real UX problem
- Hostname variable fixes in comments are a good cleanup
- The PR description clearly explains the "before and after" user experience
- Test plan is appropriate for the change

### Verdict

**Approve**

**CI Status:** All checks passing (1/1: test - SUCCESS)

The change is well-implemented and addresses a genuine UX issue. The security model is maintained through OIDC with auto-login. The manual testing in the test plan should verify no unintended exposure.

PR #99, opened 2026-02-18
Add init: true to all services for graceful shutdown

mairas:
## Summary

- Add `init: true` to all 4 services (traefik, authelia, authelia-valkey, homarr) so Docker uses tini as PID 1 for proper SIGTERM forwarding
- Change homarr command from `sh` to `bash` so run.sh's trap handler fires on SIGTERM (sh swallows signals even with tini)

Without init, containers using shell scripts as PID 1 ignore SIGTERM and only stop after Docker's 10-second SIGKILL timeout, adding ~40 seconds to every shutdown.

## Test plan

- [ ] Deploy to halos.local and verify all containers stop in <500ms (was ~10s each)
- [ ] Verify homarr starts and runs correctly with `bash` instead of `sh`
- [ ] Verify homarr graceful shutdown (run.sh trap handler fires)


mairas:
## Code Review

**Approved.** Clean, well-scoped PR.

- `init: true` correctly added to all 4 services (traefik, authelia, authelia-valkey, homarr) for proper SIGTERM forwarding via tini
- `sh` → `bash` change for homarr is correct: upstream `run.sh` uses bash traps (`trap terminate TERM INT`) that don't work under dash/busybox sh. Confirmed `/bin/bash` is present in the homarr image.
- Version bump 0.2.4 → 0.2.5 is present and correct

No issues found.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-03-02
fix(routing): log to stderr to prevent YAML corruption on first app install

mairas:
## Summary

- Fix `configure-container-routing` log function to write to stderr instead of stdout

## Motivation

When a container app is installed for the first time, `get_or_assign_port()` assigns a new port and calls `log()` to report it. Since `log()` writes to stdout and the function's return value is captured via command substitution, the log message gets concatenated with the port number:

```
external_port="[configure-container-routing] Assigned port 4431 to avnav\n4431"
```

This corrupted value is interpolated into the docker-compose override YAML template, causing:

```
yaml: line 6: could not find expected ':'
```

The service then crashes and restarts. On retry, the port is already in the registry so the log line is skipped and it works — but every first-time app install fails once.

Fix: `>&2` on the `log()` echo.

## Test plan

- [ ] Install a new container app on a test device (one not previously installed)
- [ ] Verify it starts on the first attempt without YAML errors
- [ ] Verify existing apps still start correctly after upgrade

### DIFF
--- a/assets/configure-container-routing
+++ b/assets/configure-container-routing
@@ -36,7 +36,7 @@ if [ -z "${HALOS_DOMAIN}" ]; then
 fi
 
 log() {
-    echo "[configure-container-routing] $1"
+    echo "[configure-container-routing] $1" >&2
 }
 
 # Parse YAML value - simple grep-based parser for our known format
--- a/assets/traefik/generate-routing-labels.sh
+++ b/assets/traefik/generate-routing-labels.sh
@@ -27,7 +27,7 @@ if [ -z "${HALOS_DOMAIN}" ]; then
 fi
 
 log() {
-    echo "[generate-routing-labels] $1"
+    echo "[generate-routing-labels] $1" >&2
 }
 
 # Main: Process all routing files using configure-container-routing


mairas (commented):
## Adversarial Review

### Gut Reaction

Correct one-line fix for a real bug. Well-diagnosed, well-described PR.

### Critical Issues

None.

### Major Concerns

**1. `generate-routing-labels.sh` has the same stdout `log()` pattern**
- **Location**: `assets/traefik/generate-routing-labels.sh:29-31`
- **Problem**: This script defines its own `log()` that writes to stdout (no `>&2`). It is not currently affected because it never captures function output via command substitution -- it calls `configure-container-routing` as a subprocess (line 50), not via `$(...)`. However, it creates an inconsistency: the same conceptual function (`log`) behaves differently in two sibling scripts. If someone later adds command substitution to this script, they will hit the same bug.
- **Suggestion**: Change to `>&2` in `generate-routing-labels.sh` for consistency, or at minimum add a comment explaining the convention.

**2. Stale built artifact in `debian/` still has the old `log()` without `>&2`**
- **Location**: `debian/halos-core-containers/usr/bin/configure-container-routing:30`
- **Problem**: The `debian/` directory contains a checked-in copy of the script that is significantly out of date (no port registry, no path redirects, no `write_external_port`). It still has `log()` writing to stdout. If anyone installs from this built artifact rather than rebuilding, they get the buggy version.
- **Risk**: Confusing at minimum. Could mask whether the fix is actually deployed.
- **Suggestion**: Either regenerate the `debian/` build artifacts as part of this PR, or (better) stop checking built artifacts into version control and generate them in CI only. But this is pre-existing tech debt, not something this PR introduced -- so not a blocker.

**3. No test for the fixed bug**
- **Problem**: The test suite (`run-tests` action) validates YAML syntax of metadata and runs asset-server security tests, but has zero tests for `configure-container-routing`. The exact scenario described in this PR (first-time port assignment corrupting YAML) is not tested and could regress.
- **Risk**: Future changes to `get_or_assign_port()` or `log()` could reintroduce the same class of bug with no automated detection.
- **Suggestion**: A minimal shell test that runs `get_or_assign_port` with a fresh (empty) registry and asserts the captured output is a bare integer would catch this entire class of bug. Not necessarily a blocker for this one-line fix, but worth tracking.

### Minor Issues

**1. `yaml_get` functions use `grep` with unescaped user-derived patterns**
- **Location**: `assets/configure-container-routing:43-61`
- **Problem**: `app_id` values are used directly in `grep` and `sed` regex patterns. An `app_id` containing regex metacharacters (`.`, `*`, `[`, etc.) would cause incorrect parsing or errors. This is pre-existing and low-risk since `app_id` values come from controlled routing files, not user input.

### Questions

1. The `debian/` directory appears to contain a substantially older version of this script (no port registry, no path redirects, different Authelia endpoint path). Is this directory supposed to be kept in sync with `assets/`, or is it generated during build? If the latter, should it be `.gitignore`d?

### What's Actually Good

- The fix is exactly one line, targeting the root cause precisely.
- The PR description is excellent -- it explains the failure mechanism, why retries succeed, and the exact corruption that occurs.
- The commit message body is thorough and explains the "why" well.
- Single atomic commit for a single logical change.
- CI passes.

Issue #118, opened 2026-04-13
Add Docker healthchecks to Traefik and Homarr containers


mairas:
## Context

Currently only Authelia and Valkey have Docker healthchecks in the core containers. Traefik and Homarr are missing them. Without healthchecks, there's no automated way to detect if these services become unresponsive.

### Traefik
- `wget` is available in the image
- Traefik has a built-in ping endpoint that can be enabled via `--ping`
- Suggested: `wget -q --spider http://localhost:8082/ping || exit 1` (after enabling the ping entrypoint)
- Alternatively: `wget -q --spider http://localhost:80/ || exit 1`

### Homarr
- `wget` is available in the image
- Listening on port 7575
- Suggested: `wget -q --spider http://localhost:7575/ || exit 1`

## Additional work

A container health watchdog mechanism is also needed since Docker's `restart: unless-stopped` does not auto-restart unhealthy containers. See halos-org/halos-marine-containers for the parallel issue.

PR #119, opened 2026-04-13
Add autoheal container for automatic unhealthy container recovery

mairas:
## Summary

- Adds `willfarrell/autoheal` container that monitors all Docker containers with healthchecks
- When a container becomes unhealthy, autoheal automatically restarts it
- Checks every 30 seconds, with a 120-second grace period after container start

## Motivation

Docker's `restart: unless-stopped` only restarts containers when the process exits. It does not act on healthcheck failures. This means a hung container (process alive but unresponsive) will stay broken indefinitely.

The immediate trigger: Signal K's admin UI restart button hangs the server when WebSocket connections are active (SignalK/signalk-server#2589). The healthcheck is added in halos-org/halos-marine-containers#164; this PR provides the recovery mechanism.

Autoheal also covers Authelia and Valkey (which already have healthchecks) and any future containers that add them.

## Test plan

- [x] Deployed to halosdev.local alongside Signal K healthcheck
- [x] Autoheal container starts and runs healthy
- [x] After Signal K becomes unhealthy (restart hang), autoheal detects and restarts it within ~2 minutes
- [x] Signal K recovers to healthy state after autoheal restart
- [x] All other core containers unaffected

Refs: SignalK/signalk-server#2589, #118


mairas:
## Security Review

### Docker socket access scope (informational, not blocking)

The `:ro` mount flag on `/var/run/docker.sock` does **not** limit API operations. It prevents the container from unlinking or replacing the socket file on disk, but the Docker daemon happily processes write operations (POST, DELETE) received through it. The autoheal container can -- and by design must -- call `POST /containers/{id}/restart`, which works fine over a "read-only" mounted socket.

This is the same access level Traefik and Homarr already have. Within the threat model of a local-only appliance where three containers already have socket access, adding a fourth does not materially change the attack surface. Noting it here so nobody assumes `:ro` constrains API-level capabilities.

### Findings

**1. Unpinned image tag -- supply chain risk (moderate confidence, ~0.65)**

`willfarrell/autoheal:latest` is a community-maintained Docker Hub image with no digest pin. A compromised or hijacked tag would give the attacker a container with Docker socket access on every HaLOS device that pulls the image.

Traefik and Authelia are pinned to specific versions (`traefik:v3.6.13`, `authelia/authelia:4.39.16`). Homarr is pinned to a specific tag on GHCR. Autoheal should follow the same pattern.

**Recommendation:** Pin to a specific version tag and/or SHA256 digest. The current latest is `1.2.0`. Example:
```yaml
image: willfarrell/autoheal:1.2.0
```
Or with digest for stronger guarantees:
```yaml
image: willfarrell/autoheal:1.2.0@sha256:<digest>
```

**2. No network isolation (low-moderate confidence, ~0.60)**

Autoheal only needs the Docker socket -- it makes no network connections. However, it is placed on the default bridge network, which gives it network access to other containers and the host. If the autoheal image were compromised (see finding 1), network access makes lateral movement easier.

The existing `authelia-internal` network pattern shows this codebase already uses network isolation where appropriate.

**Recommendation:** Add `network_mode: "none"` to deny all network access, since autoheal genuinely needs none:
```yaml
  autoheal:
    image: willfarrell/autoheal:1.2.0
    network_mode: "none"
    ...
```

### Not flagged

- Docker socket access itself: consistent with existing architecture (Traefik, Homarr already have it), and required for the stated purpose on a local-only appliance.
- `AUTOHEAL_CONTAINER_LABEL=all`: restarts any unhealthy container. This is the intended behavior per the PR description and is a functionality choice, not a security gap. Scoping to labeled containers only would reduce blast radius but is an operational tradeoff.

mairas:
## Code Review — Correctness

**Bug: Docker socket mounted read-only prevents restarts**

The volume mount uses `:ro`:
```yaml
volumes:
  - /var/run/docker.sock:/var/run/docker.sock:ro
```

Autoheal needs to issue `POST /containers/{id}/restart` via the Docker socket API. A read-only mount blocks write operations — autoheal will detect unhealthy containers but fail to restart them. The [upstream README](https://github.com/willfarrell/docker-autoheal#readme) and all its examples mount the socket without `:ro`. This must be changed to:
```yaml
volumes:
  - /var/run/docker.sock:/var/run/docker.sock
```

I understand the instinct to use `:ro` since other services (traefik, homarr) only read from the socket. But autoheal's whole purpose is to write to it.

**Minor observations (non-blocking):**

1. **Unpinned image tag** — `willfarrell/autoheal:latest` is a floating tag. The upstream repo publishes `1.1.0` as a pinned version. Using `latest` means behavior can change silently on any `docker compose pull`. Consider pinning to `1.1.0` for reproducibility, consistent with how every other image in this compose file is pinned.

2. **`network_mode: none`** — The upstream compose example uses `network_mode: none` since autoheal only communicates via the Docker socket and needs no network access. Adding this is a small security hardening that prevents the container from making or receiving network connections.

3. **`restart: always` vs `unless-stopped`** — The upstream example uses `restart: always`. The difference: after a manual `docker stop` followed by a Docker daemon restart, `always` restarts the container while `unless-stopped` leaves it stopped. For a monitoring service you always want running, `always` is arguably more correct. That said, the existing services in this file all use `unless-stopped`, so consistency is a reasonable argument too.

**Verdict:** The `:ro` mount is a blocking bug — autoheal will silently fail to perform its only job. The other points are suggestions.

mairas:
Addressed the review findings:

**Image pinned** to `willfarrell/autoheal:1.2.0` (from `:latest`).

**Network isolation** added with `network_mode: "none"` — autoheal only needs the Docker socket.

**Re: Docker socket `:ro`** — this is a false positive. The `:ro` flag controls filesystem-level write access to the socket *file* (prevents deletion/replacement), not the API operations sent *through* it. Docker API calls (including `POST /containers/{id}/restart`) work fine over a read-only mounted socket. Confirmed by the live test where autoheal successfully restarted Signal K with `:ro` in place.

Tested on halosdev.local: autoheal 1.2.0 with `network_mode: none` runs healthy and successfully restarts unhealthy containers.

PR #122, opened 2026-04-28
feat: multi-hostname Traefik support (TLS SANs, per-host cookies, OIDC redirect_uris)

mairas:
## Summary

HaLOS now answers Traefik-routed traffic on multiple admin-listed hostnames (LAN mDNS, VPN/DNS aliases, raw IPs). The single-string `HALOS_DOMAIN` baked into cert SANs, Traefik `Host()` matchers, Authelia session cookies, and OIDC `redirect_uris` is replaced by an admin-managed list at `/etc/halos/hostnames.conf`.

**The motivating use case is VPN access.** Today, a device reachable over VPN as `halosdev.example.com` shows a self-signed cert warning (cert SAN doesn't match) and then a Traefik 404 (every router is `Host(\`halosdev.local\`)`). After this change, an admin lists the VPN hostname in `hostnames.conf` and the device just works.

Closes the work tracked in `docs/plans/2026-04-28-001-feat-multi-hostname-traefik-plan.md`.

## What changes

- **`/etc/halos/hostnames.conf`** — new Debian conffile, `root:root 0640`. One DNS hostname or IP per line. The literal token `${hostname}` expands to the device's short hostname at every prestart, so the shipped default `${hostname}.local` tracks `hostnamectl set-hostname` automatically.
- **Shared loader `lib-hostnames.sh`** at `/usr/lib/halos-core-containers/`. Sourced by `prestart.sh` and `reload-oidc-clients`. Five public functions; pinned RFC 1123 DNS regex + IPv4 octet bounds + IPv6 via `getent` round-trip; fail-closed fallback on any malformed entry with `HALOS_HOSTNAMES_FALLBACK` journal diagnostic. Hard cap of 16 entries.
- **Multi-SAN TLS cert** with hash-based change detection. The `.domain` sentinel migrates from a hostname string to a 64-char SHA256 of the sorted hostname list; one regen on upgrade, then only when the list actually changes. Cert + key are written to `.new` siblings and renamed atomically.
- **Authelia per-DNS-hostname session cookies.** Each cookie's `authelia_url` matches its own domain (Authelia 4.39 validates cookie scope). IPs are excluded — RFC 6265 forbids cookies scoped to IP literals. IPs remain valid as cert SANs only.
- **OIDC `redirect_uris` placeholder expansion.** A `${HALOS_DOMAIN}` token in any client snippet at `/etc/halos/oidc-clients.d/*.yml` expands to one URI per configured DNS hostname. Logic is shared between `prestart.sh` and `reload-oidc-clients` via `halos_expand_oidc_redirect_uri`.
- **Path-only Traefik routing.** Authelia (`/sso/`), Cockpit (`/cockpit/`), Homarr (catch-all with explicit `priority: 1`), and per-app redirects from `configure-container-routing` drop their `Host()` matchers. Auth still gates at OIDC's `redirect_uri` allowlist + Authelia ForwardAuth's per-hostname cookies. Localhost redirect keeps `Host(\`localhost\`)` — the documented exception.

## Backward compatibility

- Default config (no admin edits) ships `${hostname}.local` as the single canonical entry → cert is single-SAN, Authelia has one cookie entry, OIDC has one `redirect_uri`. **Identical effective behavior to today.**
- First prestart after upgrade detects the legacy `.domain` hostname string, regenerates the cert once, and writes the new hash. Verified on halosdev.local.

## Testing

- 18 new bash tests in \`tests/test-lib-hostnames.sh\` covering happy paths, fallback paths (missing/empty/only-IPs/cap-exceeded/invalid/dangerous-chars/unreadable), \`\${hostname}\` expansion, hash stability, and OIDC redirect-URI expansion.
- New \`./run test\` task; wired into the CI \`run-tests\` action.
- Real-device integration verified on halosdev.local: cert SANs (DNS×2 + IP×1), path-only routing on \`/cockpit/\`, \`/sso/\`, \`/\`, Host preservation through redirects, localhost redirect intact, OIDC \`redirect_uris\` expansion, Authelia + Traefik + Homarr healthy.

## Code-review (ce:review autofix) findings applied in this branch

- **IP cookies bug (P1)** — Authelia cookies block now excludes IPs. Browsers silently dropped the previous IP-domain Set-Cookies, so ForwardAuth on raw-IP access never worked. Fix verified.
- **AGENTS.md drift (P1)** — the new "Hostname-list contract" section incorrectly listed \`configure-container-routing\` as a loader consumer; corrected.
- **CI test gap (P1)** — \`run-tests/action.yml\` previously ran only pytest, so the 18 new bash tests had no CI signal. Now invokes \`./run test\`.

## Known follow-ups (not blocking)

- Plan doc §93 specifies canonical-only \`authelia_url\`; reality (post-fix 55a0a54) is per-entry — Authelia 4.39 requires per-entry. Plan needs an addendum.
- No automated tests yet for \`prestart.sh\` cert generation, the full OIDC merger function, the Authelia template processor, or \`configure-container-routing\` output. All verified manually on halosdev.local; coverage gap to backfill.
- Cert atomic write is two-step \`mv\` (key then cert); the comment overclaims atomicity. Pre-existing pattern.

## Out of scope (per plan §44)

- Cockpit admin UI for the hostname list.
- Updating the Cockpit \`:9090\` cert (separate trust chain).
- NetworkManager/DHCP/VPN auto-detection.
- Cross-hostname SSO for ForwardAuth (per-hostname sessions are deliberate).
- Grafana heredoc fix in \`halos-marine-containers\` — separate PR.

- **Logs to watch**: \`journalctl -u halos-core-containers.service\` for the prestart output. The signature \`HALOS_HOSTNAMES_FALLBACK\` indicates a malformed config; the device will still boot in single-SAN mode.
- **Health signals**: \`docker ps\` shows authelia, traefik, homarr healthy after a service restart. Authelia restart-loop is the first failure mode if a future config change re-introduces an Authelia-incompatible cookie shape.
- **Validation queries on a deployed device**:
  - \`openssl x509 -in /var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.crt -noout -ext subjectAltName\` — confirm SAN list matches \`hostnames.conf\`.
  - \`sudo grep -A1 'cookies:' /var/lib/container-apps/halos-core-containers/data/authelia/configuration.yml\` — one entry per DNS hostname.
  - \`sudo cat /var/lib/container-apps/halos-core-containers/data/authelia/oidc-clients.yml | grep redirect_uris -A4\` — N entries per OIDC client where N = DNS hostname count.
- **Failure signal / mitigation**: If Authelia crash-loops after a \`hostnames.conf\` edit, revert the file (or empty it to fall back to default), then \`systemctl restart halos-core-containers.service\`. The device boots in single-canonical mode with the original \`\${hostname}.local\`.
- **Validation window / owner**: First production install of v0.3.0 — verify the three queries above and that Cockpit + Homarr + Grafana login flows still work. Owner: matti.airas@hatlabs.fi.


mairas:
## ce:review synthesis (5-reviewer parallel pass)

Reviewers run: correctness, security, adversarial, project-standards, testing.

### Findings applied in this branch (commit b1b13f3)

| # | Severity | Reviewers | File | Finding |
|---|----------|-----------|------|---------|
| 1 | P1 | correctness, security, adversarial (3-way agreement) | `prestart.sh` cookies block | IP entries emitted Authelia `session.cookies` with `domain: <IP>`. RFC 6265 forbids the `Domain` cookie attribute from being an IP literal — browsers silently drop those Set-Cookies, so ForwardAuth on IP-addressed access cannot work. **Fix:** iterate `halos_dns_hostnames` instead of `halos_all_hostnames` for cookies; IPs remain in cert SANs only. README updated to clarify IPs are cert-SAN-only (e.g. for direct HTTPS to Cockpit on `:9090`). |
| 2 | P1 | project-standards | `AGENTS.md` "Hostname-list contract" | Section claimed `configure-container-routing` sources `lib-hostnames.sh` — it doesn't. **Fix:** removed from the "sourced by" sentence and added a clarifying note that the script is now path-only and doesn't need hostname interpolation. |
| 3 | P1 | adversarial, testing | `.github/actions/run-tests/action.yml` | The 18 new bash tests (`tests/test-lib-hostnames.sh`) had no CI signal — the action only ran `pytest`. **Fix:** added a "Run shell tests" step that invokes `./run test`. |

### Residual findings (deferred to follow-ups)

| # | Severity | Reviewer | Finding | Why deferred |
|---|----------|----------|---------|--------------|
| 4 | P2 | adversarial | Cert atomic write is two-step `mv` (key then cert). If the second `mv` fails (ENOSPC, power loss between calls), Traefik may end up with new key + old cert and refuse the handshake on next reload. | Pre-existing pattern in 0.2.13 — this PR didn't introduce the race. The comment overclaims atomicity but the code path matches today's behavior. Worth a follow-up that writes both into a staging dir or uses a single rename of a wrapper directory. |
| 5 | P2 | testing | No automated tests for `prestart.sh` cert generation, the full OIDC merger function, the Authelia template processor, or `configure-container-routing` output. All verified manually on halosdev.local. | The shared loader (`lib-hostnames.sh`) has 18 unit tests. The integration code that consumes it is the gap — should add fixture-based tests in a follow-up PR rather than expand scope here. |
| 6 | P2 | correctness, adversarial | Plan §93 specifies `authelia_url: 'https://CANONICAL/sso'` for every cookie entry; reality (post-fix `55a0a54`) is per-entry. Authelia 4.39's validator rejects canonical-only authelia_url with "does not share a cookie scope". | Plan doc needs an addendum noting the deviation. Code is correct as shipped; verified on-device. |
| 7 | P2 | adversarial | `configuration.yml.template` is not declared as a Debian conffile. Admin-side edits are silently overwritten on package upgrade. | Pre-existing behavior; the template is regarded as system-managed. Out of scope for this PR. |
| 8 | P2 | adversarial, correctness | Authelia config is not validated before commit. If a future hostnames.conf shape produces a config Authelia rejects, the device boots into an Authelia restart-loop with no automatic rollback. | Adding `authelia validate-config` requires running the Authelia container as part of prestart; significant scope addition. The current real-device test confirmed Authelia accepts the post-fix config. Worth a follow-up if we add operations like dynamic IP-cookie support. |
| 9 | P3 | security | TLS private key is briefly mode 0644 between `openssl req -keyout` and `chmod 600`. Local user could race to read it. | Pre-existing (same window in 0.2.13). Mitigated by `${CERTS_DIR}` directory perms. Defense-in-depth fix would be `(umask 077; openssl req ...)`. |
| 10 | P3 | correctness | Loader doesn't deduplicate. If admin lists the same hostname twice, the cert and Authelia config get duplicate entries. | Authelia 4.39 may or may not reject duplicate cookie domains. Cosmetic in cert. Worth a `awk '!seen[$0]++'` pass in the loader. |
| 11 | P3 | adversarial, testing | No tests for IPv6 entries; no tests for multiple `${hostname}` tokens on one line; no test for comment-only file; no golden-value hash assertion to catch silent algorithm drift. | The 18 existing tests cover the documented happy and fail paths well. Additional cases worth backfilling but not blocking. |
| 12 | P3 | security | Stale-alias takeover: an admin-listed DNS name whose ownership later changes can be used by a new owner to intercept HTTPS. Documented as admin's responsibility. | No code-level mitigation; documented in README's "Limitations". Future Cockpit UI could add a journal nag if a configured DNS name fails to resolve. |
| 13 | P3 | adversarial | `homarr-secure` priority=1 contract isn't published for sibling apps. If a future package emits a router with priority=1 on path `/foo/`, behavior is non-deterministic. | Worth adding to AGENTS.md as a "all non-Homarr routers MUST set priority >= 10" note. |

### Verdict

**Ready to merge** once CI passes. The three P1 findings were applied and re-verified on `halosdev.local` (Authelia healthy, cookies block has 2 entries for 2 DNS hostnames + 0 for the 1 IP, cert still has 3 SANs). All residual findings are deferred follow-ups, none blocking.

### Cross-PR coordination

Pairs with [halos-org/halos-marine-containers#170](https://github.com/halos-org/halos-marine-containers/pull/170) (Unit 8 — Grafana heredoc placeholder fix). Recommended ship order: this PR first, then #170, since the Grafana fix relies on the multi-hostname OIDC merger in this PR being live.

🤖 ce:review synthesis from 5 reviewer agents (correctness, security, adversarial, project-standards, testing)

PR #123, opened 2026-04-28
feat: ${fqdn} and ${domain} hostname template tokens

mairas:
## Summary

Extends `/etc/halos/hostnames.conf` token expansion (added in #122) with two new tokens — `${fqdn}` and `${domain}` — that resolve the device's DNS domain at service start. The shipped default now ships three active lines so a freshly installed device answers on whichever LAN resolution mechanism is available without admin action:

```
${hostname}.local    # mDNS — works on any LAN
${hostname}          # bare short name — DHCP-DNS-integrating routers (UniFi, OpenWrt, pfSense, Fritz!Box)
${fqdn}              # full DHCP/admin domain — DHCP option 15
```

**Real-world target:** `halosdev.local` with DHCP option 15 = `hal` now produces cert SANs `halosdev`, `halosdev.hal`, `halosdev.local` automatically.

## Resolver chain

Domain resolution (first non-empty wins):

1. `$HALOS_DOMAIN_RESOLVER` — test injection seam, gated on `declare -F` so a stray env export cannot short-circuit production resolution or execute arbitrary commands.
2. `hostname -d` — admin-set domain (via `/etc/hosts` or `hostnamectl`). Wins over DHCP because explicit configuration beats auto-discovery.
3. `nmcli -t -f IP4.DOMAIN device show` — DHCP option 15, wrapped in `timeout 2` so a hung NetworkManager cannot block prestart.

Resolved values are validated by a new `_halos_domain_safe` predicate (rejects whitespace, NUL, shell metacharacters, anything outside the DNS label set).

## Behavior changes

- **DNS validation regex relaxed** to accept single-label hostnames — both for `${hostname}` (works on DHCP-DNS LANs) and for bare `${domain}` returning a single-label DHCP domain. Defense-in-depth `_halos_has_dangerous_chars` is unchanged.
- **Soft-drop semantics** for `${fqdn}`/`${domain}` lines whose expansion is invalid solely because no domain resolved. Logged as `HALOS_HOSTNAMES_SKIP` (distinct from the existing `HALOS_HOSTNAMES_FALLBACK` for hard failures), so admins can grep network-state issues vs configuration errors apart. All other validation failures keep the existing fail-closed behavior.
- **Single-label hostnames excluded from Authelia session cookies** (mirroring the existing IP exclusion). Authelia 4.39+ rejects single-label cookie Domain attributes at config-load time per RFC 6265 §5.3 step 5; single-label entries remain valid as cert SANs, OIDC \`redirect_uris\`, and Traefik path-only routing targets.

## Trust boundary (deliberate)

When `${fqdn}`/`${domain}` falls through to nmcli, DHCP option 15 flows into cert SANs / Authelia cookies / OIDC `redirect_uris`. Documented in AGENTS.md. Defenses: (a) `hostname -d` wins so admins can pin the domain via `hostnamectl`, (b) `_halos_domain_safe` validates resolver output, (c) the regex still rejects malformed DNS strings. Devices on hostile networks should remove the `${fqdn}` line.

## Tests

Test suite grows from 18 → 37. New tests cover:
- Resolver injection seam (returns / empty / whitespace-trim / cache-clears-on-reload / non-existent-function-falls-through / cache-prevents-repeated-calls / dangerous-chars-rejected)
- Single-label literal and `${hostname}` token
- `${fqdn}` expansion, `${hostname}.${domain}` equivalence, bare `${domain}` (multi-label and single-label)
- Soft-drop with surviving entries / only-fqdn / bare-`${domain}` / typo-neighbor-still-hard-fails
- Full shipped-default layout with both multi-label and single-label DHCP domains
- Soft-drop diagnostic asserts the dropped line content (not just the prefix)

The nmcli code path itself is exercised by real-device verification rather than unit tests, since CI lacks NetworkManager.

## Verified on halosdev.local

- Service active, all containers healthy (traefik, authelia, homarr, authelia-valkey, autoheal)
- Cert SANs: `halosdev`, `halosdev.hal`, `halosdev.local`
- Authelia cookies: 2 entries (`halosdev.local`, `halosdev.hal`) — bare `halosdev` correctly filtered
- OIDC `redirect_uris`: 3 entries per app
- HTTP 200 on all three hostnames

## Related

- Plan: `docs/plans/2026-04-28-002-feat-fqdn-domain-tokens-plan.md`
- Builds on: #122 (multi-hostname Traefik support)

## Conffile upgrade caveat

`/etc/halos/hostnames.conf` is a Debian conffile. Existing devices with an unmodified file pick up the new defaults on upgrade automatically. Devices where an admin has edited the file will see a `dpkg` conffile prompt; admins who want the new `${hostname}` and `${fqdn}` lines should accept the new file or merge in those lines manually.

**Logs to watch:** `journalctl -u halos-core-containers.service` for prestart output. Two stable signatures:
- `HALOS_HOSTNAMES_SKIP: domain unresolved, dropping line: <line>` — informational, expected on devices without DHCP option 15.
- `HALOS_HOSTNAMES_FALLBACK: <reason>` — admin error in hostnames.conf; device boots in single-SAN default mode.

**Health signals:** `docker ps` shows authelia, traefik, homarr healthy after restart. Authelia restart-loop is the first failure mode if a future change re-introduces a single-label cookie domain.

**Validation queries on a deployed device:**
- `openssl x509 -in /var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.crt -noout -ext subjectAltName` — confirm SANs match the expected expansion of hostnames.conf.
- `sudo grep -A1 'cookies:' /var/lib/container-apps/halos-core-containers/data/authelia/configuration.yml` — one cookie entry per multi-label DNS hostname (single-labels and IPs excluded).
- `sudo grep redirect_uris -A4 /var/lib/container-apps/halos-core-containers/data/authelia/oidc-clients.yml` — N entries per OIDC client where N = DNS hostname count.
- `nmcli -t -f IP4.DOMAIN device show` — see what DHCP advertises if `${fqdn}` is unexpectedly soft-dropped.

**Failure signal / mitigation:** If Authelia crash-loops after a `hostnames.conf` edit, revert the file (or empty it to fall back to default), then `systemctl restart halos-core-containers.service`.

**Validation window / owner:** First production install — verify the three queries above on a device with DHCP option 15 set, and on one without (typical home LAN). Owner: matti.airas@hatlabs.fi.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -95,11 +95,32 @@ halos-core-containers/
 
 `/etc/halos/hostnames.conf` is the admin-managed source of truth for every
 hostname this device answers to (cert SANs, Authelia cookies, OIDC
-`redirect_uris`). The shared loader at
-`/usr/lib/halos-core-containers/lib-hostnames.sh` is sourced by
-`prestart.sh` and `reload-oidc-clients`; do not duplicate parsing logic —
-extend the loader instead. (`configure-container-routing` no longer needs
-hostname interpolation because all per-app routers are now path-only.)
+`redirect_uris`). User-facing reference: [docs/HOSTNAMES.md](docs/HOSTNAMES.md).
+
+The shared loader at `/usr/lib/halos-core-containers/lib-hostnames.sh` is
+sourced by `prestart.sh` and `reload-oidc-clients`; do not duplicate
+parsing logic — extend the loader instead. (`configure-container-routing`
+no longer needs hostname interpolation because all per-app routers are
+now path-only.)
+
+**Loader extension points.** When adding a new consumer of the parsed
+hostname list:
+- Read `HALOS_HOSTNAMES_DNS[]` and `HALOS_HOSTNAMES_IPS[]` after calling
+  `halos_load_hostnames`. Don't reparse the file.
+- If the consumer can't accept single-label DNS entries (e.g., Authelia
+  session cookies — see RFC 6265 §5.3 step 5), filter inside the consumer,
+  not in the loader. The cookies block in `prestart.sh` is the reference
+  pattern. IP entries are similarly excluded there.
+- Diagnostics: `HALOS_HOSTNAMES_SKIP` for soft-drops (network-state),
+  `HALOS_HOSTNAMES_FALLBACK` for hard fails (admin error). Keep these
+  signatures stable — operators grep them.
+
+**Test injection seam.** `HALOS_DOMAIN_RESOLVER` names a defined shell
+function whose stdout becomes the resolved domain. Gated on `declare -F`
+so a stray environment export cannot short-circuit production resolution
+or trigger command execution. When honored, the function's output
+(including empty) is authoritative — no fall-through. Tests use this to
+pin empty-domain behavior independently of the host network.
 
 **OIDC client snippet placeholder convention.** App `prestart.sh` scripts
 that drop OIDC client snippets into `/etc/halos/oidc-clients.d/` should
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,6 +72,10 @@ halos-core-containers/
 │       └── icon.png
 ├── tools/
 │   └── build-all.sh         # Build all packages
+├── docs/
+│   ├── HOSTNAMES.md         # operator reference for /etc/halos/hostnames.conf
+│   ├── SSO_SPEC.md, SSO_ARCHITECTURE.md
+│   └── solutions/           # documented solutions to past problems (bugs, gotchas, patterns), with YAML frontmatter for searchability; relevant when implementing or debugging in documented areas
 ├── .github/workflows/
 │   └── main.yml             # CI/CD
 └── README.md
--- a/README.md
+++ b/README.md
@@ -25,16 +25,27 @@ See [docs/SSO_SPEC.md](docs/SSO_SPEC.md) and [docs/SSO_ARCHITECTURE.md](docs/SSO
 
 ## Multi-Hostname Configuration
 
-By default, HaLOS only answers to its `${hostname}.local` mDNS name. To make
-the device reachable over a VPN, an admin-configured DNS alias, or a raw IP,
-add additional entries to `/etc/halos/hostnames.conf`:
+By default, HaLOS answers to two names:
+
+- `${hostname}.local` — mDNS, works on any LAN.
+- `${fqdn}` — full DHCP/admin domain (e.g., `halosdev.example.com` when
+  DHCP option 15 is set or the admin runs `hostnamectl set-hostname`).
+
+Both are listed in the shipped `/etc/halos/hostnames.conf`. To add VPN
+aliases or other multi-label DNS names, edit the file:
 
 ```
 # /etc/halos/hostnames.conf
-${hostname}.local           # canonical (first non-IP entry)
+${hostname}.local           # canonical (first entry)
+${fqdn}                     # auto-resolved DHCP/admin FQDN
 halosdev.example.com        # VPN/DNS alias
-10.0.0.50                   # raw IP, cert SAN only (no auth on IP-addressed access)
 ```
+
+Single-label hostnames and raw IPs are deliberately not in the example
+set — they cannot carry an Authelia session cookie (RFC 6265 §5.3 step
+5; Authelia 4.39+ rejects them outright), so SSO breaks on them in
+non-obvious ways. See [docs/HOSTNAMES.md](docs/HOSTNAMES.md) for the
+full token reference, resolver chain, and trust-boundary notes.
 
 After editing, restart the service:
 
--- a/assets/hostnames.conf
+++ b/assets/hostnames.conf
@@ -1,20 +1,32 @@
 # HaLOS hostname list (admin-managed)
 #
-# One DNS hostname or IP address per line.
-#   - The first non-comment, non-IP line is canonical (used as the OIDC issuer).
+# One DNS hostname per line. Use multi-label hostnames only — bare
+# single-label hostnames and raw IP addresses cannot carry an Authelia
+# session cookie, so SSO breaks on them in non-obvious ways. See
+# /usr/share/doc/halos-core-containers/HOSTNAMES.md for the full rationale.
+#
+#   - The first non-comment line is canonical (used as the OIDC issuer).
 #   - Lines starting with '#' are comments. Blank lines are ignored.
 #   - Maximum 16 entries. Malformed entries cause fallback to single-SAN default.
 #
-# The literal token ${hostname} expands to this device's short hostname,
-# so the default below stays correct when you rename the device with
-# `hostnamectl set-hostname`. Replace it with literal entries to pin.
+# Supported tokens (expanded at service start):
+#   ${hostname}   the device's short hostname (e.g., halosdev)
+#   ${domain}     the device's DNS domain (see resolution chain below)
+#   ${fqdn}       shorthand for ${hostname}.${domain}
+#
+# Domain resolution order (first non-empty wins):
+#   1. `hostname -d` (admin-set via /etc/hosts or hostnamectl)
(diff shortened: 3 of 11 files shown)


mairas:
## ce:review (final-state, mode:report-only)

Six reviewers (correctness, maintainability, project-standards, testing, security, reliability) on the post-refinement state. Prior round's P1s — `nmcli` timeout, `declare-F` gating, soft-drop semantics — already addressed. New findings below.

### P1 — High (worth fixing before merge)

|---|---|---|---|---|
| 1 | `prestart.sh:539-562` | **Empty `cookies_block` produces invalid Authelia config → crash-loop.** When every entry in `HALOS_HOSTNAMES_DNS` is single-label (admin pathological config), the case-statement filter strips them all, `cookies_block` ends as the empty string, the rendered template has a `cookies:` key with no list items, and Authelia 4.39 rejects at config-load time. Prestart exits 0 (systemd "success"); only container logs reveal the failure. Fix: after the loop, detect empty and either synthesize a fallback entry from `${hostname}.local` (boot-availability semantics) or fail prestart with a clear error. | security, reliability | 0.85 |
| 2 | `assets/lib-hostnames.sh:296-368` + `prestart.sh:547-550` | **No deduplication of `HALOS_HOSTNAMES_DNS`.** When DHCP option 15 = `local` (a real SOHO config), `${fqdn}` expands to `<host>.local` exactly colliding with the canonical mDNS line. Result: duplicate `DNS:` SANs in the cert and two identical cookie blocks; Authelia rejects duplicate cookie-domain entries at load. Same crash-loop class as #1. Fix: dedup (case-insensitive, preserve first occurrence) inside `halos_load_hostnames` before populating `HALOS_HOSTNAMES_DNS[]`. | security | 0.85 |

### P2 — Moderate (doc/test drift from the design pivot)

|---|---|---|---|---|
| 3 | `README.md:28-37` | **Multi-Hostname Configuration section is stale.** Says "By default, HaLOS only answers to its `${hostname}.local` mDNS name" — false; default now also ships `${fqdn}`. Example block uses raw-IP and literal-FQDN as the only suggested patterns, contradicting the conffile's new policy of suggesting `${fqdn}` first and discouraging single-label/IP entries. Fix: update to match `assets/hostnames.conf` and `docs/HOSTNAMES.md`. | correctness | 0.95 |
| 4 | `tests/test-lib-hostnames.sh:454` | **Comment says "Mirrors the shipped default: `${hostname}.local`, `${hostname}`, `${fqdn}`"** — but the shipped default no longer contains bare `${hostname}`. The test itself is still valuable as loader-API coverage (literal + token + fqdn + IP mix), but its self-description encodes a falsehood. Fix: update comment to "exercises the loader with a mixed token+literal+IP config; the actually-shipped default is mDNS + `${fqdn}` only — see `test_shipped_default_with_single_label_dhcp_domain`". | correctness, maintainability, testing | 0.90 |
| 5 | `docs/plans/2026-04-28-002-feat-fqdn-domain-tokens-plan.md:37, 170-196` | **Plan R7 / Unit 3 still describe the rejected three-line default** (`${hostname}.local`, `${hostname}`, `${fqdn}`). The pivot rationale (RFC 6265 single-label cookie incompatibility, Authelia 4.39 hard-fail) lives in `docs/HOSTNAMES.md` but the plan has no breadcrumb to it. Future agents using the plan as design reference will find it asserts the opposite of the shipped state. Fix: add a one-line "Refinement 2026-04-28: dropped bare `${hostname}` from default — see `docs/HOSTNAMES.md` §single-label" near R7, or strike R7's third entry. | correctness, maintainability | 0.80 |

### P3 — Low (advisory)

|---|---|---|---|---|
| 6 | `prestart.sh:539-562` | No automated test for the new single-label cookie filter. The loader actively produces single-label entries (admin opt-in path), and this filter is the sole guard between the loader output and an Authelia config Authelia 4.39+ refuses. PR #122 set the scope precedent of "no automated tests for `prestart.sh`", so this is a continuation of that gap rather than a new one — but it's now larger. | maintainability, testing | 0.85 |
| 7 | `assets/lib-hostnames.sh` `_halos_resolve_domain` | **Empty resolver result has no diagnostic.** Operator sees `HALOS_HOSTNAMES_SKIP` per affected line but no signal as to *why* the chain returned empty (nmcli absent? DHCP option 15 missing? hostname -d unsafe-and-dropped?). Symmetric to the existing "hostname -d returned unsafe value" log. Fix: emit one summary line per `halos_load_hostnames` call when the chain exhausts. | reliability | 0.75 |
| 8 | `assets/lib-hostnames.sh` soft-drop logs | `HALOS_HOSTNAMES_SKIP` for the shipped-default `${fqdn}` is unconditional journal noise on no-DHCP-15 LANs (very common). Fires on every prestart and every `reload-oidc-clients` invocation. Consider suppressing when the dropped line matches a shipped-default token-only entry, or downgrading to one summary line per load. | reliability | 0.65 |
| 9 | `docs/HOSTNAMES.md` | Missing the Version/Date/Status header block that sibling docs (`SSO_SPEC.md`, `SSO_ARCHITECTURE.md`) carry. Not a codified rule, but a de-facto convention in this repo's `docs/`. | project-standards | 0.55 |

- DHCP option 15 trust extension into cookie `Domain` and OIDC `redirect_uris` — already documented in `docs/HOSTNAMES.md` as a deliberate tradeoff.
- Conffile dpkg prompt on upgrade for devices with edited `hostnames.conf` — standard Debian conffile semantics; release notes should mention.

### Verdict

**Ready with fixes.** P1 items (#1 and #2) are real reliability holes — both produce silent Authelia crash-loops on plausible admin configs (#1: only-single-label admin opt-in; #2: DHCP option 15 = `local` on a SOHO router). The dedup fix in #2 also subsumes most of #1 since deduping single-label-with-multi-label avoids the all-filtered scenario in many cases.

P2 items (#3–#5) are cheap doc/test drift cleanup that should land with the same PR; future readers will conflate the plan's R7 with the shipped reality otherwise.

P3 items can defer.

🤖 Generated with [ce:review](https://claude.com/claude-code)

mairas:
## P1+P2 fixes applied — commit 7b47c35

### P1 — reliability

- **Dedup in `halos_load_hostnames`** (lib-hostnames.sh): case-insensitive for DNS, exact for IPs, first-wins. Verified on halosdev.local with `hostnamectl set-hostname halosdev.local` so `${fqdn}` collides with the mDNS canonical → 1 SAN, 1 cookie (was 2 and 2). Two new tests: `test_dedup_dns_entries_case_insensitive`, `test_dedup_ip_entries`.
- **Empty cookies_block fallback** (prestart.sh): when every DNS entry is filtered out (admin-pinned all-single-label config), synthesize a fallback entry from `<short>.local` and emit `WARN: hostnames.conf produced no multi-label DNS entries...`. Verified on halosdev.local with `/etc/halos/hostnames.conf` containing only `${hostname}`: prestart logs the WARN, Authelia stays healthy on the synthesized canonical instead of crash-looping.

### P2 — doc/test drift

- **README.md**: replaced the stale "only answers to ${hostname}.local" + IP/single-label examples with the real two-line shipped default and a pointer to `docs/HOSTNAMES.md`.
- **tests/test-lib-hostnames.sh:454**: comment now correctly identifies the test as loader-API coverage, not the "shipped default".
- **plan R7**: added a Refinement note explaining why bare `${hostname}` was dropped (RFC 6265 §5.3 step 5 + Authelia 4.39 hard-fail).

### Verification

All three scenarios passed on halosdev.local:
- Golden path (shipped default + DHCP option 15 = `hal`): 2 SANs (`halosdev.hal`, `halosdev.local`), 2 cookies.
- Dedup (collision via `hostnamectl set-hostname halosdev.local`): 1 SAN, 1 cookie, no duplicate.
- Empty-cookies fallback (only `${hostname}` in conffile): WARN logged, synthesized `halosdev.local` cookie, Authelia healthy.

Tests: 37 → 39, all green.

P3 items deferred per the original review.

🤖 Generated with [ce:review](https://claude.com/claude-code)

Issue #133, opened 2026-05-24
Serve /halos-ca.crt download endpoint via Traefik


mairas:
Sub-issue of #129 (Unit 4 of 4).

## Scope

Serve the active CA certificate at `https://<host>/halos-ca.crt` so operators can fetch + install it once into their OS/browser trust store and stop seeing warnings across this device's services.

## Open implementation question

Traefik does not serve static files natively. Three approaches:

### Option A: Tiny static-file sidecar

Add a minimal nginx/busybox container to `docker-compose.yml` that serves only `/halos-ca.crt`. Mount `serving-ca.crt` (symlink from Unit 2 — #131) read-only. Traefik routes `/halos-ca.crt` to this sidecar.

**Pros:** clean separation of concerns; sidecar is replaceable.
**Cons:** extra container (RAM + image pull on first boot); one more thing to ship and version.

### Option B: Piggyback on Homarr's nginx

Homarr's container has an internal nginx serving its static assets. Mount `serving-ca.crt` into the Homarr container at a path nginx already serves, add a Homarr nginx route for `/halos-ca.crt`.

**Pros:** no new container.
**Cons:** couples cert distribution to Homarr; breaks if Homarr image changes its nginx config; trust-anchor distribution shouldn't depend on dashboard availability.

### Option C: Cockpit-served on :9090

Have Cockpit serve `/halos-ca.crt` at `https://<host>:9090/halos-ca.crt`. Traefik routes `/halos-ca.crt` at `:443` through to Cockpit's static-file mechanism.

**Pros:** no new container; ties trust-bootstrap to the always-available backdoor.
**Cons:** Cockpit isn't a static-file server by design; mixing concerns; requires Cockpit branding/static work in `halos-cockpit-config`.

**Recommendation: Option A.** Smallest blast radius, clearest ownership, easiest to reason about. Image candidates: `nginx:alpine` (~7MB), `busybox` httpd (~1MB), or even a tiny socat-based static server. Pick at implementation time.

## Headers

```
Content-Type: application/x-x509-ca-cert
Content-Disposition: attachment; filename=halos-ca.crt
```

`attachment` ensures browsers offer the cert-install dialog rather than rendering as text.

## Transport

[...]

Issue #129, opened 2026-05-24
Add custom-CA support with shared Cockpit leaf


mairas:
## Problem

Today HaLOS generates a self-signed *leaf* TLS certificate at first boot (`halos-core-containers/prestart.sh`). The cert has no CA Basic Constraints, so Windows/macOS/Linux trust stores refuse to use it as a trust anchor — importing it into "Trusted Root CAs" doesn't suppress browser warnings, because Chrome/CryptoAPI chain validation requires the anchor to have `CA:TRUE`. The only working UX today is per-site click-through ("Advanced → Proceed"), which is invalidated every time the leaf re-signs.

Cockpit on `:9090` separately uses its own upstream-generated self-signed cert (`/etc/cockpit/ws-certs.d/0-self-signed.cert` from `cockpit-certificate-ensure`), giving users a *second* cert warning if they hit the device directly on `:9090` instead of going through Traefik.

## Design

Introduce a proper CA. Auto-generate one on first boot; let advanced operators replace it with their own (which lets fleet operators install one trust anchor on their laptop and trust every HaLOS device that uses it). Share the same leaf with Cockpit `:9090` so both surfaces serve the same cert.

### File layout

| Path | Owner | Notes |
|---|---|---|
| `/var/lib/container-apps/halos-core-containers/certs/ca/ca.{crt,key}` | system | Auto-generated on first boot. 20y validity, `notBefore = now - 24h`. |
| `/etc/halos/ca/ca.{crt,key}` | operator | Optional drop slot. Presence + validity = "custom CA in use". |
| `/var/lib/.../certs/ca/serving-ca.crt` | system | Symlink to the active CA, mounted by Traefik for the download URL. |
| `/var/lib/.../certs/halos.{crt,key}` | system | Existing leaf paths (unchanged location); now signed by the active CA instead of being self-signed. |
| `/etc/cockpit/ws-certs.d/99-halos.cert` | system | Combined PEM (cert + key), mode 0640, group `cockpit-ws`. Override that makes Cockpit `:9090` serve the same leaf as Traefik. |

### Active CA selection

On every prestart:
1. If `/etc/halos/ca/ca.{crt,key}` both exist → validate (CA:TRUE in basicConstraints, keyCertSign in keyUsage, key matches cert, dates parse).
   - Valid → active = custom.
[...]

PR #135, opened 2026-05-25
feat(certs): share TLS leaf with Cockpit on :9090

### LINKED ISSUE #132: Share TLS leaf with Cockpit on :9090
Sub-issue of #129 (Unit 3 of 4).

## Scope

Make Cockpit's `:9090` port serve the same TLS leaf as Traefik so operators see one trust experience across both surfaces — and so the operator-installed CA from Unit 2 (#TBD) covers `:9090` too.

`:9090` is the deliberate backdoor: if Traefik is wedged, it must remain reachable. The shared leaf gives it trusted TLS on every configured hostname (SANs come from `halos_all_hostnames`) without making the backdoor depend on Traefik being healthy. cockpit-ws / cockpit-tls runs natively under systemd, independent of the container stack.

## Implementation

In prestart, after `halos_ca_sign_leaf` succeeds:

1. Write `/etc/cockpit/ws-certs.d/99-halos.cert.new` containing the concatenated PEM (leaf cert + leaf key), mode `0640`, group `cockpit-ws`.
2. Validate the file: parse the cert, parse the key, confirm the key matches the cert.
3. On validation pass: atomic-rename to `99-halos.cert`.
4. On validation fail: log loud, exit non-zero, leave previous `99-halos.cert` in place.

### Why `99-` prefix

Cockpit picks the lexicographically last `.cert` file. `99-` overrides the upstream `0-self-signed.cert` (generated by `cockpit-certificate-ensure` from `cockpit-ws`). The upstream cert stays as the safety-net floor: if our override is ever absent, Cockpit falls back to the upstream-generated self-signed and `:9090` stays reachable. **Removing `99-halos.cert` is the documented "revert" path.**

### Permissions

The source files (`/var/lib/.../certs/halos.{crt,key}`) are owned by root. The copy into `/etc/cockpit/ws-certs.d/99-halos.cert` needs `cockpit-ws` group read so `cockpit-tls` can read the key. Two copies of the key on disk; both on the same device, both 0640 or stricter. Acceptable security delta.

### No auto-reload

Per the discussion in #129, no systemd path unit. `cockpit-tls` picks up new certs at socket activation; existing sessions keep the old cert until disconnect. Cert rotation is rare (admin-initiated hostname-list change or CA swap), so the manual `systemctl restart cockpit` workflow is fine. Document this in `docs/CERTS.md`.

## Tests to add

- After leaf re-sign, `99-halos.cert` exists and contains valid PEM cert + key
- File mode is 0640, group is `cockpit-ws`
- Cockpit picks the file (`cockpit.socket` restart, then verify the served cert matches)
- Validation failure: malformed concatenation → previous override preserved, prestart non-zero
- Removing `99-halos.cert` → next prestart re-creates it
- Removing prestart-write capability (e.g., dir RO) → prestart fails with clear error, previous override survives

## Sequencing

Independent of Units 2 and 4. Could land in parallel with Unit 2. Depends on #130 (merged).

mairas:
## Why

Cockpit on `:9090` is the deliberate backdoor — when Traefik wedges, it must stay reachable. Today it uses its own upstream-generated self-signed cert (`/etc/cockpit/ws-certs.d/0-self-signed.cert` from `cockpit-certificate-ensure`), giving operators a *second* cert warning distinct from Traefik's. After the umbrella plan in #129 introduced a per-device CA, we can make both surfaces serve the same leaf so an operator-installed CA covers `:9090` too.

## How

After leaf signing, prestart writes the concatenated PEM (cert + key) to `/etc/cockpit/ws-certs.d/99-halos.cert`, mode `0640`, group `cockpit-ws` when that group exists. `cockpit-tls` picks the lex-last `.cert` at socket activation, so `99-` overrides the upstream `0-self-signed.cert`. The upstream cert stays as the safety-net floor — removing `99-halos.cert` reverts to upstream-generated self-signed without further action.

The new helper `halos_cockpit_install_leaf` validates parse-as-cert + parse-as-key + key-matches-cert before the atomic `.new`+`mv` swap. Cannot reuse `halos_ca_validate_pair` because that one demands `CA:TRUE` (fails on a leaf); the key-match technique is inlined.

Skips cleanly when `/etc/cockpit/ws-certs.d/` is absent so install-order variance (cockpit-ws is pulled in via `halos-cockpit-config`'s Recommends) doesn't fail prestart. Runs unconditionally otherwise — cheap on no-change, self-heals if the override file was deleted manually.

No systemd path unit / auto-reload — operators run `sudo systemctl restart cockpit` for immediate effect (existing sessions keep the old cert until disconnect). Documented in `docs/CERTS.md`.

## Tests

`bash tests/test-lib-ca.sh` — 51 passed, 0 failed (44 pre-existing + 7 new). New tests cover:

- valid combined PEM (cert + key both parse from output)
- output mode 0640
- mismatched cert/key pair → rc=1, no output file
- existing file preserved on validation failure
- no `.new` debris on success
- rc=2 on missing args, rc=1 on missing source files

Group ownership and the cockpit-ws-not-installed fallback are not unit-testable cross-environment; runtime fallback (root:root + NOTICE) is implicitly exercised on environments without the group.

## Notes

- VERSION not bumped: latest stable tag is `v0.3.2+1`; VERSION is already `0.3.4` (ahead) so no further bump needed per workspace policy.
- `docs/CERTS.md` is a new file — #133's PR adds its own "CA download endpoint" section to the same file, so whichever of these two PRs lands second will hit a clean merge conflict (additive sections, easy resolve).

Refs #129, closes #132.


mairas:
## ce:review (multi-persona)

**Scope:** `feat/cockpit-cert-share` vs `main` — 4 files, +318 lines (`assets/lib-ca.sh`, `prestart.sh`, `tests/test-lib-ca.sh`, `docs/CERTS.md` new).

**Mode:** report-only (no autofix applied; findings posted for the author's policy call).

**Intent:** After leaf signing, install combined-PEM override at `/etc/cockpit/ws-certs.d/99-halos.cert` (mode 0640, group `cockpit-ws` when present) so Cockpit `:9090` serves the same TLS leaf as Traefik. New `halos_cockpit_install_leaf` helper validates parse + key-match via `openssl pkey -pubout`, atomic `.new`+`mv`. Skip-when-dir-absent for install-order variance.

**Reviewer team (9):** correctness, testing, maintainability, project-standards, agent-native, learnings, security, reliability, adversarial. Conditional picks: security (cert handling + private key + group perms), reliability (prestart + `set -e` + atomic writes + init-order), adversarial (>50 LOC + same security-adjacent area as `#134` which surfaced 3 P1/P2).

---

### P1 — Critical

|---|---|---|---|---|---|
| 1 | `prestart.sh:216` | Unguarded helper call under `set -e` makes Cockpit cert failure fatal to entire web stack | correctness, reliability, adversarial (3-way) | 0.95 | gated_auto → downstream-resolver |

**Detail:** `prestart.sh:4` sets `set -e`. The call at line 216 is unguarded — any non-zero return from the helper (any of: missing source files, openssl parse, pubkey extract/derive, mv across FS boundary, transient I/O) aborts prestart **before** Traefik dynamic TLS config (`tls-default.yml`), Cockpit redirect router, Authelia OIDC, Homarr setup. A cosmetic feature (sharing the leaf on `:9090`) gates the primary serving surface (`:443`). This contradicts the documented "safety-net floor" design — `docs/CERTS.md` claims the upstream `0-self-signed.cert` is the fallback, but if prestart aborts, Traefik never starts either. Fix: wrap with `|| echo "WARN: cockpit leaf install failed; :9090 will keep its previous cert" >&2`.

---

### P2 — High

|---|---|---|---|---|---|
| 2 | `assets/lib-ca.sh:531-535` | Silent chown failure leaves cockpit unable to read the leaf private key | correctness, maintainability, reliability, adversarial + security residual (4-way) | 0.92 | gated_auto → downstream-resolver |
| 3 | `assets/lib-ca.sh:497, 502` | Parse-failure branches in helper untested (cert/key both unparseable) | testing | 0.85 | manual → review-fixer |
| 4 | `assets/lib-ca.sh:531` | `getent group cockpit-ws` present-branch and absent-NOTICE branch both untested | testing | 0.80 | manual → review-fixer |
| 5 | `assets/lib-ca.sh:488` | Missing existing-symlink check on `.new` write path (defense-in-depth) | adversarial | 0.72 | manual → downstream-resolver |

**Detail [2]:** `chown root:cockpit-ws "$out_new" 2>/dev/null || true` swallows chown failure. If the group exists but chown fails (NFS, SELinux/AppArmor denial, weird ACLs, transient race), the file lands `0640 root:root`. cockpit-tls runs as the cockpit-ws GID and cannot read a root:root file. Helper returns 0 — silent failure that hides the breakage from operators. The else-branch (group missing) logs a NOTICE; the chown-fails-but-group-exists path has no diagnostic. Asymmetric. Fix: drop `|| true`, rm `.new`, log clearly, return 1.

**Detail [3]:** The 7 new tests exercise mismatched-but-parseable pairs only. The cert-parse and key-parse early-exit branches in the helper are never executed in tests. A regression that weakens either parse check still passes all tests. Fix: add two tests with malformed cert and malformed key.

**Detail [4]:** Tests don't `getent`-shadow or otherwise exercise either branch of the group-existence check. A regression that swapped the branches (or returned non-zero from the absent path) would break Cockpit TLS readability without unit-test signal. Fix: function-shadow `getent() { return 0; }` and `getent() { return 2; }` to drive both paths and assert NOTICE-on-stderr + file install.

**Detail [5]:** `cat ... > "$out_new"` follows symlinks. If a pre-existing symlink at `99-halos.cert.new` points elsewhere, the redirection writes the leaf cert+key to the symlink target. Stock Debian ws-certs.d is `0755 root:root` so requires root; some hardening guides chgrp it to `cockpit-ws` with group-write — at which point this becomes exploitable by any cockpit-ws member. Fix: `rm -f "$out_new"` before the `cat` redirection.

---

### P3 — Low

|---|---|---|---|---|---|
| 6 | `assets/lib-ca.sh:537` | `mv` failure not rc-checked — `.new` debris and helper returns mv's rc without cleanup | correctness | 0.70 | safe_auto → review-fixer |
| 7 | `assets/lib-ca.sh:524` | `chmod 0640` failure on `.new` not rc-checked | reliability | 0.60 | safe_auto → review-fixer |
| 8 | `assets/lib-ca.sh:537` | `mv` onto pre-existing directory at `$out_path` silently succeeds (mv-into-dir) | adversarial | 0.78 | manual → downstream-resolver |
| 9 | `assets/lib-ca.sh:484` | Concurrent prestart invocations race on shared `${out_path}.new` filename | reliability | 0.62 | manual → downstream-resolver |
| 10 | `tests/test-lib-ca.sh:845` | Missing-files test passes both paths missing — leaf_key branch never exercised | testing | 0.85 | safe_auto → review-fixer |
| 11 | `assets/lib-ca.sh:488` | Write-failure branch (missing parent dir) untested | testing | 0.75 | safe_auto → review-fixer |
| 12 | `prestart.sh:214` | No re-prestart trigger after cockpit-ws is later installed — indefinite trust inconsistency window | adversarial | 0.74 | advisory → human |
| 13 | `assets/lib-ca.sh:497` | Validate-on-`$out_new`-not-inputs rationale undocumented (one-line comment gap) | maintainability | 0.62 | safe_auto → review-fixer |
| 14 | `prestart.sh:209` | Prestart skip-branch (dir absent) untested — integration only | testing | 0.70 | advisory → release |

---

### Residual risks (no concrete fix; documented or operational)

- **Two on-disk copies of the leaf private key** (source `halos.key` + combined PEM in `/etc/cockpit/ws-certs.d/`). cockpit-ws group can now read the leaf private key — documented design choice in `docs/CERTS.md`, but expands lateral-movement blast radius (cockpit-tls CVE → device TLS leaf exfiltration).
- cockpit-tls fail-closed behavior on an unreadable lex-last file (after chown failure) is **assumed** — not verified against cockpit-tls source.
- Combined PEM is **cert-then-key** order (per `cat "$leaf_crt" "$leaf_key"`). cockpit-tls expects cert-then-key per its docs; correct.
- TOCTOU between `[ -d "$COCKPIT_WS_CERTS_DIR" ]` and helper openssl calls (low — install-order variance only).

---

- **Clean** on the openssl-text-grep substring-injection pattern fixed in #134. Helper uses `openssl pkey -pubout` exact-string equality, not `openssl x509 -text | grep`. See `docs/solutions/2026-05-25-openssl-text-grep-defeated-by-subject-dn.md`.
- **Clean** on the `set -e` cmd-subst blind spot **inside the helper** (every `$(openssl …)` capture is rc-guarded). See `docs/solutions/2026-05-24-set-e-cmdsubst-blind-spot.md`. The **call site at `prestart.sh:216`** is unguarded — that's finding [1].

---

### Agent-native

**PASS** on all three axes (introspection / control / recoverability). The override file is filesystem-mediated and shell-tool-accessible. Revert is documented (`rm + systemctl restart cockpit`); verification via `openssl s_client -connect :9090` is scriptable.

Nit: `docs/CERTS.md` could include fingerprint-compare one-liners for agent introspection (PR #136's CERTS.md has them — both will merge into the same file anyway).

---

### Project standards

Clean. Conventional commit, branch name, VERSION (already ahead of `v0.3.2+1` per workspace policy), debian/changelog discipline, `Refs #129, closes #132` umbrella linkage, hostname policy — all pass.

---

### Verdict

**Ready with fixes.** Recommended order:

1. **[P1 #1]** Wrap helper call with `|| echo "WARN: ..." >&2` — single-line, restores the documented "safety-net floor" semantic
2. **[P2 #2]** Drop `|| true` on chown when group exists; surface as rc=1
3. **[P2 #3, #4]** Add tests for parse-failure branches and `getent`-shadowed group branches
4. **[P2 #5]** `rm -f "$out_new"` before `cat` (symlink defense-in-depth)
5. **[P3 #6, #7]** Rc-check `mv` and `chmod`
6. **[P3 #8, #10, #11]** Add `[ -d $out_path ]` guard before `mv`; split missing-files tests; add write-failure test
7. Other P3s as time permits

Findings [1]–[5] are the merge blockers. P3s are hygiene improvements the maintainer can defer.

🤖 Generated with `ce:review` skill (9 reviewers, 51 findings synthesized + deduped). Reviewers were given full diff via `/tmp/pr135.diff` and PR body. No code mutated.

mairas:

Fix commit: [`ca641c0`](https://github.com/halos-org/halos-core-containers/commit/ca641c0).

Tests: **60/60 pass** (44 pre-existing + 16 new). Syntax checks clean.

### P1

| # | Finding | Resolution |
|---|---|---|
| 1 | `set -e` cascade: unguarded helper call aborts whole web stack | ✅ Wrapped call site in `prestart.sh` with `if ! ...; then echo WARN; fi`. Cockpit failure now logs a WARNING and prestart continues; Traefik/Authelia/Homarr come up regardless. Matches the documented "safety-net floor" design. |

### P2

| # | Finding | Resolution |
|---|---|---|
| 2 | Silent chown failure leaves cockpit unable to read key | ✅ Dropped `\|\| true`. When the `cockpit-ws` group exists, chown failure now rm's the `.new`, logs loudly, returns rc=1. Test: `test_cockpit_install_leaf_rc1_when_chown_fails_with_group_present`. |
| 3 | Parse-failure branches untested | ✅ Added `test_cockpit_install_leaf_rejects_malformed_cert` and `test_cockpit_install_leaf_rejects_malformed_key`. Both drive the parse-fail branches in isolation. |
| 4 | `getent` group branches entirely untested | ✅ Added `test_cockpit_install_leaf_logs_notice_when_group_missing` and `test_cockpit_install_leaf_rc1_when_chown_fails_with_group_present` — function-shadow `getent()` inside subshells to drive both paths and assert NOTICE-on-stderr + correct rc. |
| 5 | Missing symlink defense on `.new` write path | ✅ Added `rm -f "$out_new"` before the `cat` redirection. Test: `test_cockpit_install_leaf_no_symlink_follow_on_new` plants a symlink at the stage path and asserts the canary is untouched. |

### P3

| # | Finding | Resolution |
|---|---|---|
| 6 | `mv` failure not rc-checked / `.new` debris | ✅ `mv` is now rc-checked; rm's the `.new` and returns 1 on failure. |
| 7 | `chmod 0640` failure on `.new` not rc-checked | ✅ Rc-checked; rm's the `.new` and returns 1 on failure. |
| 8 | `mv` onto pre-existing directory silently succeeds | ✅ Added `[ -d "$out_path" ]` guard before mv. Test: `test_cockpit_install_leaf_refuses_directory_at_out_path`. |
| 9 | Concurrent prestart races on shared `.new` filename | ✅ PID-suffixed stage filename (`${out_path}.new.$$`). All debris-checking tests updated to glob `.new.*`. |
| 10 | Missing-files test doesn't isolate cert vs key branch | ✅ Added `test_cockpit_install_leaf_missing_cert_only_isolates_branch` and `test_cockpit_install_leaf_missing_key_only_isolates_branch`. Both assert specific diagnostic text. |
| 11 | Write-failure branch untested (missing parent dir) | ✅ Added `test_cockpit_install_leaf_write_failure_with_missing_parent`. |
| 12 | No re-prestart trigger after late cockpit-ws install | ⏭ Skipped — design discussion (apt trigger or path unit); out of scope for this PR. |
| 13 | Validate-on-`$out_new` rationale undocumented | ✅ Inline comment added explaining that validating the output (not the inputs) catches concat tearing and short-writes. |
| 14 | Prestart skip-branch untested (integration only) | ⏭ Skipped — release-stage verification on the test device; no bash-level harness exists for prestart.sh end-to-end. |

### Other

- **`docs/CERTS.md`**: extended to contain both the Cockpit cert-sharing section (this PR) and the CA download endpoint section (PR #136). Byte-identical to the file in PR #136, so the add/add merge conflict between the two PRs resolves to "no change" rather than risking content loss.

### Residual risks (no code change; documented)

The original review's residual-risks section (two key copies on disk, cockpit-tls fail-closed assumption, etc.) stands. These are design choices, not findings.

Issue #131, opened 2026-05-24
Custom CA drop slot + active-CA selection + serving symlink


mairas:
Sub-issue of #129 (Unit 2 of 4).

## Scope

Operator-installable custom CA, with explicit precedence over the auto-generated CA.

### Files

| Path | Owner | Notes |
|---|---|---|
| `/etc/halos/ca/ca.{crt,key}` | operator | Optional drop slot. Debian conffile in `/etc/halos/`; presence + validity = "custom CA in use". |
| `/var/lib/container-apps/halos-core-containers/certs/ca/serving-ca.crt` | system | Symlink to the active CA cert. Mounted by Traefik for the download URL (Unit 4). |

### Validation gate

Before treating a custom CA as active, validate:
- Both `ca.crt` and `ca.key` parse cleanly
- `ca.crt` has `basicConstraints CA:TRUE`
- `ca.crt` has `keyUsage keyCertSign`
- `ca.key` matches `ca.crt` (public-key derivation comparison)
- `ca.crt` not expired (reuse `_halos_ca_is_healthy` from #130)

On any validation failure: log loud to journalctl ERROR, prestart exits non-zero (operator-visible via `systemctl status`), **no silent fallback to auto-CA**. Operators capable of installing a custom CA can SSH to fix it; visibility of failure beats stealth recovery (see #130 finding philosophy).

### Active-CA selection

In `halos_ca_ensure_auto` (or a new `halos_ca_select_active` helper):
1. If `/etc/halos/ca/ca.{crt,key}` both exist + validate → active = custom
2. Else → ensure auto-CA exists at `/var/lib/.../certs/ca/ca.{crt,key}` (current behavior) → active = auto
3. Update `serving-ca.crt` symlink (atomic via `ln -sfn`) to point at active `.crt`

### Re-sign trigger

The sentinel introduced in #130 already incorporates the CA fingerprint. Swapping in/out a custom CA produces a different fingerprint, which triggers leaf re-sign automatically. No sentinel changes needed.

### Discoverability

```
test -f /etc/halos/ca/ca.crt && echo "custom CA active" || echo "auto CA active"
```

Documented in `docs/CERTS.md` (filed separately).

## Tests to add

- Custom CA present + valid → active = custom; symlink targets custom
- Custom CA present + invalid (missing CA:TRUE) → prestart fails non-zero, symlink unchanged
- Custom CA present + key/cert mismatch → prestart fails non-zero
[...]

Issue #19, opened 2025-12-19
feat(sso): Create Authelia container package


mairas:
## Description

Create `apps/authelia/` with OIDC provider configuration for HaLOS SSO.

## Implementation Requirements

**MANDATORY**: Follow `docs/IMPLEMENTATION_CHECKLIST.md` during implementation.

This includes:
- Exploration phase WITHOUT coding
- Planning with "think hard"
- Test-Driven Development (write tests first)
- Verification with subagents
- References to SSO_SPEC.md and SSO_ARCHITECTURE.md

## Implementation Details

Create the following files:
- `docker-compose.yml` with authelia/authelia:4.39 image, joining `halos-proxy-network` (external: true)
- `configuration.yml.template` with:
  - File backend for user storage (`/data/users_database.yml`)
  - OIDC provider with Homarr client definition
  - Session configuration for `halos.local` domain
  - Access policy: `one_factor`
- `prestart.sh` to generate secrets:
  - Session secret (random hex)
  - OIDC HMAC secret (random hex)
  - OIDC private key (RSA 4096)
- `metadata.yaml` with `field::core` tag
- Traefik labels for `auth.halos.local` routing
- `halos.subdomain=auth` label for mDNS

OIDC client configuration for Homarr:
- client_id: `homarr`
- redirect_uris: `http://halos.local/api/auth/callback/oidc`
- scopes: openid, profile, groups, email
- consent_mode: implicit

## Test Scenarios

- Secrets generated on first start
- Container joins proxy network successfully
- Login page accessible at auth.halos.local (after mDNS is working)
- OIDC discovery endpoint (`/.well-known/openid-configuration`) returns valid config
- Login with test credentials works

## Acceptance Criteria

- [ ] Package builds with container-packaging-tools
- [ ] Secrets auto-generated on first boot
- [ ] OIDC discovery endpoint responds correctly
- [ ] Login with test credentials works
- [ ] Matches SSO_SPEC.md requirements
- [ ] Follows SSO_ARCHITECTURE.md patterns

## Dependencies

- Depends on: #18 (Traefik container package)

## References

- SSO_SPEC.md
- SSO_ARCHITECTURE.md
- Authelia Configuration: https://www.authelia.com/configuration/
- Authelia OIDC: https://www.authelia.com/configuration/identity-providers/oidc/

PR #136, opened 2026-05-25
feat(certs): serve /halos-ca.crt download endpoint via Traefik

### LINKED ISSUE #129: Add custom-CA support with shared Cockpit leaf

### LINKED ISSUE #133: Serve /halos-ca.crt download endpoint via Traefik

mairas:
## Why

Operators today have no straightforward path to fetch + install the device CA into their OS/browser trust store. After Unit 2 (#131, merged as #134) introduced the per-device CA, the umbrella plan (#129) calls for serving the active CA at `https://<host>/halos-ca.crt` so operators can install one trust anchor for all services on this device.

Option A from the issue was chosen — a tiny static-file sidecar over piggybacking on Homarr's nginx or overloading Cockpit's `:9090`. Smallest blast radius, clearest ownership, trust-anchor distribution doesn't depend on dashboard availability.

## How

**prestart.sh** copies the active CA's PEM bytes to `${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/public/halos-ca.crt` via atomic `.new` + `mv`. Two reasons we copy instead of mounting `serving-ca.crt` directly:

1. Docker resolves symlinks at mount time on the host — when the active CA is the operator's custom CA, the symlink target (`/etc/halos/ca/ca.crt`) doesn't exist inside the sidecar's mount namespace.
2. The custom-CA directory ALSO holds `ca.key`. Mounting that dir into a public-facing sidecar (even read-only) would expose the operator's private key.

The copy refreshes every prestart, BEFORE the leaf-signing path, so the published bytes track whichever CA `halos_ca_select_active` picked — even on boots that don't re-sign the leaf. Mode `0644` (this is a public trust anchor; world-readable is correct).

**ca-download** sidecar — `nginx:1.27-alpine` (~10MB compressed). Bind-mounts the public dir read-only and a 38-line `nginx.conf` serving exactly one path with:

- `Content-Type: application/x-x509-ca-cert`
- `Content-Disposition: attachment; filename="halos-ca.crt"` (`attachment` is the load-bearing piece — triggers the OS cert-install dialog instead of rendering the PEM as text)
- `/healthz` for the docker healthcheck; 404 for everything else

**Traefik labels** route the exact `/halos-ca.crt` path at priority 200 so it beats Homarr's catch-all `PathPrefix(/)` at priority 1. HTTP → HTTPS via the existing `redirect-to-https@file` middleware. No host port published, no Authelia ForwardAuth — public by design (the chicken-and-egg: trust the warning once or verify by SSH).

**debian/postinst** pre-creates the public directory so the bind-mount has a target before first prestart, matching the `AUTO_CA_DIR` layout convention.

## Tests

- `bash tests/test-lib-ca.sh` — 44 passed, 0 failed (lib-ca untouched).
- `bash -n prestart.sh` — syntax OK.
- `docker compose -f docker-compose.yml config` — validates cleanly with the new service.

Sidecar/route behavior is integration-shaped; the test plan with curl one-liners (cert-match, headers, HTTP redirect, swap-CA reflow, catch-all unaffected) is in `docs/CERTS.md` for on-device QA.

## Notes

- VERSION not bumped: latest stable tag is `v0.3.2+1`; VERSION is already `0.3.4` (ahead) so no further bump needed per workspace policy.
- `docs/CERTS.md` is a new file — #135 (Unit 3) adds its own "Cockpit :9090 cert sharing" section to the same file, so whichever of these two PRs lands second will hit a clean merge conflict (additive sections, easy resolve).
- This is Unit 4 of 4 in #129. Unit 3 is #132 / PR #135. When both land, #129 should be closed.

Closes #129, closes #133.

### DIFF
--- a/assets/ca-download/nginx.conf
+++ b/assets/ca-download/nginx.conf
@@ -18,10 +18,12 @@ http {
     server {
         listen 80 default_server;
 
-        # /healthz answers the docker healthcheck. Cheap, no file I/O.
+        # /healthz is kept as a cheap liveness probe (no file I/O) for any
+        # external monitoring that wants it. The docker healthcheck probes
+        # /halos-ca.crt instead so a missing/unreadable cert marks the
+        # container unhealthy — see docker-compose.yml.
         location = /healthz {
             return 200 "ok\n";
-            default_type text/plain;
         }
 
         # The active CA copy is bind-mounted at /srv/halos-ca.crt
@@ -33,6 +35,17 @@ http {
             # `attachment` triggers the OS/browser cert-install dialog instead
             # of rendering the PEM as text.
             add_header Content-Disposition 'attachment; filename="halos-ca.crt"' always;
+            # Defense-in-depth: if /srv/halos-ca.crt ever becomes a symlink
+            # (e.g., a future prestart refactor sharing the serving-ca.crt
+            # symlink pattern from lib-ca.sh), nginx must not follow it
+            # outside the bind-mount.
+            disable_symlinks on;
+            # Read-only endpoint by design — refuse anything that isn't
+            # GET/HEAD with a clean 405 + Allow header rather than silently
+            # serving the file body for POST/PUT/DELETE.
+            limit_except GET HEAD {
+                deny all;
+            }
         }
 
         # Everything else: 404.
--- a/assets/ca-download/nginx.conf
+++ b/assets/ca-download/nginx.conf
@@ -1,0 +1,43 @@
+# nginx config for the ca-download sidecar.
+#
+# Serves exactly one URL — /halos-ca.crt — with the headers browsers expect
+# for a CA-install prompt. Traefik fronts TLS termination on :443 and routes
+# the path here over the proxy network; this container listens on plain :80
+# only and is never exposed on the host.
+
+worker_processes 1;
+
+events {
+    worker_connections 256;
+}
+
+http {
+    access_log off;
+    server_tokens off;
+
+    server {
+        listen 80 default_server;
+
+        # /healthz answers the docker healthcheck. Cheap, no file I/O.
+        location = /healthz {
+            return 200 "ok\n";
+            default_type text/plain;
+        }
+
+        # The active CA copy is bind-mounted at /srv/halos-ca.crt
+        # (prestart.sh refreshes it from ${HALOS_CA_ACTIVE_CRT} each run).
+        # `alias` (not `root`) pins the location to that exact file.
+        location = /halos-ca.crt {
+            alias /srv/halos-ca.crt;
+            default_type application/x-x509-ca-cert;
+            # `attachment` triggers the OS/browser cert-install dialog instead
+            # of rendering the PEM as text.
+            add_header Content-Disposition 'attachment; filename="halos-ca.crt"' always;
+        }
+
+        # Everything else: 404.
+        location / {
+            return 404;
+        }
+    }
+}
--- a/assets/lib-ca.sh
+++ b/assets/lib-ca.sh
@@ -15,6 +15,8 @@
 #                                              pick custom-if-valid / else auto, update symlink
 #   - halos_cockpit_install_leaf <leaf_crt> <leaf_key> <output_path>
 #                                              install combined PEM override for cockpit-tls
+#   - halos_ca_publish_public <src_crt> <public_dir>
+#                                              copy active CA into a public-bindmount target
 #
 # Generated certs carry the extensions browsers + OS trust stores require:
 #   CA   — basicConstraints CA:TRUE (so importing as trust anchor actually works),
@@ -569,6 +571,89 @@ halos_cockpit_install_leaf() {
     if ! mv "$out_new" "$out_path"; then
         rm -f "$out_new"
         echo "halos_cockpit_install_leaf: failed to mv $out_new to $out_path; previous override (if any) is preserved" >&2
+        return 1
+    fi
+}
+
+# halos_ca_publish_public <src_crt> <public_dir>
+# Atomically refresh a world-readable copy of <src_crt> at
+# <public_dir>/halos-ca.crt, mode 0644. Used by prestart to publish the active
+# CA into a bind-mount target the ca-download sidecar serves over HTTP.
+#
+# Why a copy and not a direct mount of the serving-ca.crt symlink:
+#   1. The symlink's absolute target (e.g. /etc/halos/ca/ca.crt when a custom
+#      CA is in use) does not exist inside the sidecar's mount namespace —
+#      Docker resolves the symlink at mount time on the host.
+#   2. The custom-CA directory ALSO holds ca.key. Mounting that directory into
+#      a public-facing sidecar (even read-only) would expose the operator's
+#      private key. We copy only the certificate.
+#
+# Mode 0644 is intentional: this file is a public trust anchor, world-readable
+# is correct.
+#
+# Failure semantics: on any failure leaves the pre-existing public copy (if
+# any) untouched, removes the .new sibling, returns non-zero.
+#
+# Return codes:
+#   0 — success
+#   1 — runtime failure (missing source, write/chmod/mv error, parse fail)
+#   2 — caller bug (missing args)
+halos_ca_publish_public() {
+    local src_crt="$1" public_dir="$2"
+    if [ -z "$src_crt" ] || [ -z "$public_dir" ]; then
+        echo "halos_ca_publish_public: <src_crt> <public_dir> both required" >&2
+        return 2
+    fi
+    if [ ! -f "$src_crt" ]; then
+        echo "halos_ca_publish_public: source cert missing: $src_crt" >&2
+        return 1
+    fi
+
+    if ! mkdir -p "$public_dir"; then
+        echo "halos_ca_publish_public: failed to mkdir $public_dir" >&2
+        return 1
+    fi
+    # Defense-in-depth: re-tighten the dir mode on every call so an operator
+    # who narrowed it (e.g., debugging) doesn't break the bind-mount.
+    chmod 0755 "$public_dir" 2>/dev/null || true
+
+    local public_file="${public_dir}/halos-ca.crt"
+    # PID-suffixed stage filename so concurrent prestart invocations don't
+    # collide on the same staging path. The mv into public_file is still
+    # atomic per caller.
+    local public_new="${public_file}.new.$$"
+    # rm before cp so `>` inside cp doesn't follow a pre-existing symlink at
+    # the stage path. (cp itself doesn't follow target symlinks by default
+    # for overwrite, but the per-pid name minimizes the surface anyway.)
+    rm -f "$public_new"
+
+    if ! cp "$src_crt" "$public_new"; then
+        rm -f "$public_new"
+        echo "halos_ca_publish_public: failed to cp $src_crt to $public_new" >&2
+        return 1
+    fi
+    # Verify the copy parses as an X.509 cert. Catches truncation, EIO mid-
+    # cp, and the case where src_crt drifted to something that isn't a cert.
+    if ! openssl x509 -in "$public_new" -noout 2>/dev/null; then
+        rm -f "$public_new"
+        echo "halos_ca_publish_public: copied file at $public_new does not parse as an X.509 certificate" >&2
+        return 1
+    fi
+    if ! chmod 0644 "$public_new"; then
+        rm -f "$public_new"
+        echo "halos_ca_publish_public: chmod 0644 failed on $public_new" >&2
+        return 1
+    fi
+    # Refuse to install over a pre-existing directory at public_file (plain
+    # mv would silently move INTO the directory).
+    if [ -d "$public_file" ]; then
+        rm -f "$public_new"
+        echo "halos_ca_publish_public: refusing to install over directory at $public_file" >&2
+        return 1
+    fi
+    if ! mv "$public_new" "$public_file"; then
+        rm -f "$public_new"
+        echo "halos_ca_publish_public: failed to mv $public_new to $public_file; previous copy (if any) is preserved" >&2
         return 1
     fi
 }
--- a/debian/halos-core-containers.docs
+++ b/debian/halos-core-containers.docs
@@ -1,1 +1,2 @@
 docs/HOSTNAMES.md
+docs/CERTS.md
--- a/debian/postinst
+++ b/debian/postinst
@@ -24,6 +24,12 @@ EOF
         mkdir -p "${DATA_DIR}/homarr/configs"
         mkdir -p "${DATA_DIR}/homarr/data"
         mkdir -p "${DATA_DIR}/homarr/icons"
+        # Public CA bind-mount target for the ca-download sidecar.
+        # Matches AUTO_CA_DIR's "${DATA_DIR}/${PKG_NAME}/certs/..." layout.
+        # prestart.sh also ensures this directory; created here so the
+        # docker-compose bind mount has a target on first boot before any
+        # prestart run.
+        mkdir -p "${DATA_DIR}/${PKG_NAME}/certs/public"
 
         # Create routing directories used by other container apps
         mkdir -p /etc/halos/routing.d
--- a/debian/postinst
+++ b/debian/postinst
@@ -11,10 +11,17 @@ case "$1" in
         # Create configuration directory if it doesn't exist
         mkdir -p "${ETC_DIR}"
 
-        # Create env.defaults with CONTAINER_DATA_ROOT
+        # Create env.defaults with the canonical package layout. PACKAGE_NAME
+        # is exported here (not derived at runtime from $(basename $SCRIPT_DIR)
+        # in prestart.sh) so docker-compose.yml and prestart agree on every
+        # bind-mount path regardless of where prestart was invoked from. This
+        # prevents the dev-time drift where running prestart out of a worktree
+        # (.worktrees/<branch>/) would otherwise compute a different
+        # PACKAGE_NAME than the one compose hardcodes.
         cat > "${ETC_DIR}/env.defaults" << EOF
 # System-managed environment variables (do not modify)
 CONTAINER_DATA_ROOT="${DATA_DIR}"
+PACKAGE_NAME="${PKG_NAME}"
 EOF
 
         # Create data directory structure
@@ -25,7 +32,7 @@ EOF
         mkdir -p "${DATA_DIR}/homarr/data"
         mkdir -p "${DATA_DIR}/homarr/icons"
         # Public CA bind-mount target for the ca-download sidecar.
-        # Matches AUTO_CA_DIR's "${DATA_DIR}/${PKG_NAME}/certs/..." layout.
+        # Mirrors prestart's certs/{ca,public} subdir layout under PKG_NAME/.
         # prestart.sh also ensures this directory; created here so the
         # docker-compose bind mount has a target on first boot before any
         # prestart run.
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -177,6 +177,46 @@ services:
       options:
         tag: "{{.Name}}"
 
+  ca-download:
+    # Sidecar that publishes the active device CA at /halos-ca.crt so
+    # operators can install one trust anchor for all services on this device.
+    # Routed by Traefik (no host port published). prestart.sh copies the
+    # active CA's PEM bytes into the mounted dir each run.
+    image: nginx:1.27-alpine
+    container_name: ca-download
+    init: true
+    restart: unless-stopped
+    volumes:
+      - ${CONTAINER_DATA_ROOT}/halos-core-containers/certs/public:/srv:ro
+      - ./assets/ca-download/nginx.conf:/etc/nginx/nginx.conf:ro
+    healthcheck:
+      test: ["CMD", "wget", "-q", "--spider", "http://localhost/healthz"]
+      interval: 10s
+      timeout: 5s
+      start_period: 5s
+      retries: 3
+    networks:
+      - proxy
+    labels:
+      - traefik.enable=true
+      # HTTP — exact-path match, redirected to HTTPS. Priority 200 beats
+      # homarr's catch-all PathPrefix(`/`) at priority 1.
+      - traefik.http.routers.ca-download-http.rule=Path(`/halos-ca.crt`)
+      - traefik.http.routers.ca-download-http.entrypoints=web
+      - traefik.http.routers.ca-download-http.middlewares=redirect-to-https@file
+      - traefik.http.routers.ca-download-http.priority=200
+      # HTTPS — exact-path match. No ForwardAuth: this is a public trust
+      # anchor; gating it would defeat the chicken-and-egg bootstrap.
+      - traefik.http.routers.ca-download-secure.rule=Path(`/halos-ca.crt`)
+      - traefik.http.routers.ca-download-secure.entrypoints=websecure
+      - traefik.http.routers.ca-download-secure.tls=true
+      - traefik.http.routers.ca-download-secure.priority=200
+      - traefik.http.services.ca-download.loadbalancer.server.port=80
+    logging:
+      driver: journald
+      options:
+        tag: "{{.Name}}"
+
   autoheal:
     image: willfarrell/autoheal:1.2.0
     container_name: autoheal
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -187,13 +187,22 @@ services:
     init: true
     restart: unless-stopped
     volumes:
-      - ${CONTAINER_DATA_ROOT}/halos-core-containers/certs/public:/srv:ro
+      # PACKAGE_NAME is normally injected by systemd via env.defaults
+      # (debian/postinst writes it there). The :-halos-core-containers
+      # fallback keeps `docker compose` invocations outside systemd (e.g.,
+      # local dev with `docker compose config`) working with the canonical
+      # path layout.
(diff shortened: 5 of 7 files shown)


mairas:
## ce:review (multi-persona)

**Scope:** `feat/ca-download-endpoint` vs `main` — 6 files, +197 lines (`prestart.sh`, `assets/ca-download/nginx.conf` new, `docker-compose.yml`, `debian/postinst`, `debian/halos-core-containers.docs`, `docs/CERTS.md` new).

**Mode:** report-only (no autofix applied; findings posted for the author's policy call).

**Intent:** Serve active per-device CA at `https://<host>/halos-ca.crt` via `nginx:1.27-alpine` sidecar. prestart copies active CA to a public dir (not a symlink mount — would expose operator's `ca.key` or fail to resolve inside the sidecar). Sidecar bind-mounts public dir RO. Traefik routes `Path(/halos-ca.crt)` at priority 200 (beats Homarr's catch-all at 1). HTTP→HTTPS redirect via existing `redirect-to-https@file`. No host port published; no Authelia ForwardAuth (public by design — chicken-and-egg trust bootstrap).

**Reviewer team (10):** correctness, testing, maintainability, project-standards, agent-native, learnings, security, reliability, adversarial, api-contract.

---

### P2 — High

|---|---|---|---|---|---|
| 1 | `docs/CERTS.md:12, 75` | HTTP→HTTPS redirect docs claim `308` but Traefik `redirectScheme` with `permanent: true` emits **`301`** | api-contract | 0.88 | safe_auto → review-fixer |
| 2 | `docker-compose.yml:193` + `assets/ca-download/nginx.conf:22-26` | Healthcheck on `/healthz` doesn't verify served CA file — container stays healthy while `/halos-ca.crt` 404s | correctness, reliability (2-way) | 0.95 | gated_auto → downstream-resolver |
| 3 | `docker-compose.yml:190` + `prestart.sh:135` | Path drift: compose hardcodes `halos-core-containers`, prestart derives `$PACKAGE_NAME` from `$(basename $SCRIPT_DIR)` | correctness, maintainability, adversarial (3-way) | 0.95 | manual → downstream-resolver |
| 4 | `docs/CERTS.md` | Add/add conflict with PR #135 — both create the same file from scratch; merge can silently drop content | maintainability | 0.88 | manual → downstream-resolver |
| 5 | `prestart.sh:135-141` | Unguarded `cp` under `set -e` makes public-CA refresh failure fatal to entire web stack | reliability | 0.80 | gated_auto → review-fixer |
| 6 | `prestart.sh:135-141` | Inline cp/chmod/mv block has no unit test even though it's pure shell I/O | testing | 0.85 | manual → review-fixer |
| 7 | `prestart.sh:120-141` | Swap-CA scenario (the central correctness claim — "tracks active CA across reboots") only manually QA'd | testing | 0.78 | manual → review-fixer |

**Detail [1]:** Traefik's `redirectScheme` middleware with `permanent: true` returns `301 Moved Permanently`. `308` is exclusive to `redirectRegex` and only when explicitly configured. Test command #3 in `docs/CERTS.md` (`# 308 Permanent Redirect`) will fail. Fix: docs edit `308` → `301` (two places).

**Detail [2]:** `/healthz` returns 200 unconditionally; doesn't read the bind-mounted file. If `/srv/halos-ca.crt` is absent (prestart never ran, was killed mid-cp, manual cleanup), nginx 404s the cert path while the container reports healthy and autoheal doesn't restart it. Operators discover failure via downstream user reports. Fix: probe the actual served path (`wget --spider http://localhost/halos-ca.crt`); slightly slower but file-presence-aware.

**Detail [3]:** Three independent spellings of the same path: compose hardcodes the literal, prestart derives via basename, postinst uses `PKG_NAME="halos-core-containers"`. In production prestart runs from `/var/lib/container-apps/halos-core-containers/` so basename matches — but in the worktree layout used to author this PR, basename = `ca-download-endpoint`, so prestart writes to a different dir than compose mounts. Sidecar serves stale or empty file with no error. Same drift affects `AUTO_CA_DIR` (pre-existing). Fix: pin `PACKAGE_NAME` in prestart from `/etc/container-apps/halos-core-containers/env.defaults` (already populated by postinst), or expose it via that env so compose can use `${PACKAGE_NAME}`.

**Detail [4]:** Both this PR and PR #135 create `docs/CERTS.md` from scratch with the same `# Certificates` H1. Git's add/add conflict won't show as a line-by-line content conflict, so a careless merge can silently drop content from one PR. Fix: pre-merge content from PR #135's branch into this one (or vice versa), or rebase the second-to-land onto the first so it becomes an append rather than an add/add.

**Detail [5]:** Same pattern as PR #135's P1. `cp "${HALOS_CA_ACTIVE_CRT}" "${PUBLIC_CA_FILE}.new"` is unguarded under `set -e`. A transient FS hiccup, EIO on the data partition, or unreadable active CA during mid-edit aborts prestart → Traefik/Authelia/Homarr all stay down. Fix: wrap with `|| { echo "WARN: failed to refresh public CA copy; download endpoint may serve stale or no CA" >&2; }`.

**Detail [6]:** The 7-line block has no Docker/network/openssl dependencies and is trivially shell-testable. Fix: extract to `halos_ca_publish_public <src_crt> <public_dir>` in `lib-ca.sh` and add tests for mode 0644, parent-dir creation, atomic replace, swap-CA tracking.

**Detail [7]:** The "publish before leaf-signing path" ordering is the PR's central behavioral contract. `docs/CERTS.md` test #4 is the manual QA, but no automated test catches a future refactor that moves the publish inside an `if NEED_LEAF` block.

---

### P3 — Low

|---|---|---|---|---|---|
| 8 | `assets/ca-download/nginx.conf:30-37` | No HTTP method gating — POST/PUT/DELETE return 200 with PEM body | api-contract | 0.85 | safe_auto → review-fixer |
| 9 | `docker-compose.yml:185` | `nginx:1.27-alpine` pinned only by minor tag, not digest; other services in compose pin patch versions | security | 0.75 | manual → human |
| 10 | `docker-compose.yml:189` | Directory bind-mount exposes whatever's in `certs/public/`; today only protected by `location =` exact-match | security | 0.70 | manual → downstream-resolver |
| 11 | `docker-compose.yml` ordering + nginx.conf | First-boot window: sidecar serves bare 404 before prestart's first cp completes (openssl req on Pi can take 10-60s) | adversarial | 0.75 | advisory → human |
| 12 | `assets/ca-download/nginx.conf:36` | nginx `alias` follows symlinks unconditionally; no `disable_symlinks on` | adversarial | 0.65 | manual → downstream-resolver |
| 13 | `docs/CERTS.md` Test commands | Re-use `curl -sk` without per-command re-emphasis of fingerprint compare | adversarial | 0.70 | manual → downstream-resolver |
| 14 | `assets/ca-download/nginx.conf` | Header/404 contracts only validated by operator curl on flashed device | testing | 0.70 | manual → downstream-resolver |
| 15 | `prestart.sh:139-141` | Stale `.new` files if process killed between cp and mv (self-healing on next prestart) | reliability | 0.70 | safe_auto → review-fixer |
| 16 | `prestart.sh:138-141` | Concurrent prestart invocations race on shared `PUBLIC_CA_FILE.new` filename | adversarial | 0.60 | manual → downstream-resolver |
| 17 | `assets/ca-download/nginx.conf:22-25` | `default_type text/plain;` after `return 200 "ok\n";` is a no-op (cosmetic) | api-contract | 0.78 | safe_auto → review-fixer |
| 18 | `prestart.sh:135-141` | Inline publish block diverges from sibling PR #135's `halos_cockpit_install_leaf` helper pattern | maintainability | 0.62 | advisory → human |
| 19 | `debian/postinst:27` | Comment references `AUTO_CA_DIR` symbol defined only in prestart.sh, not in postinst's scope | maintainability | 0.60 | safe_auto → review-fixer |

---

### Residual risks (no concrete fix; documented or operational)

- **Chicken-and-egg MITM** during the very first cert download — intrinsic to first-time-trust bootstrap. Documented in `docs/CERTS.md`; mitigation (SSH fingerprint compare) is not enforced, but operators following the recipe verbatim on an untrusted LAN may skip step 2.
- **`nginx:1.27-alpine`** floating minor tag — consistent with current stack posture (all services pin tags, not digests). A device-wide digest-pinning policy is a larger conversation.
- **Subject DN reveals trust mode** — "HaLOS Device CA" vs operator's custom CN. Acceptable info disclosure for a public trust anchor.
- **Sidecar on shared `proxy` network** — any CVE-enabled RCE in nginx 1.27 grants lateral access to traefik/authelia/homarr. No new attack surface relative to existing services on the same network.

---

- Path-only Traefik routing with explicit priority over Homarr's catch-all matches the project's documented architecture (`AGENTS.md`: "all per-app routers are now path-only"). The priority-200-beats-1 pattern is the canonical way to add new path routes.
- No prior solution doc on nginx sidecars in `halos-core-containers` — this is the first. After merge, consider a `ce:compound` note for the priority-N pattern.
- This PR's changes are **NOT in the swap-safe envelope** (`docs/solutions/best-practices/2026-05-04-image-swap-smoke-test-without-deb-rebuild.md`) — full `./run build` + `./run deploy` cycle is required because the diff adds a sidecar, postinst mkdir, and prestart logic.
- Pre-pull `nginx:1.27-alpine` on the test device before the first `systemctl restart` to avoid blocking the core-containers stack on a registry pull.

---

### Agent-native

**PASS.** The endpoint is itself an agent-native primitive: stateless GET, deterministic content, no UI gating, no auth ceremony. Verification flow (`curl -k` + ssh fingerprint compare) is scriptable; the swap-CA test in `docs/CERTS.md` is a scriptable regression check.

Nice-to-have: a `halos-ca fingerprint` CLI helper would abstract the on-device absolute path so agent bootstrap doesn't depend on path stability.

---

### Project standards

Clean. Conventional commit, branch name, VERSION (already ahead per workspace policy), `debian/changelog` discipline, `Refs #129, closes #133` umbrella linkage, hostname policy (`halos.local` is permitted in `.md` docs per `docs/HOSTNAME_POLICY.md`), docker-compose conventions all pass.

---

### Verdict

**Ready with fixes.** Recommended order:

1. **[P2 #1]** Docs edit `308` → `301` (otherwise the documented test fails)
2. **[P2 #2]** Healthcheck probe `/halos-ca.crt` (or accept the decoupling explicitly with a comment)
3. **[P2 #3]** Resolve `PACKAGE_NAME` drift between compose and prestart — 1-2 line fix
4. **[P2 #4]** Coordinate `docs/CERTS.md` merge with PR #135 (rebase or pre-merge content)
5. **[P2 #5]** Wrap prestart cp/mv with `|| WARN` — same shape as #135's P1
6. **[P2 #6, #7]** Add unit tests for publish logic and swap-CA scenario (extract to lib-ca.sh helper)
7. **[P3 #8, #15, #17, #19]** Quick safe-auto cleanups (method gating, `.new` cleanup at function entry, drop dead `default_type`, postinst comment)
8. Other P3s as time permits

Findings [1]–[7] are the merge blockers. P3s are hygiene/defense-in-depth the maintainer can defer per scope-creep risk.

🤖 Generated with `ce:review` skill (10 reviewers, structured findings synthesized + deduped). Reviewers were given full diff via `/tmp/pr136.diff` and PR body. No code mutated.

mairas:

Fix commit: [`fe09d17`](https://github.com/halos-org/halos-core-containers/commit/fe09d17).

Tests: **53/53 pass** (44 pre-existing + 9 new). `docker compose config` validates with and without `PACKAGE_NAME` in the environment.

### P2

| # | Finding | Resolution |
|---|---|---|
| 1 | docs claim 308 but Traefik `redirectScheme(permanent: true)` emits 301 | ✅ `docs/CERTS.md` corrected (two places). Documented test command #3 will now pass. |
| 2 | Healthcheck on `/healthz` doesn't verify the served CA file | ✅ Changed `docker-compose.yml` healthcheck to probe `/halos-ca.crt` directly. Missing/unreadable file now marks the container unhealthy and autoheal restarts it. `start_period` bumped to 30s to absorb first-boot openssl-req latency on slow ARM. `/healthz` kept in nginx.conf for any external liveness probes. |
| 3 | `PACKAGE_NAME` drift: compose hardcodes literal, prestart derives via basename | ✅ `debian/postinst` writes `PACKAGE_NAME=halos-core-containers` to env.defaults; `docker-compose.yml` uses `${PACKAGE_NAME:-halos-core-containers}`. Works in systemd (env.defaults loaded) and in local `docker compose config` (fallback applies). Verified both paths render `/tmp/halos-core-containers/certs/public`. |
| 4 | `docs/CERTS.md` add/add conflict with PR #135 | ✅ Both PRs now contain the full union of cockpit-sharing + download-endpoint sections, byte-identical. Merge of the second PR will see "no change" on `docs/CERTS.md` instead of an add/add conflict. |
| 5 | Unguarded `cp` under `set -e` blocks whole container stack | ✅ Same shape as PR #135's P1. Inline cp/mv extracted to new helper `halos_ca_publish_public`; call site wrapped in `if ! ...; then echo WARN; fi`. Previous public copy is preserved by the helper on failure. |
| 6 | Public-CA publish logic has no unit test | ✅ Extracted to `halos_ca_publish_public` helper in `assets/lib-ca.sh`. Added 9 tests: happy path, parent-dir creation + mode, missing args (rc=2), missing source (rc=1), non-cert source rejection, preserve-on-failure, directory-at-output refusal, no `.new` debris, plus the swap-CA case below. |
| 7 | Swap-CA scenario only manually QA'd | ✅ Added `test_publish_public_swap_updates_content`: publishes CA-A, then CA-B, asserts the served bytes change AND match the new source. The central correctness claim of the PR is now pinned. |

### P3

| # | Finding | Resolution |
|---|---|---|
| 8 | nginx serves all HTTP methods returning 200 with PEM | ✅ Added `limit_except GET HEAD { deny all; }` to the `/halos-ca.crt` location. Non-GET/HEAD now returns 405 with `Allow:` header. |
| 9 | `nginx:1.27-alpine` pinned only by minor tag | ⏭ Skipped — workspace-wide image-pinning policy decision (other services in this compose file also pin tags, not digests). Worth a separate cross-cutting PR if you want to flip the convention. |
| 10 | Directory bind-mount exposes whatever's in `certs/public/` | ⏭ Skipped — current `location =` exact-match + `disable_symlinks on` (now added) is the gate. Switching to a single-file bind mount has its own downsides (mv-changes-inode breaks the mount). Acceptable trade-off given the dir contract is "only halos-ca.crt may live here". |
| 11 | First-boot 404 window before first prestart cp | ⏭ Partially mitigated by bumping `start_period` to 30s. Full fix would need a systemd-level dependency change or a postinst placeholder file — out of scope. |
| 12 | nginx alias follows symlinks unconditionally | ✅ Added `disable_symlinks on;` to the `/halos-ca.crt` location. |
| 13 | docs test commands use `curl -sk` without per-command fingerprint preamble | ✅ Added a load-bearing preamble paragraph + "AFTER verifying the fingerprint per the chicken-and-egg procedure above" note above the Test commands code block. |
| 14 | nginx.conf header/404 contracts only validated manually | ⏭ Skipped — needs docker-based smoke-test infrastructure in CI not currently set up. Worth a follow-up. |
| 15 | Stale `.new` files if process killed between cp and mv | ✅ Helper `rm -f`s the per-pid stage file at entry and on every failure path. |
| 16 | Concurrent prestart races on shared `.new` filename | ✅ PID-suffixed stage filename in `halos_ca_publish_public` (`${public_file}.new.$$`). |
| 17 | Dead `default_type text/plain;` after `return` on /healthz | ✅ Removed. |
| 18 | Inline publish block diverges from sibling PR #135's helper pattern | ✅ Resolved by extracting `halos_ca_publish_public` (P2 #6). |
| 19 | postinst comment references `AUTO_CA_DIR` symbol defined only in prestart.sh | ✅ Comment rephrased to describe the layout directly. |

### Other

- New helper `halos_ca_publish_public` in `assets/lib-ca.sh` follows the same convention as the rest of the library: atomic `.new`+`mv`, parse-validation, dir-refusal, `rm -f` cleanup on every failure path, rc=2 for caller-bug args / rc=1 for runtime failures. Both this PR and PR #135 append-only to `lib-ca.sh`, so the merge between them is line-by-line additive (clean resolution).

### Residual risks (no code change; documented)

Image-tag pinning, sidecar lateral movement, Subject DN info disclosure, chicken-and-egg MITM during first download — all unchanged from the original review's residual-risks section.

PR #138, opened 2026-05-25
fix(certs): three deployment bugs in /halos-ca.crt endpoint

mairas:
Three deployment bugs caught while testing `v0.3.4-4` (the unstable build from PR #136) on `halosdev.local`. Each one independently broke the `/halos-ca.crt` endpoint despite passing CI and the multi-persona `ce:review`.

## 1. `debian/rules` didn't ship `assets/ca-download/nginx.conf`

The asset-install loop hardcoded `traefik authelia homarr` — the new `ca-download/` directory was silently skipped:

```
for dir in traefik authelia homarr; do ...
```

When the bind-mount source file is missing, Docker creates an empty *directory* at the source path. The bind into the container then fails with `not a directory`, and the container loops on init error before nginx even starts. The fix adds `ca-download` to the loop and a load-bearing comment so future `docker-compose.yml` bind-mount additions don't re-introduce the same gap.

## 2. Healthcheck `localhost` → IPv6 → unhealthy → invisible to Traefik

`docker-compose.yml` healthcheck probed `http://localhost/halos-ca.crt`. `nginx:alpine` resolves `localhost` to `::1` first, and our `nginx.conf` only `listen 80` (IPv4). Every probe got `ECONNREFUSED`, the container stayed unhealthy, and **Traefik 3.7's docker provider filters unhealthy containers from discovery** — so the `ca-download` routers never registered and `/halos-ca.crt` silently fell through to Homarr's catch-all (a Next.js 404).

Fix: probe `127.0.0.1` directly. Container goes healthy → Traefik registers the routers → end-to-end works.

## 3. `docs/CERTS.md` claimed 301; actual is 308

The `ce:review` api-contract reviewer claimed Traefik's `redirectScheme(permanent: true)` emits `301 Moved Permanently`. Traefik 3.7 actually emits `308 Permanent Redirect`. Verified on the device with `curl -sI`. Updated the two places that referenced the code.

## Verification

Tested on `halosdev.local` after side-loading the missing `nginx.conf` and patching the healthcheck in place:

```
$ curl -skI https://halosdev.local/halos-ca.crt | head -3
HTTP/2 200
content-disposition: attachment; filename="halos-ca.crt"
content-type: application/x-x509-ca-cert

$ diff <(ssh halosdev.local 'sudo cat .../serving-ca.crt') <(curl -sk https://halosdev.local/halos-ca.crt)
# (no output — bytes match)

$ curl -sI http://halosdev.local/halos-ca.crt | head -1
HTTP/1.1 308 Permanent Redirect

$ curl -skI -X POST https://halosdev.local/halos-ca.crt | head -1
HTTP/2 403   # blocked correctly (limit_except returns 403, not 405)
```

## Why CI didn't catch these

- `debian/rules` install loop: no test exercises the actual `.deb` install + first-boot flow. `docker compose config` passes the YAML.
- IPv6 healthcheck: no test brings up the sidecar in an environment that mimics the alpine `localhost` resolution behavior.
- 301 vs 308: the api-contract reviewer guessed from docs/training; the actual emitter wasn't checked.

A device-bring-up smoke test in CI (or a release-gate check that does `apt install + systemctl start`) would have caught (1) and (2). Worth a follow-up issue.

Refs #129, refs #136.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -26,16 +26,19 @@ override_dh_auto_install:
 	install -D -m 644 icon.png \
 		debian/$(PKG_NAME)/usr/share/pixmaps/$(PKG_NAME).png
 
-	# Install assets (preserving directory structure, +x for .sh files).
-	# IMPORTANT: every assets/<dir>/ that docker-compose.yml bind-mounts
-	# from must appear in this list. Missing dirs ship as nothing; Docker
-	# then creates an empty *directory* at the bind-mount source path on
-	# first start, the bind into the container fails with "not a
-	# directory", and the affected container loops on container-init
-	# error. Lost ca-download/ for an entire release cycle exactly this
-	# way — keep this list in sync with docker-compose.yml volume mounts.
-	for dir in traefik authelia homarr ca-download; do \
-		find assets/$$dir -type f | while read f; do \
+	# Install every assets/<dir>/ tree, preserving directory structure, with
+	# +x on .sh files. The glob (over the previous hardcoded list) makes the
+	# class of bug structurally impossible: any new asset subdir added for a
+	# docker-compose.yml bind-mount ships automatically. We lost ca-download/
+	# for a release cycle when the hardcoded list drifted from compose mounts
+	# — when a bind-mount source is missing, Docker creates an empty directory
+	# at the source path, the bind into the container fails with "not a
+	# directory", and the affected container loops on container-init error.
+	# Single-file assets (configure-container-routing, lib-*.sh, hostnames.conf)
+	# are installed explicitly below and not under assets/<dir>/, so the glob
+	# does not over-capture.
+	for dir in assets/*/; do \
+		find "$$dir" -type f | while read f; do \
 			case "$$f" in \
 				*.sh) install -D -m 755 "$$f" "debian/$(PKG_NAME)$(LIB_DIR)/$$f" ;; \
 				*) install -D -m 644 "$$f" "debian/$(PKG_NAME)$(LIB_DIR)/$$f" ;; \
--- a/debian/rules
+++ b/debian/rules
@@ -26,8 +26,15 @@ override_dh_auto_install:
 	install -D -m 644 icon.png \
 		debian/$(PKG_NAME)/usr/share/pixmaps/$(PKG_NAME).png
 
-	# Install assets (preserving directory structure, +x for .sh files)
-	for dir in traefik authelia homarr; do \
+	# Install assets (preserving directory structure, +x for .sh files).
+	# IMPORTANT: every assets/<dir>/ that docker-compose.yml bind-mounts
+	# from must appear in this list. Missing dirs ship as nothing; Docker
+	# then creates an empty *directory* at the bind-mount source path on
+	# first start, the bind into the container fails with "not a
+	# directory", and the affected container loops on container-init
+	# error. Lost ca-download/ for an entire release cycle exactly this
+	# way — keep this list in sync with docker-compose.yml volume mounts.
+	for dir in traefik authelia homarr ca-download; do \
 		find assets/$$dir -type f | while read f; do \
 			case "$$f" in \
 				*.sh) install -D -m 755 "$$f" "debian/$(PKG_NAME)$(LIB_DIR)/$$f" ;; \
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -199,7 +199,13 @@ services:
       # /srv/halos-ca.crt marks the container unhealthy (autoheal then
       # restarts it). A bare /healthz probe stayed healthy while nginx
       # 404'd the cert, which hid breakage from operators.
-      test: ["CMD", "wget", "-q", "--spider", "http://localhost/halos-ca.crt"]
+      #
+      # Connect via 127.0.0.1 (not localhost): nginx:alpine resolves
+      # localhost to ::1 first, and our nginx.conf only `listen 80` on
+      # IPv4. localhost probes get ECONNREFUSED → container stays
+      # unhealthy → Traefik 3.7 filters unhealthy containers from
+      # discovery → endpoint silently unreachable behind a homarr 404.
+      test: ["CMD", "wget", "-q", "--spider", "http://127.0.0.1/halos-ca.crt"]
       interval: 10s
       timeout: 5s
       start_period: 30s
--- a/docs/CERTS.md
+++ b/docs/CERTS.md
@@ -70,7 +70,7 @@ The active CA is published at:
 
     https://<host>/halos-ca.crt
 
-Plain HTTP requests to the same path are redirected to HTTPS (301). The
+Plain HTTP requests to the same path are redirected to HTTPS (308). The
 sidecar that serves the file returns it with:
 
 | Header                | Value                                       |
@@ -131,7 +131,7 @@ ssh halos.local 'sudo cat /var/lib/container-apps/halos-core-containers/data/hal
 curl -skI https://halos.local/halos-ca.crt | grep -iE 'content-(type|disposition)'
 
 # 3. HTTP redirects to HTTPS.
-curl -sI http://halos.local/halos-ca.crt | head -1   # 301 Moved Permanently
+curl -sI http://halos.local/halos-ca.crt | head -1   # 308 Permanent Redirect
 curl -sI http://halos.local/halos-ca.crt | grep -i ^location
 
 # 4. After swapping the active CA (operator drops files in /etc/halos/ca/


mairas:
## ce:review (multi-persona)

**Scope:** `fix/ca-download-deployment-bugs` vs `main` — 3 files, +18/-5 (`debian/rules`, `docker-compose.yml`, `docs/CERTS.md`).

**Mode:** report-only.

**Intent:** Three independent deployment-time bugs introduced by PR #136 that each broke the `/halos-ca.crt` endpoint on the device:
1. `debian/rules` asset-install loop hardcoded `traefik authelia homarr` → `ca-download/nginx.conf` never shipped → Docker created an empty dir at the bind-mount source → container init failed
2. Healthcheck probed `http://localhost/` → nginx:alpine resolved to `::1` → IPv4-only listener gave `ECONNREFUSED` → container unhealthy → **Traefik 3.7 silently filters unhealthy containers from discovery** → `/halos-ca.crt` fell through to Homarr's catch-all 404
3. `docs/CERTS.md` claimed HTTP→HTTPS redirect was `301`; Traefik 3.7 actually emits `308`

**Reviewer team (8):** correctness, testing, maintainability, project-standards, agent-native, learnings, reliability, api-contract. Conditional picks: reliability (entire PR is about deployment reliability), api-contract (the 301→308 fix is an API-contract correction).

---

### P2 — High

|---|---|---|---|---|---|
| 1 | `debian/rules:37` | Hardcoded asset-dir list is still drift-prone — same bug class will recur on the next new sidecar | correctness, testing, maintainability, reliability (4-way) | 0.95 | gated_auto → downstream-resolver |

**Detail:** The fix adds `ca-download` to the explicit list and a warning comment ("keep this list in sync with docker-compose.yml volume mounts") — but the contract is enforced by reviewer attention only. The next new bind-mounted asset dir reproduces the exact same outage (Docker → empty dir at source → "not a directory" container init failure). **The PR text itself says this bug "lost ca-download/ for an entire release cycle" — a comment is strictly weaker than a structural fix.**

**Concrete fix options, smallest first:**

- **Option A (recommended): replace the list with a glob.** `for dir in assets/*/; do dir="${dir%/}"; ...` makes the bug class structurally impossible. The per-dir branching in the loop is identical for every entry; the `/usr/bin/` and `/usr/lib/<pkg>/` installs further down target individual files explicitly and aren't affected by the glob.
- **Option B: add a CI assertion.** Grep `docker-compose.yml` for `./assets/<dir>/` bind-mount sources and diff against the loop's enumerated dirs at build time. Testing reviewer (0.88) notes this would also catch the regression at build time, not at smoke test, with no .deb-install cost.

### P3 — Low

|---|---|---|---|---|---|
| 2 | `docker-compose.yml:208` | Healthcheck now couples Traefik route availability to prestart having published the cert; transient publish failure → silent 404 with WARNING only in journal | reliability | 0.65 | advisory → human |
| 3 | n/a | No automated contract test for the HTTP→HTTPS 308 + Content-Type/Disposition headers | api-contract, testing (2-way) | 0.75 | advisory → human |
| 4 | `docker-compose.yml:194` (cross-reference; agent-native nit) | Verification commands use `curl ... \| head -1` for HTTP/2 status — brittle against HTTP/2 status lines; prefer `curl -w '%{http_code}\n'` | agent-native | 0.6 | safe_auto → review-fixer (docs only) |

**Detail [2]:** With the healthcheck now probing the real served path, if `halos_ca_publish_public` fails mid-prestart and the helper preserves the previous public copy as designed, ca-download stays healthy. But if there's no previous copy (first-boot, or someone removed it), `/srv/halos-ca.crt` is missing → unhealthy → Traefik filters → operators see a Homarr 404 with the only WARNING buried in `journalctl -u halos-core-containers`. autoheal will loop-restart ca-download but restarting the sidecar doesn't republish the cert — only re-running prestart does. By design per the new comment, but no operational alerting bridges the 404 back to the publish WARNING. Worth a runbook entry; not blocking.

**Detail [3]:** The `att` redirect status was previously documented as 301; this PR corrects to 308. Nothing in CI keeps doc and behavior in lockstep — a future Traefik upgrade that emits different bytes would silently regress without surfacing in any health probe. Worth a smoke-test step that does `curl -sI http://halos.local/halos-ca.crt` and asserts 308.

---

### Pre-existing residual risks

- Other healthchecks (`traefik`, `homarr`) still probe `http://localhost/...`. Currently working in production because those base images bind IPv4 listeners and their respective libc returns 127.0.0.1 from `getaddrinfo("localhost")`. The IPv6-first trap is **nginx:alpine specific**. No live bug, but defensive consistency would harden the pattern. Pre-existing, not introduced by this PR.

This PR is the third hit on patterns already captured in project auto-memory:

- **`learning_traefik_filters_unhealthy.md`** — exact match for bug #2. The PR fix and its inline comment correctly cite the IPv6/IPv4 trap; future readers won't "simplify" back to `localhost`.
- **`feedback_debian_rules_asset_dirs.md`** — exact match for bug #1. The verification step (`./run build && dpkg-deb -c build/*.deb | grep <newdir>`) belongs in the PR's test plan if the glob fix isn't taken.
- **`learning_auxiliary_prestart_calls.md`** — adjacent; the publish helper IS the canonical "auxiliary, preserve-previous-on-failure" pattern.

No prior learning for bug #3 (doc/code redirect-status mismatch) — worth a `ce:compound` entry post-merge.

### Agent-native

**PASS.** All three fixes are agent-introspectable (`dpkg-query -L`, `docker inspect --format '{{.State.Health.Status}}'`, `curl -sI`) and recoverable with existing primitives. Nit: prefer `curl -w '%{http_code}\n'` over `curl -sI ... | head -1` for HTTP/2 status checks.

### Project standards

Clean. Conventional commit, branch name, no debian/changelog edits, `Refs #129, refs #136` umbrella linkage correct, no VERSION bump (already ahead of stable).

---

### Verdict

**Ready with one improvement.** The cross-reviewer consensus on finding [1] is strong (4 of 8 reviewers, including correctness, maintainability, reliability, and testing). The warning comment ships, but the same class of bug *is allowed to recur*. The glob fix (Option A) is ~3 lines and structurally eliminates the bug class.

Recommended order:
1. **[P2 #1]** Apply the `assets/*/` glob — small, mechanical, prevents the next "lost a feature for a release cycle" outage
2. **[P3 #4]** Doc/comment polish (`curl -w '%{http_code}\n'`) — optional, agent-native nit
3. **[P3 #3]** Track the CI smoke-test gap as a follow-up issue
4. **[P3 #2]** Track the publish-fail→silent-404 alerting gap as a follow-up issue

🤖 Generated with `ce:review` skill (8 reviewers, structured findings synthesized + deduped). The strongest signal — replace warning comment with structural glob — converges across 4 independent reviewers.

mairas:

Fix commit: [`a82f4c8`](https://github.com/halos-org/halos-core-containers/commit/a82f4c8).

### P2

| # | Finding | Resolution |
|---|---|---|
| 1 | Hardcoded asset-dir list still drift-prone (4-way reviewer agreement) | ✅ Replaced `for dir in traefik authelia homarr ca-download` with `for dir in assets/*/`. Bug class is now structurally impossible — any new asset subdir ships automatically. Verified glob captures identical file set as the previous list with `diff <(for d in assets/*/; do find "$d" -type f; done) <(for d in traefik authelia homarr ca-download; do find "assets/$d" -type f; done)` → empty (same set). Single-file assets at `assets/<file>` (lib-*.sh, configure-container-routing, etc.) are installed explicitly below the loop and unaffected. |

### P3

| # | Finding | Resolution |
|---|---|---|
| 2 | Healthcheck couples Traefik route availability to prestart publish success → transient publish-fail → silent 404 with WARNING only in journal | ⏭ By design per the new docker-compose comment; documented behavior. Will track as a runbook entry — adding alerting that bridges the 404 back to the prestart WARNING is out of scope for this PR and best handled in monitoring/observability follow-up. |
| 3 | No automated contract test for HTTP→HTTPS 308 + Content-Type/Disposition headers | ⏭ Requires a docker-based integration smoke-test step in CI (or a release-gate `apt install + curl` check). Out of scope for this PR; will track as a separate CI-infrastructure follow-up. The same gap was flagged for bugs #1 and #2 in the PR body. |
| 4 | `curl -sI ... \| head -1` brittle for HTTP/2 status parsing (agent-native nit at 0.6 confidence) | ⏭ Skipped — borderline (right at the 0.60 suppression threshold), and the current form is intentional for human visual verification ("HTTP/2 308" is more informative than just "308" for an operator copy-pasting from the doc). For scripted checks an automation would use `-w '%{http_code}'`; the doc form stays human-readable. |

### Pre-existing residual

- Other healthchecks (`traefik`, `homarr`) still probe `localhost` — pre-existing, not introduced by this PR. The `nginx:alpine` IPv6-first behavior is image-specific; current production images bind IPv4 listeners and work fine. Defensive consistency could be a separate hardening pass.

### Follow-ups to track (not blocking)

- CI smoke-test that brings up `nginx:1.27-alpine` with `assets/ca-download/nginx.conf` + a fixture cert and asserts the healthcheck transitions to healthy + `curl /halos-ca.crt` returns 200 with the right headers. Would have caught bugs 1, 2, and 3 from this PR at build time.
- Runbook entry mapping a Homarr 404 on `/halos-ca.crt` back to a journalctl `WARNING: failed to refresh public CA copy` (the silent-degradation cascade reliability flagged).

Issue #140, opened 2026-05-25
Add periodic cert renewal (systemd timer) so long-running devices don't silently expire their leaf


mairas:
## Context

`HALOS_CA_LEAF_VALIDITY_DAYS=824` (set in #141) complies with Apple Secure Transport's 825-day ceiling for SSL server certs. Apple is the binding constraint; Linux/Windows don't enforce it. See [Apple's TLS server cert requirements](https://support.apple.com/en-us/103769).

After #139 lands, cert provisioning runs in `halos-manage-certs.service` (oneshot, ordered before `halos-core-containers.service`). The script is idempotent and already includes a renewal-threshold check via `halos_ca_leaf_needs_renewal` — but the service only fires on container-stack activation (boot, manual restart, package upgrade). A device that boots once and runs continuously for >824 days never re-fires the service, so the leaf silently expires and Traefik starts failing TLS with no operational signal until users hit the broken cert.

In typical HaLOS deployments this isn't catastrophic (devices on `apt unattended-upgrades` or regular use see frequent restart cycles), but it's a real gap. This issue closes it with a periodic activation trigger.

## Proposed shape

Add a **single new file plus one small script-level addition** — no new service unit.

### New: `halos-manage-certs.timer`

```
[Unit]
Description=Periodic HaLOS device cert renewal check
Requires=halos-manage-certs.service

[Timer]
OnBootSec=15min
OnUnitActiveSec=24h
RandomizedDelaySec=1h
Persistent=true
Unit=halos-manage-certs.service

[Install]
WantedBy=timers.target
```

- `OnBootSec=15min`: first fire 15 min after boot so the stack settles before the timer-driven run.
- `OnUnitActiveSec=24h`: re-check every 24 hours. Well below the 60-day `HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS` window, so even fortnightly-rebooted devices get multiple chances to renew.
- `RandomizedDelaySec=1h`: spread fleet-wide hits over a one-hour window — avoids synchronized renewal storms on shared APT mirrors or upstream Authelia rotations.
- `Persistent=true`: if the device was off when a fire was due, run on next boot to catch up.

### Cockpit-reload-on-change in `halos-manage-certs`

[...]

Issue #139, opened 2026-05-25
Extract cert-artifact management out of prestart.sh into a dedicated cert-management entity


mairas:
## Context

`halos-core-containers/prestart.sh` has accumulated several distinct cert-lifecycle responsibilities that don't belong to "service init" but are convenient to pile in because the script runs at startup. The accretion happened across #129 (units 1–4: #130, #131, #132, #133):

- **Active CA selection** (`halos_ca_select_active`, line ~110) — choose custom vs auto CA
- **Public CA copy** (`halos_ca_publish_public`, line ~134) — refresh `${CONTAINER_DATA_ROOT}/.../certs/public/halos-ca.crt` for the `ca-download` sidecar
- **Leaf signing** (`halos_ca_sign_leaf` inside the `NEED_LEAF` block, lines ~155–200) — re-sign the leaf when the hostname/CA-fingerprint sentinel changes
- **Cockpit cert sharing** (`halos_cockpit_install_leaf`, line ~215) — install combined PEM at `/etc/cockpit/ws-certs.d/99-halos.cert`

These are all "cert artifact maintenance" tasks. They share the same surface (cert files, CA state, sentinel tracking) but are conceptually distinct from prestart's other responsibilities (Traefik dynamic config, OIDC client snippet install, Authelia secrets, Homarr DB seed).

The accumulation has two costs:

1. **`prestart.sh` is becoming a kitchen sink** — harder to reason about, harder to test as a unit.
2. **Cert lifecycle is tied to service restarts** — no way to refresh certs without bouncing Traefik/Authelia/Homarr. Matters for the periodic-renewal case tracked separately (#140).

## Proposed shape

Factor the cert-lifecycle steps out of `prestart.sh` into a standalone script invoked by a **dedicated systemd oneshot unit**, not by `prestart.sh`. Calling the new script from `prestart.sh` would be relocation with a stub, not extraction — `prestart.sh` would still be coupled to cert lifecycle.

### Script

- `assets/halos-manage-certs` — standalone executable script
- Installed to `/usr/lib/halos-core-containers/` per project convention
- Performs the four cert-lifecycle steps currently in `prestart.sh`:
  CA selection → leaf re-sign (gated on hostname/CA-fingerprint sentinel **and** `halos_ca_leaf_needs_renewal` from `lib-ca.sh`) → public-CA publish → cockpit cert install
[...]

PR #141, opened 2026-05-25
fix(certs): cap leaf validity at Apple's 825-day ceiling + rename ambiguous constants

mairas:
## What broke

10-year leaf validity (`HALOS_CA_LEAF_DAYS=3650`) made Safari, Brave, `/usr/bin/curl`, `nscurl`, and every other macOS Secure Transport consumer reject all device URLs — *even with the HaLOS CA installed as a trusted root*. The cause: Apple enforces a **825-day max validity for SSL server certs** regardless of whether the root is publicly trusted or user-installed (CA/B Forum baseline + [Apple policy](https://support.apple.com/en-us/103769) since 2019-07-01). `openssl s_client -CAfile` validated the chain fine, but every `SecTrustEvaluate` consumer reported `CSSMERR_TP_CERT_SUSPENDED`.

Diagnosed live on `halosdev.local` after installing the CA into the macOS System keychain with Always Trust set — the trust override was correct; the leaf validity was the problem.

## Fix

`HALOS_CA_LEAF_VALIDITY_DAYS=824` (one day below the cap). openssl's `-days N` with explicit `-not_before` yields `notAfter - notBefore = (N+1) days` inclusive — empirically verified, `825` produces 826-day certs that Apple rejects. The 1-day headroom keeps us safely at-but-not-over the ceiling.

CA validity is unchanged at 20 years — Apple's max-validity policy is leaf-only. Operators install the CA once and the same CA signs renewing leaves transparently for the CA's lifetime.

## Naming cleanup

The existing `HALOS_CA_*_DAYS` constants conflated two meanings: "lifetime at signing time" vs "remaining-validity gate for action". Renamed all on the same convention:

> `*_VALIDITY_DAYS` — lifetime stamped into the cert at signing time
> `*_THRESHOLD_DAYS` — remaining-validity gate that triggers an action

| Old | New |
|---|---|
| `HALOS_CA_DAYS` | `HALOS_CA_VALIDITY_DAYS` |
| `HALOS_CA_LEAF_DAYS` | `HALOS_CA_LEAF_VALIDITY_DAYS` |
| `HALOS_CA_AUTO_ROTATE_DAYS` | `HALOS_CA_AUTO_REGEN_THRESHOLD_DAYS` |
| `HALOS_CA_CUSTOM_MIN_DAYS` | `HALOS_CA_CUSTOM_REJECT_THRESHOLD_DAYS` |
| `HALOS_CA_CUSTOM_WARN_DAYS` | `HALOS_CA_CUSTOM_WARN_THRESHOLD_DAYS` |
| *(new)* | `HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS=60` |

Also reworked the `HALOS_CA_AUTO_REGEN_THRESHOLD_DAYS` comment to explicitly state its real purpose: **clock-skew self-heal on Pi hardware (no RTC).** The 365-day threshold is *not* about natural-aging rotation — that branch never fires within any plausible device lifetime — it catches the case where a Pi boots before NTP, signs a CA dated 1970-2020 (or whatever `/etc/fake-hwclock.data` last saved), and needs to regenerate once NTP corrects the clock.

## Library API addition

New public function `halos_ca_leaf_needs_renewal <crt>` — returns 0 when the leaf is missing/expired/within `HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS` of expiry; 1 when still healthy. **No callers in this PR** — intentionally. The function will be consumed by the cert-management entity tracked in #139 and the renewal timer in #140.

## Why no caller in prestart

A two-line addition to `prestart.sh` to call this on startup would close the rotation gap for boot-and-upgrade cycles. **Deliberately not doing it.** `prestart.sh` is already a kitchen sink of init concerns (CA selection, leaf signing, cockpit cert sharing, public CA publish, Traefik configs, Authelia secrets, OIDC merge, Homarr DB seed). Adding cert-maintenance check there compounds the problem this project's already paying for. The clean home is a dedicated cert-management entity + a systemd timer — see #139 and #140 for the proper sequencing, with this PR providing the helper they consume.

## Tests

**74/74 pass.** New cases:

- `test_sign_leaf_validity_under_apple_825_day_cap` — regression guard against any future bump
- `test_leaf_needs_renewal_when_fresh_returns_false`
- `test_leaf_needs_renewal_when_missing_returns_true`
- `test_leaf_needs_renewal_when_approaching_expiry_returns_true` (synthesizes a 30-day-remaining leaf, asserts renewal fires)
- `test_leaf_needs_renewal_when_well_outside_window_returns_false` (synthesizes a 365-day-remaining leaf, asserts no renewal)

## Hitchhiker

The first commit on this branch is a tiny chore — adds `.worktrees/` to `.gitignore` for the local-parallel-work convention. No effect on builds or shipped package; was sitting uncommitted on the workspace and tagged along.

## Refs

- Apple max-validity policy: https://support.apple.com/en-us/103769
- Follow-up #139 — extract cert lifecycle out of `prestart.sh`
- Follow-up #140 — periodic-renewal timer that consumes `halos_ca_leaf_needs_renewal`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@ debian/files
 debian/halos-core-containers/
 debian/*.substvars
 debian/*.debhelper
+
+# Local git worktrees for parallel work
+.worktrees/
--- a/assets/lib-ca.sh
+++ b/assets/lib-ca.sh
@@ -31,27 +31,73 @@
 # a bad clock self-heal after NTP recovers.
 
 # Tuning (constants, not env-overridable — keep the surface tight) -----------
+#
+# Naming convention:
+#   *_VALIDITY_DAYS   — lifetime stamped into the cert at signing time
+#   *_THRESHOLD_DAYS  — remaining-validity gate that triggers an action
+#                       (regenerate / renew / reject / warn)
 
-HALOS_CA_DAYS=7300            # 20 years
-HALOS_CA_LEAF_DAYS=3650       # 10 years
+# --- Validity (lifetime at signing time) ---
+
+HALOS_CA_VALIDITY_DAYS=7300       # 20 years
+
+# Leaf validity. Capped at 825 days because Apple's Secure Transport rejects
+# SSL server certs with longer validity (CA/B Forum baseline + Apple policy
+# since 2019-07-01), regardless of whether the root is publicly trusted or
+# user-installed. Affects Safari, Chrome/Brave/Edge, /usr/bin/curl, nscurl,
+# and every macOS app using SecTrustEvaluate. Linux/Windows don't enforce
+# this, but the 825-day ceiling is the binding constraint.
+# Reference: https://support.apple.com/en-us/103769
+#
+# Set to 824, not 825: openssl's `-days N` with an explicit -not_before
+# yields notAfter - notBefore = (N+1) days (inclusive count). Empirically
+# verified — 825 produces 826-day certs that Apple rejects. The 1-day
+# headroom keeps us safely at-but-not-over the ceiling.
+HALOS_CA_LEAF_VALIDITY_DAYS=824
+
 HALOS_CA_BACKDATE_HOURS=24
 HALOS_CA_SUBJECT="/CN=HaLOS Device CA"
 
-# Auto-CA rotation trigger: rotate the device-generated CA when its remaining
-# validity drops below this many days. Set well below the 20y validity, so
-# the realistic trigger is clock skew / corruption — not natural aging.
-HALOS_CA_AUTO_ROTATE_DAYS=365
+# --- Thresholds (remaining-validity gates for actions) ---
 
-# Custom (operator-supplied) CA acceptance thresholds:
-#   - hard floor: refuse to use a custom CA with less than this many days
-#     remaining. Lower than the auto-rotate threshold because operators
-#     deliberately choose short-lived intermediates in some deployments
-#     and we don't want to refuse a valid 6-month CA out of the box.
-#   - warn window: emit a loud WARNING in prestart logs when remaining
-#     validity is below this, even if still above the hard floor. Gives
-#     operators a documented runway to rotate before the hard cliff.
-HALOS_CA_CUSTOM_MIN_DAYS=30
-HALOS_CA_CUSTOM_WARN_DAYS=90
+# Re-sign the leaf when remaining validity drops below this. Sized so
+# operators have generous runway (~7% of leaf lifetime); any reboot during
+# the window triggers transparent renewal. The CA stays trusted across leaf
+# rotation because the same CA signs the new leaf — installed trust anchors
+# do NOT need to be re-installed (the CA itself is valid for
+# HALOS_CA_VALIDITY_DAYS).
+HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS=60
+
+# Regenerate the auto-CA when its remaining validity drops below this.
+# Despite the threshold being set well below the 20-year validity, the
+# realistic trigger is NOT natural aging — that branch never fires within
+# any plausible device lifetime.
+#
+# The actual purpose is clock-skew self-heal. Raspberry Pi hardware has no
+# RTC; first boot starts with system time = whatever /etc/fake-hwclock.data
+# last saved (often 1970, or the timestamp of the most recent build). If
+# prestart signs the auto-CA at that point, notBefore and notAfter are both
+# wrong by years. When NTP later corrects the clock, the cert appears
+# already expired (or expiring soon), this check fires, and the next
+# prestart regenerates with correct dates. Without this, a Pi that booted
+# before NTP would ship a permanently-broken CA until manual recovery.
+#
+# Same mechanism also covers half-deleted CA files, disk corruption, and
+# any other "cert exists on disk but is unusable" state. Operators who
+# installed the auto-CA into their trust store will see it become invalid
+# when this fires — accepted trade-off for the self-heal property (custom
+# CAs, by contrast, are never auto-regenerated for exactly this reason).
+HALOS_CA_AUTO_REGEN_THRESHOLD_DAYS=365
+
+# Custom (operator-supplied) CA acceptance thresholds. Operators may
+# deliberately ship short-lived intermediates, so REJECT is lower than the
+# auto-CA regen threshold — we don't want to reject a valid 6-month CA on
+# arrival. WARN gives operators a documented runway to rotate before
+# hitting the REJECT cliff. Unlike the auto-CA, custom CAs are never
+# auto-regenerated; operator owns rotation (silent regen would orphan the
+# trust anchors they distributed to a fleet).
+HALOS_CA_CUSTOM_REJECT_THRESHOLD_DAYS=30
+HALOS_CA_CUSTOM_WARN_THRESHOLD_DAYS=90
 
 # Internal -------------------------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


mairas:

**Scope:** 200 additions / 31 deletions across `assets/lib-ca.sh`, `tests/test-lib-ca.sh`, `.gitignore`. Base `5035177`. CI green.

**Intent:** Stop macOS/iOS Secure Transport rejecting HaLOS TLS by capping leaf validity below Apple's 825-day ceiling. Tighten naming on `*_VALIDITY_DAYS` vs `*_THRESHOLD_DAYS`. Add `halos_ca_leaf_needs_renewal` helper for follow-ups #139/#140 (deliberately uncalled in this PR).

**Reviewer team:** correctness, testing, maintainability, project-standards, agent-native, learnings-researcher (always-on) + security, reliability, adversarial (>50 LOC + cert lifecycle), api-contract (public bash constants renamed + new public function).

---

### P1 — High

|---|---|---|---|---|
| 1 | `assets/lib-ca.sh:183` + `prestart.sh` | **Renewal gap window between this merge and #139/#140 landing.** `halos_ca_leaf_needs_renewal` is defined but has no caller in this PR. For *new* signings post-upgrade (after a hostname/CA change or fresh bootstrap), the 824-day leaf won't be renewed until something else trips the sentinel. Devices that subsequently boot without config changes for ~2.3 years will serve an expired leaf with no automatic recovery. Pre-existing 3650-day leaves are unaffected (sentinel doesn't fire), so this is a **forward-looking** risk, not a fleet-wide regression. The PR body acknowledges this and routes the fix to #139/#140; the question is whether you accept the gap until those land. | security, reliability, adversarial, correctness, maintainability (5×) | 0.90 |

### P2 — Moderate

|---|---|---|---|---|
| 2 | `assets/lib-ca.sh:424–437` | `halos_ca_sign_leaf` docstring still implies the legacy 10-year default; doesn't note the 3650→824 behavior change or cite the Apple constraint. Callers omitting the 7th arg now silently get sub-3-year certs. | api-contract | 0.85 |
| 3 | `assets/lib-ca.sh:7–19` | New public function `halos_ca_leaf_needs_renewal` is missing from the file-header `Provides:` API listing. | testing | 0.72 |

### P3 — Low

|---|---|---|---|---|
| 4 | `assets/lib-ca.sh:168–181` | `_halos_ca_is_healthy` header says its default `min_remaining_days` is set "for backwards compatibility with the auto-CA call site". The function is private (underscore prefix); "backwards compatibility" is the wrong frame — it's a design default. Reword to "matches the auto-CA regen gate". | maintainability | 0.72 |
| 5 | `tests/test-lib-ca.sh` | No test for the already-expired leaf path (notAfter in the past). Existing tests cover fresh/missing/approaching/far-out; the expired branch is taken by `_halos_ca_is_healthy` (`-checkend` returns non-zero), which makes `halos_ca_leaf_needs_renewal` return 0. Correct behavior, untested. | correctness | 0.75 |
| 6 | `tests/test-lib-ca.sh` | No test for symlink/directory inputs to `halos_ca_leaf_needs_renewal`. `[ -f ]` follows symlinks, but a broken symlink would return "needs renewal" silently. Low likelihood (callers pass admin-controlled paths); worth covering if #139/#140 introduce path indirection. | testing, adversarial, security | 0.65 |
| 7 | `tests/test-lib-ca.sh:1190` | No explicit test that `halos_ca_sign_leaf` produces an 824-day cert *when the 7th arg is omitted*. The Apple-cap test only checks `validity_days <= 825`, so a future regression to a different default that still happens to be ≤825 would slip through. | api-contract | 0.65 |
| 8 | `assets/lib-ca.sh:44–89` | Inline comments on `HALOS_CA_LEAF_VALIDITY_DAYS` (13 lines) and `HALOS_CA_AUTO_REGEN_THRESHOLD_DAYS` (18 lines) are dense. Consider moving the Apple-policy explanation to `docs/CERTS.md` with a short pointer at the constant. Stylistic; the level of detail is justified by the precision required. | maintainability | 0.65 |
| 9 | `.gitignore` | `.worktrees/` hitchhikes a cert-security PR. Author flagged transparently in the body ("Hitchhiker"). 3 lines, no build/package impact. Technically violates "one logical change per PR"; in practice acceptable. | project-standards | 0.65 |
| 10 | PR body | No `Closes #` / `Refs #` linkage to a tracking issue for the Apple-rejection bug itself (only `Refs #139`, `#140` for follow-ups). AGENTS.md requires issue linkage. The fix is fully self-documenting in the PR body; creating a retroactive issue purely for compliance is bureaucratic, but it's a literal standards miss. | project-standards | 0.70 |

---

### Pre-existing (not introduced by this PR)

- **Auto-CA regeneration orphans operator-installed trust anchors.** `HALOS_CA_AUTO_REGEN_THRESHOLD_DAYS=365` fires when a device boots with a bad clock and later resyncs via NTP. Operators who pushed CA-v1 to a fleet's trust stores will see CA-v2 silently take over with no client-side notification. This PR makes the trade-off *more honest* in the comment but doesn't fix it — and shouldn't, scope-wise.

### Coverage

- **Cross-platform openssl variance.** Adversarial flagged "Alpine could produce 826-day certs from `-days 824`." Moot in practice: the cert is signed on the device (Raspberry Pi OS / Debian), not in Alpine containers. CI runs the test in a `debian:trixie` container per workspace convention.
- **External constant consumers.** api-contract searched the workspace; no external scripts reference the old `HALOS_CA_*_DAYS` names. Library is effectively internal to `halos-core-containers`.
- **Renewal threshold (60d) vs device reboot cadence.** Reliability flagged as a deployment assumption. Will be revisited when #140 lands the systemd timer.

- `docs/solutions/2026-05-25-openssl-text-grep-defeated-by-subject-dn.md` — the validate-pair extension-grep pattern is preserved; no regression. ✓
- `docs/solutions/2026-05-24-set-e-cmdsubst-blind-spot.md` — **load-bearing for #139/#140**: when those PRs wire `halos_ca_leaf_needs_renewal` into `prestart.sh`, callers must use `if halos_ca_leaf_needs_renewal "$leaf"; then …` (return-code form), NOT `result=$(halos_ca_leaf_needs_renewal …)` under `set -e`. Worth flagging in #139's plan.

### Agent-Native

No gaps — library-only change with no new user-facing surface.

---

**Verdict: Ready to merge with one design decision and three safe doc fixes.**

The P1 renewal-gap is the only real decision: accept the window before #139/#140 land, or wire a minimal stop-gap caller in this PR (which the PR body explicitly declined for principled reasons)? Read: the author's framing is defensible — existing devices keep their old 10-year leaves (unaffected), new signings get 824-day leaves that will most likely be re-signed during normal config churn, and the proper fix is structurally cleaner in #139/#140. Accept the deferral provided #139/#140 don't slip past a routine release cadence.

The P2/P3 doc fixes (#2, #3, #4) are one-liners worth folding in before merge. The test additions (#5, #6, #7) can defer to a follow-up.

mairas:

Fixes pushed in fc197ce. P1 dropped per author intent to land #139/#140 in parallel (renewal-gap closes when those merge, not by stop-gap caller here).

### P1
- ⏭ **#1 Renewal gap** — skipped. #139 and #140 are being implemented next; the gap exists only between this merge and theirs. A stop-gap caller in this PR would re-introduce the prestart.sh accretion the rewrite is meant to eliminate.

### P2
- ✅ **#2 `halos_ca_sign_leaf` docstring missing default/Apple rationale** — fc197ce, `assets/lib-ca.sh:431–437`.
- ✅ **#3 `halos_ca_leaf_needs_renewal` missing from `Provides:` block** — fc197ce, `assets/lib-ca.sh:11`.

### P3
- ✅ **#4 `_halos_ca_is_healthy` "backwards compatibility" comment** — fc197ce, `assets/lib-ca.sh:168–175`. Rewritten to name the three real call sites instead.
- ✅ **#5 No test for already-expired leaf** — fc197ce, new `test_leaf_needs_renewal_when_already_expired_returns_true`.
- ⏭ **#6 No symlink/directory tests for `halos_ca_leaf_needs_renewal`** — skipped. Function is called with admin-controlled paths only; speculative coverage.
- ✅ **#7 No test for `halos_ca_sign_leaf` default-7th-arg** — fc197ce, new `test_sign_leaf_default_validity_uses_apple_compliant_days`. Tests both the Apple ceiling and a 30-day floor against accidental drops.
- ⏭ **#8 Comment density on validity / regen constants** — skipped. The verbose inline form was a deliberate authoring choice; moving to docs/CERTS.md would split the rationale from the code that depends on it.
- ⏭ **#9 `.gitignore` hitchhiker** — skipped. Flagged transparently in the original PR body; 3 lines with no build impact.
- ⏭ **#10 Missing tracking-issue link in PR body** — skipped. Bug is fully self-documented in the PR body; retroactive issue purely for compliance is bureaucratic.

Local test run: 76/76 lib-ca pass, 39/39 lib-hostnames pass.

PR #142, opened 2026-05-26
feat(certs): extract cert management into halos-manage-certs.service (closes #139)

### LINKED ISSUE #139: Extract cert-artifact management out of prestart.sh into a dedicated cert-management entity

mairas:
Closes #139.

## What

Move CA selection, leaf signing, public-CA publish, and Cockpit cert sharing out of `prestart.sh` into a dedicated standalone script invoked by a new systemd oneshot unit ordered `Before=halos-core-containers.service`. `prestart.sh` now contains zero cert-lifecycle logic — `grep "halos_ca_" prestart.sh` returns 0 matches.

## Why a oneshot unit, not a call from prestart

The goal of extraction is to **decouple** cert lifecycle from container service restarts. Calling the new script from `prestart.sh` would be relocation with a stub — prestart would still own activation, #140's renewal timer would need a duplicate code path, and operators couldn't force a cert refresh without bouncing Traefik/Authelia/Homarr.

The oneshot pattern collapses #139 and #140 into a single execution path:

```
halos-manage-certs.service  (oneshot, Before=halos-core-containers.service)
                  ↑                ↑
                  |                |
       boot/restart of stack    halos-manage-certs.timer (#140)
```

One script, one service unit, two activation triggers. #140 becomes a thin `.timer` file with no new logic.

## Renewal gate

The script adds `halos_ca_leaf_needs_renewal` as a second `NEED_LEAF` gate alongside the existing sentinel check, so unchanged-config devices still rotate their leaf before Apple's 825-day ceiling expires it (the gap the #141 helper was added to close).

## Acceptance criteria (from #139)

- [x] All cert-related operations move to the new script — `assets/halos-manage-certs`
- [x] `prestart.sh` has zero `halos_ca_*` references — verified by grep
- [x] Existing behavior preserved (sentinel logic, ordering, `|| WARN` for auxiliary calls per #135 pattern)
- [x] Tests cover the script as a unit — 7 cases in `tests/test-halos-manage-certs.sh`: first-boot bootstrap, idempotent rerun, hostname-change leaf re-sign, renewal-threshold leaf re-sign, expired-CA rotation cascade, cockpit-dir-absent skip, broken-custom-CA abort
- [x] `docs/CERTS.md` updated to point at the new unit (new "Where cert lifecycle lives" section + cockpit revert/refresh procedures + custom-CA swap instructions)

## Files

| File | Change |
|---|---|
| `assets/halos-manage-certs` | new — standalone executable script |
| `debian/halos-core-containers.halos-manage-certs.service` | new — systemd oneshot unit |
| `tests/test-halos-manage-certs.sh` | new — 7 tests, all passing |
| `prestart.sh` | strip cert blocks; replace with explanatory pointer |
| `debian/rules` | install new script to `/usr/lib/<pkg>/halos-manage-certs` |
| `debian/postinst` | `systemctl enable halos-manage-certs.service` |
| `assets/lib-ca.sh` | update "Sourced by" header |
| `assets/lib-hostnames.sh` | update "Sourced by" header |
| `docs/CERTS.md` | new top-section + revert/refresh/swap procedure updates |

## Test plan

- [x] `./run test` — 7/7 halos-manage-certs + 76/76 lib-ca + 39/39 lib-hostnames pass locally
- [x] `bash -n` syntax check on both scripts
- [x] `shellcheck -x assets/halos-manage-certs` — only the expected SC1091 info about runtime-sourced env files
- [ ] CI run on this PR
- [ ] Smoke test on `halosdev.local` after package build: `systemctl status halos-manage-certs.service` shows `active (exited)`, the leaf at `/var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.crt` is fresh, and `halos-core-containers.service` starts cleanly with `After=halos-manage-certs.service` ordering
- [ ] Force-refresh smoke: `sudo systemctl start halos-manage-certs` re-runs cleanly without touching the rest of the stack

## Operator notes

To force a cert refresh post-merge without restarting the full container stack:

```
sudo systemctl start halos-manage-certs.service
```

To revert to upstream Cockpit self-signed:

```
sudo rm /etc/cockpit/ws-certs.d/99-halos.cert
sudo systemctl restart cockpit
```

(Next `halos-manage-certs.service` activation will re-create the override — see `docs/CERTS.md`.)

## Follow-ups

- #140 — add `halos-manage-certs.timer` for periodic renewal. Now collapses to a single `.timer` file with no new logic.

### DIFF
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -1,0 +1,216 @@
+#!/bin/bash
+# halos-manage-certs — provision and maintain HaLOS device CA + leaf cert
+#
+# Runs as a systemd oneshot ordered Before=halos-core-containers.service.
+# Also invoked by halos-manage-certs.timer (see #140) for periodic renewal
+# checks against Apple's 825-day leaf-validity ceiling. Idempotent on every
+# invocation:
+#   - reuses an existing healthy auto-CA; rotates on expiry / clock-skew
+#   - re-signs the leaf only when the sentinel mismatches OR the leaf is
+#     within HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS of expiry
+#   - re-publishes the public CA copy unconditionally (cheap, self-heals)
+#   - re-installs the cockpit override unconditionally (cheap, self-heals)
+#
+# Auxiliary failures (public-CA publish, cockpit override install) log WARN
+# and continue. The leaf and its CA are the hard prerequisites for the
+# container stack; everything else is opportunistic.
+#
+# Path-related state can be overridden via env for tests:
+#   HALOS_ETC_DIR              /etc/container-apps/<pkg>
+#   HALOS_LIB_HOSTNAMES        /usr/lib/<pkg>/lib-hostnames.sh
+#   HALOS_LIB_CA               /usr/lib/<pkg>/lib-ca.sh
+#   HALOS_CUSTOM_CA_DIR        /etc/halos/ca
+#   HALOS_COCKPIT_WS_CERTS_DIR /etc/cockpit/ws-certs.d
+
+set -e
+
+PACKAGE_NAME="${PACKAGE_NAME:-halos-core-containers}"
+ETC_DIR="${HALOS_ETC_DIR:-/etc/container-apps/${PACKAGE_NAME}}"
+
+# env.defaults carries CONTAINER_DATA_ROOT and PACKAGE_NAME (the postinst
+# writes both). `set -a` exports newly-assigned vars so lib functions see
+# them without per-assignment export. Tests skip env.defaults by passing
+# CONTAINER_DATA_ROOT directly and pointing HALOS_ETC_DIR at an empty dir.
+set -a
+[ -f "${ETC_DIR}/env.defaults" ] && . "${ETC_DIR}/env.defaults"
+[ -f "${ETC_DIR}/env" ] && . "${ETC_DIR}/env"
+set +a
+
+: "${CONTAINER_DATA_ROOT:=/var/lib/container-apps/${PACKAGE_NAME}}"
+
+LIB_HOSTNAMES="${HALOS_LIB_HOSTNAMES:-/usr/lib/halos-core-containers/lib-hostnames.sh}"
+LIB_CA="${HALOS_LIB_CA:-/usr/lib/halos-core-containers/lib-ca.sh}"
+
+# shellcheck source=assets/lib-hostnames.sh
+. "$LIB_HOSTNAMES"
+# shellcheck source=assets/lib-ca.sh
+. "$LIB_CA"
+
+halos_load_hostnames
+HALOS_DOMAIN="$(halos_canonical_hostname)"
+
+echo "HaLOS cert management — domain: ${HALOS_DOMAIN}"
+
+# TLS Certificate Hierarchy ---------------------------------------------------
+#
+# Per-device CA signs a leaf used by Traefik (:443) and shared into Cockpit
+# (:9090) via a combined-PEM override. Operators can install the CA in their
+# OS trust store so all of the device's HTTPS endpoints validate.
+#
+# Two CA-source modes, picked by halos_ca_select_active:
+#   - custom: operator dropped /etc/halos/ca/ca.{crt,key} and the pair validates
+#   - auto:   no custom present → auto-generated per-device CA
+# The active CA is exposed via the serving-ca.crt symlink so the ca-download
+# sidecar always reads "whichever CA is active right now".
+#
+# Layout:
+#   ${CUSTOM_CA_DIR}/ca.{crt,key}         optional operator-supplied custom CA
+#   ${AUTO_CA_DIR}/ca.{crt,key}           auto-generated device CA (20y)
+#   ${AUTO_CA_DIR}/serving-ca.crt         symlink → active CA cert
+#   ${CERTS_DIR}/halos.{crt,key}          leaf signed by active CA (~824d)
+#   ${CERTS_DIR}/.domain                  sentinel: "<hostname-hash>:<ca-fingerprint>"
+CUSTOM_CA_DIR="${HALOS_CUSTOM_CA_DIR:-/etc/halos/ca}"
+AUTO_CA_DIR="${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/ca"
+SERVING_CA="${AUTO_CA_DIR}/serving-ca.crt"
+TRAEFIK_DATA="${CONTAINER_DATA_ROOT}/traefik"
+CERTS_DIR="${TRAEFIK_DATA}/certs"
+CERT_FILE="${CERTS_DIR}/halos.crt"
+KEY_FILE="${CERTS_DIR}/halos.key"
+DOMAIN_FILE="${CERTS_DIR}/.domain"
+PUBLIC_CA_DIR="${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/public"
+COCKPIT_WS_CERTS_DIR="${HALOS_COCKPIT_WS_CERTS_DIR:-/etc/cockpit/ws-certs.d}"
+COCKPIT_LEAF_OVERRIDE="${COCKPIT_WS_CERTS_DIR}/99-halos.cert"
+
+mkdir -p "${CERTS_DIR}"
+
+# Select active CA. Custom (when present + valid) wins; otherwise auto is
+# ensured/rotated. A broken custom CA aborts cert management rather than
+# silently falling back — operators who installed a custom CA can SSH to
+# diagnose, and silent fallback would orphan the trust anchor they
+# distributed.
+halos_ca_select_active "${CUSTOM_CA_DIR}" "${AUTO_CA_DIR}" "${SERVING_CA}"
+CA_CRT="${HALOS_CA_ACTIVE_CRT}"
+CA_KEY="${HALOS_CA_ACTIVE_KEY}"
+echo "Active CA: ${CA_CRT} (mode=${HALOS_CA_ACTIVE_MODE})"
+
+# Publish the active CA's PEM bytes into the public dir the ca-download
+# sidecar bind-mounts. Refreshed unconditionally so the published bytes
+# track whichever CA halos_ca_select_active picked.
+#
+# Auxiliary: failure does NOT abort. The previous public copy (if any) is
+# preserved by the helper; consumers of /halos-ca.crt keep serving the
+# old bytes until the next successful run.
+if ! halos_ca_publish_public "${HALOS_CA_ACTIVE_CRT}" "${PUBLIC_CA_DIR}"; then
+    echo "WARNING: failed to refresh public CA copy at ${PUBLIC_CA_DIR}/halos-ca.crt; the /halos-ca.crt download endpoint may serve stale or no CA until the next successful run." >&2
+fi
+
+# Explicit guard: `set -e` does NOT abort on command-substitution failure
+# in an assignment. Without it, a corrupt CA silently yields CA_FINGERPRINT=""
+# and feeds a malformed sentinel into downstream comparison — an infinite
+# re-sign loop.
+CA_FINGERPRINT=$(halos_ca_fingerprint "${CA_CRT}") || {
+    echo "halos-manage-certs: failed to compute CA fingerprint; cert generation cannot continue" >&2
+    exit 1
+}
+
+HOSTNAMES_HASH="$(halos_hostnames_hash)"
+SENTINEL="$(halos_ca_sentinel_compose "${HOSTNAMES_HASH}" "${CA_FINGERPRINT}")"
+NEED_LEAF=false
+if [ ! -f "${CERT_FILE}" ] || [ ! -f "${KEY_FILE}" ]; then
+    echo "Leaf certificate files not found, generating..."
+    NEED_LEAF=true
+elif [ -f "${DOMAIN_FILE}" ]; then
+    STORED=$(cat "${DOMAIN_FILE}")
+    case "$(halos_ca_sentinel_classify "${STORED}")" in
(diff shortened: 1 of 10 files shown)


mairas:

**Scope:** 9 files, +588/-188. Base `eb5bcef`. Reviewers: correctness, testing, maintainability, project-standards, agent-native, learnings-researcher + security, reliability, adversarial, api-contract.

---

### P1 — High

|---|---|---|---|---|
| 1 | `debian/halos-core-containers.halos-manage-certs.service` + `debian/halos-core-containers.service` | **`WantedBy=` is a soft dependency — halos-core-containers.service starts even if cert provisioning fails.** Unit uses `WantedBy=halos-core-containers.service` in `[Install]` only. When halos-manage-certs.service fails (broken custom CA, openssl error), `Wants=` semantics let halos-core-containers.service start anyway → prestart.sh runs, writes Traefik YAML referencing `/certs/halos.{crt,key}` → Traefik container fails to bind :443 → silent partial outage (HTTP redirects to dead HTTPS). The PR body and `docs/CERTS.md` both assert "broken custom CA aborts the unit — and therefore blocks halos-core-containers.service start," but the current wiring doesn't deliver that. The pre-PR prestart.sh did fail-fast in the same `ExecStartPre`, so this is a regression in failure semantics. Fix: add `Requires=halos-manage-certs.service` + `After=halos-manage-certs.service` to `debian/halos-core-containers.service` `[Unit]`. | reliability, security, adversarial, correctness (4×) | 0.95 |

### P3 — Low

|---|---|---|---|---|
| 2 | `debian/halos-core-containers.halos-manage-certs.service:3` | `Documentation=` points at repo root; should point at `docs/CERTS.md` so `systemctl status` / `systemctl cat` operators land on the right doc. | maintainability | 0.62 |
| 3 | `prestart.sh:70-75` | 6-line explanatory comment about the new unit duplicates `docs/CERTS.md` and will rot if the unit is ever renamed. Shorten to a one-liner pointer. | maintainability, adversarial | 0.72 |
| 4 | `tests/test-halos-manage-certs.sh:119-141` | `test_idempotent_rerun_preserves_leaf_and_sentinel` only verifies leaf/CA inodes — sentinel content is never read or compared. Name claims more than the assertion delivers. Add a `cat .domain` byte-comparison. | testing | 0.93 |
| 5 | `tests/test-halos-manage-certs.sh` | No test for the `halos_ca_fingerprint \|\| exit 1` guard at script lines 110-114 — the load-bearing protection against sentinel-corruption infinite-loop. Add a test with a corrupted CA cert. | testing | 0.81 |
| 6 | `tests/test-halos-manage-certs.sh` | IP SAN branch (script lines 173-178) never exercised — all tests use DNS-only `hostnames.conf`. Add a test with at least one IP entry. | testing | 0.88 |
| 7 | `tests/test-halos-manage-certs.sh` | No happy-path test for operator-supplied custom CA — only the broken-files-abort path is covered. The `halos_ca_select_active` custom-takes-precedence branch could regress invisibly. | testing | 0.85 |
| 8 | `tests/test-halos-manage-certs.sh` | No tests parse cert PEM content — no validation of CN, SANs, validity period, basicConstraints, EKU. A regression producing a 0-byte or malformed cert would still pass existence checks. | testing | 0.92 |
| 9 | `assets/halos-manage-certs:27` | PACKAGE_NAME default at line 27 happens *before* env.defaults sourcing under `set -a`, so env.defaults clobbers any caller-supplied PACKAGE_NAME. Works in tests only because they point HALOS_ETC_DIR at an empty dir. Clarify the docstring or move the default *after* the env source. | maintainability, correctness | 0.85 |
| 10 | `assets/halos-manage-certs` | 5 HALOS_* test-injection env vars are documented in the script header but not gated by a `HALOS_TESTMODE=1` sentinel. Production exposure is low (systemd doesn't pass env through by default; an attacker who can write `/etc/systemd/system/.../override.conf` already has root), but the seams are a defense-in-depth concern. Consider path validation or test-mode gating. | security, adversarial | 0.65 |

---

- **`halos_cockpit_install_leaf` group-absent race.** If `cockpit-ws` group doesn't exist at first run, file lands as `root:root 0640` — unreadable to cockpit-tls until the next run. Pre-existing behavior in `lib-ca.sh`; the lib already logs NOTICE.
- **postinst re-enables on every upgrade** — same pattern as the existing halos-core-containers.service enable.
- **Custom CA dir permission race** — operator could create `/etc/halos/ca` with 0755 between drop and first run. Pre-existing; mitigated by postinst's chmod 0700 on install.

### Renewal gap (advisory)

Adversarial flagged that without #140's timer, halos-manage-certs.service only runs at boot/manual-start — a device that never reboots still hits the 825-day expiry. Correctly scoped: PR body defers the periodic-fire path to #140 (which is the planned next PR). Not blocking.

### Coverage

- **project-standards** returned clean. Issue link present, no VERSION bump (correct per release-cycle policy), no protected-path edits, no hard-coded hostnames.
- **agent-native** returned clean. `systemctl start halos-manage-certs.service` and `journalctl -u halos-manage-certs.service` are agent-accessible.

- `learning_auxiliary_prestart_calls.md` — `|| WARN` pattern preserved on both auxiliary calls. ✓
- `2026-05-24-set-e-cmdsubst-blind-spot.md` — guard pattern applied at the `CA_FINGERPRINT=$(...) || exit 1` site. ✓
- `feedback_debian_rules_asset_dirs.md` — new script installed via explicit `install -D` (the `assets/*/` glob would miss the top-level file). ✓
- `openssl-version-ci-vs-prod.md` — test suite uses `openssl -not_before` / `-not_after`. Verify CI green (debian:trixie container) before merge.

---

**Verdict: Not ready as-is. P1 #1 needs fixing before merge — the docs make a contract the wiring doesn't deliver. P3 doc/test fixes (#2–#9) are worth folding in same PR.**

mairas:

Fixes pushed in fa23800. 11/11 halos-manage-certs + 76/76 lib-ca + 39/39 lib-hostnames pass locally.

### P1
- ✅ **#1 Wants→Requires** — fa23800, `debian/halos-core-containers.service:3-4`. Added `Requires=halos-manage-certs.service` + `After=halos-manage-certs.service` so cert-provisioning failure now blocks the container stack with a clear error rather than producing a silent partial outage. Kept `WantedBy=halos-core-containers.service` on the new unit's `[Install]` — harmless given the explicit Requires=, and keeps the wants/ symlink visible to operators inspecting unit state.

### P3
- ✅ **#2 Documentation= URL** — fa23800, `debian/halos-core-containers.halos-manage-certs.service:3`. Now points at `docs/CERTS.md` directly.
- ✅ **#3 prestart.sh comment trim** — fa23800, `prestart.sh:70`. Now a one-liner pointer to `docs/CERTS.md`.
- ✅ **#4 Sentinel content assertion in idempotent test** — fa23800, `tests/test-halos-manage-certs.sh:120-148`. Added byte-comparison on `.domain` alongside the existing inode checks.
- ✅ **#5 Fingerprint-guard test** — fa23800, new `test_fingerprint_guard_exits_when_helper_fails`. Stubs `halos_ca_fingerprint` via the `HALOS_LIB_CA` seam and asserts the script exits non-zero.
- ✅ **#6 IP SAN test** — fa23800, new `test_leaf_includes_ip_san_when_hostnames_conf_has_ip`. Validates the previously-dead IP iteration branch in the SAN builder.
- ✅ **#7 Custom-CA happy path** — fa23800, new `test_valid_custom_ca_takes_precedence_over_auto`. Creates a valid operator CA, asserts `mode=custom` is logged, and verifies the leaf against it.
- ✅ **#8 Cert structural validation** — fa23800, new `test_leaf_has_expected_cert_structure`. Parses CN, SAN, basicConstraints, EKU, and validity (≤825 days).
- ✅ **#9 Env precedence docstring** — fa23800, `assets/halos-manage-certs:30-37`. Rewrote the comment to spell out the precedence explicitly (`env.defaults > env > caller-supplied env > line defaults`) and the test workaround (point HALOS_ETC_DIR at an empty tmpdir).
- ⏭ **#10 HALOS_LIB_CA hardening** — skipped. Production exposure requires write access to `/etc/systemd/system/...`, which is already root-only; an attacker with that capability has more direct paths than the env-injection seam.

Pre-existing items from the review (cockpit-ws group race, postinst re-enable behavior, custom CA dir permission race) are not addressed in this PR — they predate the extraction.

PR #143, opened 2026-05-26
feat(certs): periodic renewal timer + cockpit reload on leaf change (closes #140)

### LINKED ISSUE #140: Add periodic cert renewal (systemd timer) so long-running devices don't silently expire their leaf

mairas:
Closes #140.

## What

A periodic activation trigger for `halos-manage-certs.service` (from #139), plus a Cockpit auto-reload when the leaf actually rotates. Closes the renewal gap for devices that boot once and run uninterrupted for >824 days — without a timer the leaf silently expires and Apple Secure Transport clients start rejecting TLS.

## Files

| File | Change |
|---|---|
| `debian/halos-core-containers.halos-manage-certs.timer` | new — `OnBootSec=15min`, `OnUnitActiveSec=24h`, `RandomizedDelaySec=1h`, `Persistent=true`. Auto-installed by `dh_installsystemd`. |
| `assets/halos-manage-certs` | add post-cockpit-install block that emits the cockpit-reload signal **only when** `NEED_LEAF=true`. |
| `debian/postinst` | `systemctl enable halos-manage-certs.timer` alongside the existing service enables. |
| `docs/CERTS.md` | document the timer cadence, the inspect/force commands, and the Cockpit auto-reload-on-change behavior. |
| `tests/test-halos-manage-certs.sh` | 3 new tests for the reload-decision branch (fires on leaf change, no-op on idempotent rerun, no-op when cockpit-ws absent). |

## Why a timer, not a separate service

The original #140 body proposed a new `halos-cert-renewal.service` that would call `halos-manage-certs`. After #139 landed, that's pure duplication: the existing oneshot is already idempotent and already includes the `halos_ca_leaf_needs_renewal` gate (from #141). The timer just adds a second activation trigger — same script, same unit, two triggers (boot/restart ordering + periodic).

#140's issue body was rewritten to reflect this minimalism before implementation.

## Why gate the Cockpit reload on `NEED_LEAF=true`

Without the gate, a timer firing every 24h would bounce `cockpit.socket` daily even when no rotation happened — exactly the no-op thrash the gate exists to avoid. Once a device is steady-state, the timer is silent on `:9090`.

## How the cockpit-reload-decision branch is tested

The reload itself depends on `/run/systemd/system` and `cockpit.socket` being enabled, neither of which exists in the shell test environment. The tests assert the **log line** that signals the script took the reload branch (`"Leaf rotated; signaling cockpit.socket to reload..."`), which fires before the systemd guards and is therefore assertable without a real systemd:

- `test_cockpit_reload_signal_emitted_on_leaf_change` — first-boot run rotates the leaf → signal fires
- `test_cockpit_reload_signal_skipped_on_idempotent_rerun` — second back-to-back run does not rotate → signal does not fire (and `"Using existing leaf"` does)
- `test_cockpit_reload_signal_skipped_when_cockpit_dir_absent` — belt-and-suspenders: even with a rotation, no signal when cockpit-ws isn't installed

## Acceptance criteria (from #140)

- [x] `halos-manage-certs.timer` lands at `debian/halos-core-containers.halos-manage-certs.timer`
- [x] `debian/postinst` enables the timer alongside the existing service enable
- [ ] `systemctl list-timers` on a deployed device shows the timer firing on schedule — verifies post-deploy
- [x] `halos-manage-certs` script reloads `cockpit.socket` only when the leaf was re-signed (not on every no-op run)
- [x] Tests cover: (a) the cockpit-reload-trigger path fires when leaf changes, (b) no reload on idempotent no-op rerun (+1 bonus case for cockpit-ws absent)
- [x] `docs/CERTS.md` updated: timer cadence, inspect commands (`systemctl list-timers`, `journalctl -u halos-manage-certs.service`), and cockpit-reload-on-change behavior

## Test plan

- [x] `./run test` — 14/14 halos-manage-certs (was 11) + 76/76 lib-ca + 39/39 lib-hostnames pass locally
- [x] `bash -n` on the script
- [ ] CI run
- [ ] Smoke test on `halosdev.local` after package build:
  - `systemctl list-timers halos-manage-certs.timer` — shows next fire timestamp
  - `systemctl status halos-manage-certs.timer` — `active (waiting)`
  - `sudo systemctl start halos-manage-certs.service` — runs cleanly, logs "Using existing leaf" since the boot-time run already provisioned
  - Force a rotation (e.g., touch hostnames.conf to change the sentinel, then `systemctl start`) — verify journal shows "signaling cockpit.socket to reload" and `:9090` picks up the new leaf without manual restart

## Follow-ups

None expected — this closes out the cert-lifecycle umbrella (#141 added the renewal helper + Apple cap, #139 extracted the unit, #140 wires periodic activation + cockpit auto-reload).

### DIFF
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -204,30 +204,35 @@ fi
 # cockpit-ws dir doesn't exist — the cockpit package is only pulled in via
 # halos-cockpit-config's Recommends, so install-order variance can leave the
 # dir absent for a window.
+COCKPIT_INSTALL_SUCCEEDED=false
 if [ -d "${COCKPIT_WS_CERTS_DIR}" ]; then
     echo "Installing Cockpit leaf cert override at ${COCKPIT_LEAF_OVERRIDE}..."
     # Tolerate failure: cockpit override is auxiliary. Upstream's
     # 0-self-signed.cert remains the safety-net floor. Letting a transient
     # cockpit-install failure abort would mean cert management never
     # completes successfully, blocking halos-core-containers from starting.
-    if ! halos_cockpit_install_leaf "${CERT_FILE}" "${KEY_FILE}" "${COCKPIT_LEAF_OVERRIDE}"; then
+    if halos_cockpit_install_leaf "${CERT_FILE}" "${KEY_FILE}" "${COCKPIT_LEAF_OVERRIDE}"; then
+        COCKPIT_INSTALL_SUCCEEDED=true
+    else
         echo "WARNING: Cockpit leaf cert install failed; :9090 keeps its previous cert (upstream self-signed if no prior override)." >&2
     fi
 else
     echo "Skipping Cockpit leaf cert install: ${COCKPIT_WS_CERTS_DIR} does not exist (cockpit-ws not yet installed)"
 fi
 
-# When the timer-driven run rotates the leaf, cockpit-tls keeps the OLD
-# cert in memory until the next cockpit.socket activation. Force a reload
-# so :9090 picks up the fresh leaf immediately. Gated on NEED_LEAF=true
-# so no-op timer fires don't bounce :9090 — the typical case once a
-# device is steady-state.
+# When the timer-driven run rotates the leaf AND the new override was
+# actually installed, cockpit-tls keeps the OLD cert in memory until the
+# next cockpit.socket activation. Force a reload so :9090 picks up the
+# fresh leaf immediately. Gated on NEED_LEAF=true so no-op timer fires
+# don't bounce :9090, and on COCKPIT_INSTALL_SUCCEEDED so a failed
+# install doesn't trigger a pointless reload (cockpit-tls would just
+# re-read the same stale/missing override).
 #
 # Gated on systemd being available AND cockpit.socket being enabled so the
 # script remains usable from tests / minimal environments where neither
 # exists. The "Leaf rotated; signaling cockpit.socket to reload" log line
 # is the test signal — assertable without depending on a real systemd.
-if [ "${NEED_LEAF}" = true ] && [ -d "${COCKPIT_WS_CERTS_DIR}" ]; then
+if [ "${NEED_LEAF}" = true ] && [ "${COCKPIT_INSTALL_SUCCEEDED}" = true ]; then
     echo "Leaf rotated; signaling cockpit.socket to reload..."
     if [ -d /run/systemd/system ] && command -v systemctl >/dev/null 2>&1; then
         if systemctl is-enabled cockpit.socket >/dev/null 2>&1; then
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -217,4 +217,29 @@ else
     echo "Skipping Cockpit leaf cert install: ${COCKPIT_WS_CERTS_DIR} does not exist (cockpit-ws not yet installed)"
 fi
 
+# When the timer-driven run rotates the leaf, cockpit-tls keeps the OLD
+# cert in memory until the next cockpit.socket activation. Force a reload
+# so :9090 picks up the fresh leaf immediately. Gated on NEED_LEAF=true
+# so no-op timer fires don't bounce :9090 — the typical case once a
+# device is steady-state.
+#
+# Gated on systemd being available AND cockpit.socket being enabled so the
+# script remains usable from tests / minimal environments where neither
+# exists. The "Leaf rotated; signaling cockpit.socket to reload" log line
+# is the test signal — assertable without depending on a real systemd.
+if [ "${NEED_LEAF}" = true ] && [ -d "${COCKPIT_WS_CERTS_DIR}" ]; then
+    echo "Leaf rotated; signaling cockpit.socket to reload..."
+    if [ -d /run/systemd/system ] && command -v systemctl >/dev/null 2>&1; then
+        if systemctl is-enabled cockpit.socket >/dev/null 2>&1; then
+            if ! systemctl reload-or-restart cockpit.socket; then
+                echo "WARNING: cockpit.socket reload-or-restart failed; :9090 will pick up the new leaf at its next activation." >&2
+            fi
+        else
+            echo "Skipping cockpit.socket reload: cockpit.socket not enabled"
+        fi
+    else
+        echo "Skipping cockpit.socket reload: systemd not running"
+    fi
+fi
+
 echo "HaLOS cert management complete"
--- a/debian/halos-core-containers.halos-manage-certs.service
+++ b/debian/halos-core-containers.halos-manage-certs.service
@@ -5,8 +5,15 @@ After=local-fs.target
 Before=halos-core-containers.service
 
 [Service]
+# Type=oneshot WITHOUT RemainAfterExit so the timer's `start` action
+# actually re-runs ExecStart on each fire. systemd's `start` on a
+# RemainAfterExit=yes oneshot in active(exited) state is a documented
+# no-op, which would silently break periodic renewal. The Requires= +
+# After= chain on halos-core-containers.service still gates correctly:
+# a successfully-exited oneshot satisfies Requires= even after it goes
+# naturally inactive (verified empirically — only failed exits drop
+# the consumer).
 Type=oneshot
-RemainAfterExit=yes
 ExecStart=/usr/lib/halos-core-containers/halos-manage-certs
 StandardOutput=journal
 StandardError=journal
--- a/debian/halos-core-containers.halos-manage-certs.timer
+++ b/debian/halos-core-containers.halos-manage-certs.timer
@@ -1,0 +1,22 @@
+[Unit]
+Description=Periodic HaLOS device cert renewal check
+Documentation=https://github.com/halos-org/halos-core-containers/blob/main/docs/CERTS.md
+Requires=halos-manage-certs.service
+
+[Timer]
+# Wait 15 min after boot for the rest of the stack to settle before the
+# first timer-driven run; the boot-time activation via
+# halos-core-containers.service has already provisioned certs by then.
+OnBootSec=15min
+# Re-check every 24h. Well below HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS (60d),
+# so even fortnightly-rebooted devices get multiple chances to renew.
+OnUnitActiveSec=24h
+# Spread fleet-wide hits over a 1h window to avoid synchronized renewal
+# storms.
+RandomizedDelaySec=1h
+# Catch up if the device was off when the timer would have fired.
+Persistent=true
+Unit=halos-manage-certs.service
+
+[Install]
+WantedBy=timers.target
--- a/debian/halos-core-containers.halos-manage-certs.timer
+++ b/debian/halos-core-containers.halos-manage-certs.timer
@@ -14,8 +14,14 @@ OnUnitActiveSec=24h
 # Spread fleet-wide hits over a 1h window to avoid synchronized renewal
 # storms.
 RandomizedDelaySec=1h
-# Catch up if the device was off when the timer would have fired.
-Persistent=true
+# Deliberately NOT Persistent=true: Raspberry Pi has no RTC, so first
+# boot starts with a fake-hwclock timestamp (often 1970). With Persistent
+# set, the post-NTP clock jump from 1970 → 2026 would trigger a catch-up
+# fire storm of "missed" runs. Boot-time renewal still happens because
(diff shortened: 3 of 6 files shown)


mairas:

**Scope:** 5 files, +136/-8 LOC. Base `69de1b8`. Reviewers: correctness, testing, maintainability, project-standards, agent-native, learnings-researcher + reliability, adversarial, api-contract.

---

### P1 — High

|---|---|---|---|---|
| 1 | `debian/halos-core-containers.halos-manage-certs.service` | **The timer will silently no-op once the service has run.** `Type=oneshot + RemainAfterExit=yes` leaves the unit in `active (exited)` after the boot-time run. The timer's default activation semantic is `systemctl start`, and `start` on an already-active oneshot+RemainAfterExit unit is a no-op — `ExecStart` does NOT re-run. The timer's `OnBootSec=15min`, `OnUnitActiveSec=24h`, and every subsequent fire return success without doing anything. The renewal mechanism this PR is supposed to deliver never runs after the initial boot-time activation. **Verified empirically on halosdev.local** with a synthetic unit: Run 1 (start, fresh) appends to log; Run 2 (start, already active(exited)) does NOT append; Run 3 (restart, not start) does append. The fix is to drop `RemainAfterExit=yes`: a second test confirmed the `Requires=` chain from halos-core-containers.service remains satisfied when the oneshot dep completes and goes naturally inactive — the consumer stays active. | correctness, adversarial (2×) | 0.95 (empirically verified) |

### P2 — Moderate

|---|---|---|---|---|
| 2 | `assets/halos-manage-certs:213+230` | **Cockpit reload fires even when the override install failed.** `halos_cockpit_install_leaf` can fail (chown/chmod/disk/mv errors) — failure logs WARNING and continues. The reload block at line 230 gates only on `NEED_LEAF=true && [ -d COCKPIT_WS_CERTS_DIR ]`, with no check on install success. Result: `systemctl reload-or-restart cockpit.socket` fires even though the override file was never written. Cockpit either falls back to upstream self-signed (first boot, no prior override) or keeps serving the stale prior override. Fix: set a `COCKPIT_INSTALL_SUCCEEDED` flag inside the install block and add it to the reload gate. | correctness | 0.85 |
| 3 | `debian/halos-core-containers.halos-manage-certs.timer:17` | **`Persistent=true` triggers a renewal fire-storm on Pi devices after NTP clock-jump.** Pi hardware has no RTC (documented in `lib-ca.sh:77-85`). Boot sequence: fake-hwclock places system at 1970 → systemd records "last fire = 1970" → NTP corrects clock to 2026 → systemd re-evaluates Persistent timer → `last fire (1970) + 24h < now (2026)` → fires. Subsequent clock-related daemon-reloads can trigger additional fires in quick succession. The renewal check itself is idempotent (cheap when nothing needed), but the log spam and CPU bursts are real. Also: removing `Persistent=true` doesn't lose functionality — the cert manager already runs at every container-stack activation (via `Requires=` from #142), so a long-off device renews at boot regardless. | adversarial | 0.82 |

### P3 — Low

|---|---|---|---|---|
| 4 | `tests/test-halos-manage-certs.sh` | No test for the install-failure case combined with reload-decision (P2 #2). If a test stubbed `halos_cockpit_install_leaf` to fail and asserted the reload signal does NOT fire, the bug above would be caught. | correctness, testing | 0.78 |
| 5 | `tests/test-halos-manage-certs.sh` | `test_renewal_threshold_triggers_leaf_only_resign` (pre-existing) sets up the renewal-driven NEED_LEAF=true path but doesn't grep for the reload signal. A regression that breaks reload on renewal-driven re-signs (vs first-boot re-signs) would slip through. | testing | 0.75 |
| 6 | `debian/postinst:78` | `systemctl enable halos-manage-certs.timer` creates the symlink but doesn't START the timer on a running system. Operators upgrading mid-day won't see the first fire until next reboot. Add `systemctl start halos-manage-certs.timer` after the enable. | adversarial | 0.72 |
| 7 | `docs/CERTS.md` | No operator guidance on disabling / masking / rescheduling the timer (e.g., for maintenance windows). The `.timer` unit is now an operator-facing public surface. Document `systemctl disable/mask halos-manage-certs.timer` for opt-out. | api-contract | 0.68 |
| 8 | `assets/halos-manage-certs:230-243` | The 3-level nested if-then-else for the reload gate has three `Skipping...` else branches. Could be flattened with early-return pattern. Stylistic. | maintainability | 0.68 |

---

- **`Type=oneshot` semantics for cockpit.socket reload.** Adversarial flagged that reload-or-restart on a `.socket` unit falls through to restart (sockets don't support reload), which the api-contract reviewer framed as session disruption. In cockpit's actual architecture (socket-activated, per-connection cockpit-tls instances), the socket restart refreshes the LISTENER without affecting active per-connection cockpit-tls processes. New connections naturally pick up the new cert. So the reload is benign-but-possibly-unnecessary — the cert is on disk after install, new sessions read it fresh at activation. Not addressed here; would require Cockpit-architecture analysis beyond this PR's scope.

### Coverage

- **project-standards** returned clean. `Closes #140`, no VERSION bump (correct), no protected-path edits, no hard-coded hostnames.
- **agent-native** returned clean. All operator commands are standard systemd / journalctl.
- **The P1 finding required empirical verification** (synthetic systemd unit on halosdev.local) — confirmed real. Fix path also verified.

- `learning_auxiliary_prestart_calls.md` — the `|| WARN` pattern correctly preserved on the cockpit reload error path. ✓
- The timer/oneshot pattern: standard distro `.timer` units (apt-daily, logrotate, fstrim) all use `Type=oneshot` *without* `RemainAfterExit=yes` for exactly this reason — periodic re-firing requires the unit to go fully inactive between runs.

---

**Verdict: Not ready as-is. P1 #1 breaks the entire renewal mechanism — the timer would silently never re-run cert checks. P2 #2 and #3 are real bugs worth fixing in the same PR. P3 items are doc/test additions worth folding in.**

mairas:

Fixes pushed in aafb0ee. 16/16 halos-manage-certs + 76/76 lib-ca + 39/39 lib-hostnames pass locally. P1 verified empirically on halosdev.local before the fix landed.

### P1
- ✅ **#1 RemainAfterExit + timer no-op** — aafb0ee, `debian/halos-core-containers.halos-manage-certs.service:7`. Dropped `RemainAfterExit=yes`. Empirically verified that the Requires= chain from halos-core-containers.service still holds when the oneshot dep exits cleanly and goes naturally inactive — only failed exits propagate to the consumer.

### P2
- ✅ **#2 Reload fires when cockpit install fails** — aafb0ee, `assets/halos-manage-certs:209+230`. Added `COCKPIT_INSTALL_SUCCEEDED` flag set inside the install block; reload gate now requires both `NEED_LEAF=true` AND install success.
- ✅ **#3 Persistent=true → Pi RTC fire storm** — aafb0ee, `debian/halos-core-containers.halos-manage-certs.timer:18`. Dropped `Persistent=true`. Boot-time renewal still happens via the Requires= chain; losing Persistent only delays a post-power-cycle renewal by ≤24h, well below the 60-day threshold.

### P3
- ✅ **#4 No test for install-failure-skips-reload** — aafb0ee, new `test_cockpit_reload_signal_skipped_when_install_fails`. Stubs `halos_cockpit_install_leaf` via the HALOS_LIB_CA seam to return non-zero, asserts the reload signal is NOT emitted, and asserts the install-failure WARNING IS emitted.
- ✅ **#5 No test for renewal-driven reload** — aafb0ee, new `test_cockpit_reload_signal_emitted_on_renewal_threshold_resign`. Synthesizes a 30-day leaf + matching sentinel so only the renewal gate trips NEED_LEAF=true on the next run; asserts both the reload signal AND the "within renewal threshold" log line (to confirm the right code path was exercised, not a sentinel mismatch).
- ✅ **#6 postinst doesn't start timer on upgrade** — aafb0ee, `debian/postinst:81`. Added `systemctl start halos-manage-certs.timer` after the enable so upgrades on a long-uptime device begin the renewal schedule immediately.
- ✅ **#7 No docs on how to opt out of the timer** — aafb0ee, `docs/CERTS.md`. New "Disabling the timer" subsection covering `systemctl stop/disable/mask` for maintenance windows.
- ⏭ **#8 Nested if-then-else readability** — skipped (stylistic, no behavior change), per the review's own recommendation.

### Pre-existing (not addressed)
- Cockpit-architecture analysis of whether `systemctl reload-or-restart cockpit.socket` actually disrupts active per-connection cockpit-tls instances (likely benign given socket-activated per-connection design, but out of this PR's scope).

PR #145, opened 2026-05-26
fix(packaging): install halos-manage-certs.{service,timer} via --name=

mairas:
## What broke

#142 and #143 added `debian/halos-core-containers.halos-manage-certs.service` and `.timer` to the source tree, expecting `dh_installsystemd` to pick them up automatically. It doesn't: that tool's auto-detection only handles `debian/<pkg>.<extension>` (one file per extension per binary package), NOT `debian/<pkg>.<name>.<extension>`. Multi-service files for the same binary package need an explicit `dh_installsystemd --name=<name>` invocation.

Result: since #142 merged this morning, every published `.deb` has shipped the `halos-manage-certs` script at `/usr/lib/halos-core-containers/halos-manage-certs` but neither the service nor the timer unit. Confirmed on `halosdev.local` post-upgrade to 0.3.4-7:

```
$ dpkg -L halos-core-containers | grep -E 'systemd|halos-manage'
/usr/lib/halos-core-containers/halos-manage-certs
/usr/lib/systemd
/usr/lib/systemd/system
/usr/lib/systemd/system/halos-core-containers.service
```

No `halos-manage-certs.service`, no `halos-manage-certs.timer`. The activation chain from #142 (Requires= + After= on halos-core-containers.service) and the renewal timer from #143 are both completely inoperative. The package upgrades cleanly, but devices remain on the pre-#141 10-year-validity leaf that Apple rejects.

## Fix

Override `dh_installsystemd` in `debian/rules` to invoke it twice — once for the default (`halos-core-containers.service`) and once with `--name=halos-manage-certs` which picks up both `.service` and `.timer` for that name. Verified locally:

```
$ docker run --rm -v "$(pwd):/workspace" debian:trixie bash -c \
    "dpkg-deb -c /workspace/build/halos-core-containers_*_all.deb | grep systemd"
drwxr-xr-x root/root         0 ./usr/lib/systemd/
drwxr-xr-x root/root         0 ./usr/lib/systemd/system/
-rw-r--r-- root/root       704 ./usr/lib/systemd/system/halos-core-containers.service
-rw-r--r-- root/root       912 ./usr/lib/systemd/system/halos-manage-certs.service
-rw-r--r-- root/root      1312 ./usr/lib/systemd/system/halos-manage-certs.timer
```

## Test plan

- [x] Local build via `./run build` succeeds
- [x] `dpkg-deb -c` on the built .deb shows all three unit files
- [ ] CI green
- [ ] Apt pipeline publishes new version
- [ ] `apt upgrade` on `halosdev.local`; `dpkg -L halos-core-containers | grep manage-certs` shows the unit files; `systemctl status halos-manage-certs.service` shows `active (exited)`; `systemctl list-timers halos-manage-certs.timer` shows the next fire; `openssl x509 -in /var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.crt -noout -enddate` shows ≤824-day validity

## Severity

This regression has been live since #142 merged at 08:27 UTC today, affecting every device that upgraded to 0.3.4-7 or later. The blast radius is "renewal mechanism doesn't run" — pre-existing certs continue working (until their natural expiry), but the whole cert-lifecycle umbrella (#141 + #139 + #140) has been delivering zero functionality to upgraded devices.

## Follow-up

Should add packaging-validation to CI: `dpkg-deb -c build/*.deb | grep -q halos-manage-certs.service` post-build, fail if missing. Out of scope for this hotfix.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -64,21 +64,13 @@ EOF
             # Reload systemd to recognize new service
             systemctl daemon-reload
 
-            # Enable services. halos-manage-certs.service has
-            # WantedBy=halos-core-containers.service in [Install], so
-            # enabling it drops a wants/ symlink that pulls it in
-            # automatically whenever the container stack activates.
-            # halos-manage-certs.timer periodically re-fires the service
-            # so long-running devices renew their leaf before Apple's
-            # 825-day SSL ceiling expires it.
+            # halos-manage-certs.{service,timer} are enabled and started
+            # by #DEBHELPER# below via dh_installsystemd --name= in
+            # debian/rules. The service's WantedBy=halos-core-containers
+            # .service pulls it in whenever the stack activates; the
+            # timer renews the leaf on a 24h cadence to stay under
+            # Apple's 825-day SSL ceiling.
             systemctl enable ${PKG_NAME}.service || true
-            systemctl enable halos-manage-certs.service || true
-            systemctl enable halos-manage-certs.timer || true
-            # Start the timer immediately so an upgrade on a long-uptime
-            # device doesn't wait for the next reboot to begin its
-            # 24h-cadence renewal schedule. Safe to call even if the
-            # timer is already running.
-            systemctl start halos-manage-certs.timer || true
         fi
         ;;
 esac
--- a/debian/rules
+++ b/debian/rules
@@ -72,4 +72,12 @@ override_dh_auto_install:
 override_dh_auto_test:
 	# Skip tests - validated manually
 
-# dh_installsystemd handles service file installation from debian/halos-core-containers.service
+# dh_installsystemd handles service file installation from debian/halos-core-containers.service.
+# Additional unit files (multiple service/timer for the same binary package) require
+# explicit --name= invocations — dh_installsystemd's autodetect only picks up
+# debian/<pkg>.<extension>, NOT debian/<pkg>.<name>.<extension>. Without the
+# --name=halos-manage-certs call below, debian/halos-core-containers.halos-manage-certs.{service,timer}
+# would be present in the source tree but silently absent from the built .deb.
+override_dh_installsystemd:
+	dh_installsystemd
+	dh_installsystemd --name=halos-manage-certs


mairas:

**Scope:** 1 file, +9/-1 in `debian/rules`. Base `2f51f0b`. Reviewers: correctness, testing, maintainability, project-standards, agent-native, learnings-researcher (always-on only — no conditional triggers for a 9-LOC packaging fix).

### P2 — Moderate

|---|---|---|---|---|
| 1 | `.github/workflows/pr.yml` + `tools/build-all.sh` | **No CI validation that systemd unit files actually ship in the .deb.** The whole class of "source has the file, .deb doesn't" went undetected for ~2h across two PRs (#142 + #143). After this fix lands, the same class of regression is still invisible — `./run test` exercises only the shell suites; nothing inspects the built artifact. Fix: `dpkg-deb -c build/*.deb \| grep -q halos-manage-certs.service` after `./run build` in CI (or as a step in `.github/actions/run-tests`) would prevent recurrence. | testing, maintainability (2×) | 0.85 |

### P3 — Low

|---|---|---|---|---|
| 2 | `debian/rules:74-83` | Comment explains the gap but doesn't tell a future maintainer adding a third named unit what to do. Add a one-liner: "to ship `debian/<pkg>.<name>.{service,timer}` for any additional `<name>`, append another `dh_installsystemd --name=<name>` invocation below." | maintainability | 0.75 |
| 3 | `debian/rules:81-83` | The two `dh_installsystemd` invocations don't share flags. If the default ever needs args, both must be updated in lockstep — silently divergent flags = silently divergent behavior. Marginal risk; nothing to do today. | maintainability | 0.65 |
| 4 | PR body | References #142/#143 in prose but no `Closes #`/`Refs #` syntax. This is a regression hotfix without a tracking issue. If strict AGENTS.md compliance is wanted, file a regression issue and `Closes` it; otherwise the prose reference is informally fine. | project-standards | 0.70 |

- **Double-enable in postinst.** Manual `systemctl enable halos-manage-certs.{service,timer}` now coexists with auto-generated `#DEBHELPER#` enable snippets from the new `dh_installsystemd --name=` call. Correctness reviewer confirms idempotent — `deb-systemd-helper` is designed to tolerate manual `systemctl enable`.

- **`feedback_debian_rules_asset_dirs.md`** — same family of bug: hardcoded list in `debian/rules` drifts from compose-declared sources → silent failure. Generalizable lesson: any `debian/rules` change that *enumerates* installable inputs needs a build-artifact CI check, not just a shell-test CI check.
- The `dh_installsystemd --name=` convention (multi-unit packages) is now twice-burned. Worth a `docs/solutions/` entry.

### Agent-native

Clean. Build-system fix; no operator surface change.

---

**Verdict: Ready to merge.** Fix is correct and locally verified (`dpkg-deb -c` shows all three unit files ship). The P2 CI gap is real but meta — about test infrastructure, not this fix.

mairas:
## Review resolution

Addressing the "Double-enable in postinst" item from the review synthesis, which I had previously parked under "Pre-existing / out-of-scope." On reflection that label was wrong — global guidance forbids phantom follow-ups, and this fix is one file change adjacent to the diff already under review, so deferring it was the wrong call.

| Review finding | Commit | Notes |
|---|---|---|
| Double-enable in postinst | `2643f30` | Removed manual `systemctl enable halos-manage-certs.{service,timer}` and `systemctl start halos-manage-certs.timer` from `debian/postinst`. These are now covered by `#DEBHELPER#` expansion of the new `dh_installsystemd --name=halos-manage-certs` override added in `e10bd11`. Comment block updated to point future readers at `debian/rules`. |

The `systemctl enable ${PKG_NAME}.service` line is left alone — it predates this PR and isn't part of the regression being fixed. If we want to lean on `dh_installsystemd` for the default service too, that's a separate cleanup PR.

**Verification**: Extracted maintainer scripts from the rebuilt `.deb` and confirmed `#DEBHELPER#` now emits:
- `deb-systemd-helper enable 'halos-manage-certs.service'`
- `deb-systemd-helper enable 'halos-manage-certs.timer'`
- `deb-systemd-invoke start 'halos-manage-certs.service' 'halos-manage-certs.timer'`

Shell-library test suite (39 tests) still passes.

Remaining P2/P3 items from the synthesis (CI artifact validation, doc nudge on `dh_installsystemd --name=`, flag-sharing risk) are genuinely out of scope for this hotfix and will need follow-up issues if we want to track them.

PR #148, opened 2026-05-26
fix(certs): touch Traefik dynamic config on leaf rotation

### LINKED ISSUE #147: halos-manage-certs doesn't signal Traefik on leaf rotation
## Problem

After `halos-manage-certs.service` rotates the device leaf certificate, Traefik continues serving the previous cert in-memory until the container is restarted. The current code path signals `cockpit.socket` to reload but takes no equivalent action for Traefik.

This means:

- **Fresh installs**: invisible. Traefik starts after the leaf is on disk, so the first cert it loads is correct.
- **Renewal via the timer** (the main reason #140 exists): visible. The next time the timer fires and produces a new leaf — whether through the renewal window, a hostname-list change, or a manual `systemctl start halos-manage-certs.service` — Traefik keeps serving the old cert until the operator restarts `halos-core-containers.service`.
- **Operator-triggered rotation** (e.g., delete leaf → re-run service to escape a pre-#141 long-validity cert): visible. Observed firsthand on `halosdev.local` 2026-05-26.

## Reproduction (halosdev, 0.3.4-9)

```
$ sudo rm /var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.{crt,key}
$ sudo systemctl start halos-manage-certs.service
$ echo | openssl s_client -servername halosdev.local -connect halosdev.local:443 2>/dev/null | openssl x509 -noout -dates
# ← still shows the previously-cached cert from before deletion
$ sudo systemctl restart halos-core-containers.service
$ echo | openssl s_client -servername halosdev.local -connect halosdev.local:443 2>/dev/null | openssl x509 -noout -dates
# ← now shows the new leaf
```

## Why it matters

This is what makes the renewal pipeline actually deliver the renewed leaf to clients. Without it, the renewal mechanism still rotates files on disk, but the user-facing surface (Traefik on 443) serves the stale cert until the next stack restart — which on a long-uptime device might be months. That defeats the point of #140 / #141 for the device's primary entrypoint.

Cockpit is already covered (`halos_cockpit_install_leaf` signals `cockpit.socket` after rotation); Traefik is the asymmetry.

## Out of scope

- Re-issuance gating logic (already correct per #141)
- Cockpit reload (already handled)
- Force-renewing pre-#141 long-validity leaves on upgraded devices — design intent is a one-shot operator action (`rm halos.{crt,key}` + `systemctl start halos-manage-certs.service`), this issue makes that one-shot actually take effect without a stack restart

## Suggested approach

Two options, both defensible — pick when implementing:

**Option A: signal Traefik to reload.** Traefik watches its dynamic config but doesn't watch cert files referenced by it. Sending SIGHUP causes Traefik to re-read the dynamic config (which re-reads referenced files including certs). Implementable as `docker kill --signal=HUP traefik` from the rotation script. Cheap, no container restart, no service interruption.

**Option B: bounce the Traefik container.** `docker compose restart traefik` via the package's compose root. Heavier, brief connection drop, but minimally coupled to Traefik internals.

A is preferred unless SIGHUP is unreliable in the Traefik version we ship; B is the conservative fallback.

Either way the implementation lives next to `halos_cockpit_install_leaf` in `assets/lib-ca.sh` (or `halos-manage-certs`), gated on the same "leaf rotated" condition that already signals Cockpit.

## Severity / release blocker

Release-blocker per @mairas. Should land before the next stable cut (0.4.x or whatever follows the 0.3.4 cycle).

mairas:
## Summary

After `halos-manage-certs` rotates the device leaf, Traefik kept serving the previous cert in-memory until `halos-core-containers.service` was restarted. The script already signals `cockpit.socket` to reload but did nothing equivalent for Traefik — the asymmetry called out in #147. On long-uptime devices this defeats the renewal pipeline's purpose on the device's primary entrypoint (`:443`).

Traefik watches its **dynamic-config file** via inotify but does **not** watch the cert files referenced from `defaultCertificate.certFile`/`keyFile` ([traefik/traefik#5495](https://github.com/traefik/traefik/issues/5495)). Touching the YAML forces Traefik to re-evaluate the dynamic config, which re-reads the cert from disk — no container restart, no connection drop.

## Why not SIGHUP (Option A in the issue)

The issue suggested SIGHUP as the preferred approach. I investigated and found this is **unsafe** on the v3.x line:

- [traefik/traefik#9993](https://github.com/traefik/traefik/pull/9993) added SIGHUP file-provider reload in v3.0.0.
- [traefik/traefik#11624](https://github.com/traefik/traefik/issues/11624) reports that SIGHUP **kills** Traefik on the v3.x line (confirmed on 3.3.3, unfixed at filing).
- Sending SIGHUP on every rotation would silently take `:443` down.

The touch-the-config-file approach is the documented Traefik pattern and is what the upstream community recommends for exactly this case. Inline comments in the script cite both upstream tickets.

## Behaviour

The new block mirrors the cockpit-reload block's gating:

| Path | Behaviour |
|------|-----------|
| `NEED_LEAF=true`, tls-default.yml exists | `touch` the YAML, log `Leaf rotated; touching Traefik dynamic config to reload certs ...` |
| `NEED_LEAF=true`, file missing (first boot) | Skip with log: `Skipping Traefik cert reload: ... does not exist (halos-core-containers prestart will generate it on next start)` — `halos-manage-certs.service` runs `Before=` `halos-core-containers.service`, so on first provisioning Traefik isn't running yet and will load the fresh leaf on its first start |
| `NEED_LEAF=false` (idempotent timer fire) | No touch — prevents 24h churn |
| `touch` itself fails (read-only FS etc) | WARN, don't abort (cert management has completed) |

## Tests

`tests/test-halos-manage-certs.sh` gains 4 cases that mirror the cockpit-reload battery:

- `test_traefik_reload_touch_fires_on_leaf_change` — first-run signing fires touch, mtime advances past sentinel
- `test_traefik_reload_skipped_on_idempotent_rerun` — second run with no changes does NOT touch
- `test_traefik_reload_skipped_when_dynamic_config_absent` — first-boot path: skip log, no file created
- `test_traefik_reload_fires_on_renewal_threshold_resign` — renewal-gate triggers reload (regression guard if someone gates touch on the sentinel branch only)

mtime assertions use `[ file -nt sentinel ]` instead of `stat -c '%Y'` so the suite runs on BSD `stat` (macOS dev) and GNU `stat` (CI) without divergence.

All 20 `test-halos-manage-certs.sh` tests pass; full repo `./run test` green (76 + 39 + 20 cases across lib-ca, lib-hostnames, halos-manage-certs).

## Reproduction & verification

Issue #147's reproduction steps:

\`\`\`
sudo rm /var/lib/container-apps/halos-core-containers/data/traefik/certs/halos.{crt,key}
sudo systemctl start halos-manage-certs.service
echo | openssl s_client -servername halosdev.local -connect halosdev.local:443 2>/dev/null \
  | openssl x509 -noout -dates
\`\`\`

Before this PR: dates unchanged until \`systemctl restart halos-core-containers.service\`.
After this PR (per test coverage + Traefik's documented watch behaviour): dates advance within the file-watcher's debounce window.

I have **not** smoke-tested this on \`halosdev.local\` (deploying the updated script to a shared device required permission I didn't have in this session). The unit tests fully cover the new branches and the upstream behaviour is well-documented. Happy to wire up a one-shot device smoke test on request — script swap + the repro above is trivial once the deb is built and deployed via the normal CI/CD pipeline.

## Out of scope (and intentionally so, per issue)

- Re-issuance gating (already correct, #141)
- Cockpit reload (already wired in this script)
- Pre-#141 long-validity leaf force-renewal mechanism — this PR is what makes the documented one-shot operator action (\`rm halos.{crt,key}\` + \`systemctl start halos-manage-certs.service\`) actually take effect on \`:443\` without a stack restart

Closes #147

## Test plan

- [x] \`./run test\` — all suites green locally
- [ ] CI passes
- [ ] Device smoke test against \`halosdev.local\` after CI/CD pipeline deploys the new \`.deb\` (operator-triggered repro per the issue)

### DIFF
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -277,7 +277,11 @@ fi
 if [ "${NEED_LEAF}" = true ]; then
     if [ -f "${TRAEFIK_TLS_CONFIG}" ]; then
         echo "Leaf rotated; touching Traefik dynamic config to reload certs at ${TRAEFIK_TLS_CONFIG}..."
-        if ! touch "${TRAEFIK_TLS_CONFIG}"; then
+        # -c (no-create): if the file is removed in the narrow window between
+        # the existence check above and the touch, do NOT create an empty
+        # placeholder Traefik can't parse. The check + -c together make the
+        # path strictly mtime-only.
+        if ! touch -c "${TRAEFIK_TLS_CONFIG}"; then
             echo "WARNING: touch ${TRAEFIK_TLS_CONFIG} failed; Traefik will keep serving the previous leaf until the next halos-core-containers.service restart." >&2
         fi
     else
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -21,6 +21,7 @@
 #   HALOS_LIB_CA               /usr/lib/<pkg>/lib-ca.sh
 #   HALOS_CUSTOM_CA_DIR        /etc/halos/ca
 #   HALOS_COCKPIT_WS_CERTS_DIR /etc/cockpit/ws-certs.d
+#   HALOS_TRAEFIK_DYNAMIC_DIR  /etc/halos/traefik-dynamic.d
 
 set -e
 
@@ -84,6 +85,8 @@ DOMAIN_FILE="${CERTS_DIR}/.domain"
 PUBLIC_CA_DIR="${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/public"
 COCKPIT_WS_CERTS_DIR="${HALOS_COCKPIT_WS_CERTS_DIR:-/etc/cockpit/ws-certs.d}"
 COCKPIT_LEAF_OVERRIDE="${COCKPIT_WS_CERTS_DIR}/99-halos.cert"
+TRAEFIK_DYNAMIC_DIR="${HALOS_TRAEFIK_DYNAMIC_DIR:-/etc/halos/traefik-dynamic.d}"
+TRAEFIK_TLS_CONFIG="${TRAEFIK_DYNAMIC_DIR}/tls-default.yml"
 
 mkdir -p "${CERTS_DIR}"
 
@@ -244,6 +247,41 @@ if [ "${NEED_LEAF}" = true ] && [ "${COCKPIT_INSTALL_SUCCEEDED}" = true ]; then
         fi
     else
         echo "Skipping cockpit.socket reload: systemd not running"
+    fi
+fi
+
+# Traefik (:443) reads halos.crt/halos.key via its file-provider dynamic
+# config at ${TRAEFIK_TLS_CONFIG}. Its file watcher fires on changes to
+# the dynamic-config FILE but does NOT watch the referenced cert files,
+# so a fresh leaf on disk is ignored until the YAML is touched. Bump the
+# YAML's mtime to force Traefik to re-evaluate the dynamic config, which
+# re-reads the cert from disk — see traefik/traefik#5495 for the
+# upstream limitation.
+#
+# Why not SIGHUP: PR traefik/traefik#9993 (v3.0.0) intended SIGHUP for
+# file-provider reload, but issue traefik/traefik#11624 reports SIGHUP
+# *kills* Traefik on the v3.x line (confirmed on 3.3.3, unfixed at the
+# time of writing). Touch is the safe path; SIGHUP would silently take
+# the proxy down on every rotation.
+#
+# Gated on NEED_LEAF=true so a no-op timer fire doesn't churn Traefik's
+# config every 24h. On first boot the dynamic-config file doesn't yet
+# exist (halos-manage-certs.service runs Before= halos-core-containers,
+# which is what generates the file via prestart.sh); the missing-file
+# branch logs and skips because Traefik isn't running yet and will load
+# the freshly-signed leaf on its first start.
+#
+# The "Leaf rotated; touching Traefik dynamic config to reload certs"
+# log line is the test signal — assertable without depending on a real
+# Traefik container.
+if [ "${NEED_LEAF}" = true ]; then
+    if [ -f "${TRAEFIK_TLS_CONFIG}" ]; then
+        echo "Leaf rotated; touching Traefik dynamic config to reload certs at ${TRAEFIK_TLS_CONFIG}..."
+        if ! touch "${TRAEFIK_TLS_CONFIG}"; then
+            echo "WARNING: touch ${TRAEFIK_TLS_CONFIG} failed; Traefik will keep serving the previous leaf until the next halos-core-containers.service restart." >&2
+        fi
+    else
+        echo "Skipping Traefik cert reload: ${TRAEFIK_TLS_CONFIG} does not exist (halos-core-containers prestart will generate it on next start)"
     fi
 fi
 
--- a/prestart.sh
+++ b/prestart.sh
@@ -74,7 +74,14 @@ DYNAMIC_DIR="/etc/halos/traefik-dynamic.d"
 DYNAMIC_SRC_DIR="${SCRIPT_DIR}/assets/traefik/dynamic"
 mkdir -p "${DYNAMIC_DIR}"
 
-# Generate dynamic TLS configuration
+# Generate dynamic TLS configuration.
+#
+# IMPORTANT: assets/halos-manage-certs hardcodes the filename "tls-default.yml"
+# in TRAEFIK_TLS_CONFIG and touches it on leaf rotation to force Traefik's
+# file-watcher to re-read the cert. Renaming this file requires updating that
+# script as well, or rotation-driven cert reload silently falls into the
+# "Skipping Traefik cert reload" branch and Traefik keeps serving the stale
+# leaf until the next halos-core-containers.service restart.
 TLS_CONFIG_FILE="${DYNAMIC_DIR}/tls-default.yml"
 cat > "${TLS_CONFIG_FILE}" << EOF
 # Default TLS certificate configuration
--- a/tests/test-halos-manage-certs.sh
+++ b/tests/test-halos-manage-certs.sh
@@ -68,9 +68,24 @@ run_test() {
 # COCKPIT_WS_CERTS_DIR is created so the cockpit-override branch runs.
 new_scratch() {
     local d="$TMPDIR_ROOT/$1"
-    mkdir -p "$d/data" "$d/etc" "$d/custom-ca" "$d/cockpit-certs"
+    mkdir -p "$d/data" "$d/etc" "$d/custom-ca" "$d/cockpit-certs" "$d/traefik-dynamic"
     printf '%s\n' "fixture.test" > "$d/hostnames.conf"
     printf '%s' "$d"
+}
+
+# Synthesize the Traefik dynamic config file that prestart.sh would have
+# created when halos-core-containers.service starts. Production wiring:
+# the halos-manage-certs touch branch is gated on the file existing, so
+# tests that exercise the "touch fired" path need this preplaced.
+_seed_traefik_tls_config() {
+    cat > "$1/traefik-dynamic/tls-default.yml" <<'EOF'
+tls:
+  stores:
+    default:
+      defaultCertificate:
+        certFile: /certs/halos.crt
+        keyFile: /certs/halos.key
+EOF
 }
 
 # Invoke the script with HALOS_* env overrides pointing at the scratch
@@ -84,6 +99,7 @@ run_script() {
     HALOS_LIB_CA="$LIB_CA" \
     HALOS_CUSTOM_CA_DIR="$d/custom-ca" \
     HALOS_COCKPIT_WS_CERTS_DIR="$d/cockpit-certs" \
+    HALOS_TRAEFIK_DYNAMIC_DIR="$d/traefik-dynamic" \
     HALOS_HOSTNAMES_FILE="$d/hostnames.conf" \
     PACKAGE_NAME="halos-core-containers" \
     CONTAINER_DATA_ROOT="$d/data" \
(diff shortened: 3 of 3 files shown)


mairas:
## ce:review synthesis — PR #148

**Scope:** `assets/halos-manage-certs` (+38), `tests/test-halos-manage-certs.sh` (+150). Base `143c5a2` (main).

**Intent:** Make Traefik (`:443`) pick up rotated leaf certs without a stack restart, mirroring the existing cockpit-reload path. SIGHUP explicitly rejected because [traefik/traefik#11624](https://github.com/traefik/traefik/issues/11624) kills the proxy on v3.x; touching the dynamic-config file is the documented file-watcher trigger.

**Review team:** correctness, testing, maintainability, project-standards, agent-native, learnings-researcher (always-on) + security (TLS rotation) + reliability (error paths, gating, ordering).

**Verdict: Ready with fixes.** No P0/P1 correctness or security issues. One P2 test-isolation leak should be fixed before merge; a small batch of P3 doc/test/defense improvements piggy-back cheaply.

---

### P2 — Should fix before merge

|---|------|-------|-------------|------|-------|
| 1 | `tests/test-halos-manage-certs.sh:477-485, 691-699` | Two test cases (`test_cockpit_reload_signal_skipped_when_install_fails`, `test_fingerprint_guard_exits_when_helper_fails`) inline their own env block (because they need to override `HALOS_LIB_CA`) but the new `HALOS_TRAEFIK_DYNAMIC_DIR="$d/traefik-dynamic"` line was added only to `run_script`. With `NEED_LEAF=true` the touch block evaluates against the **real host path** `/etc/halos/traefik-dynamic.d/tls-default.yml`. On a developer machine that is also a HaLOS install, the test attempts to touch the production file (fails harmlessly on root-owned files, but pollutes the test environment with WARN logs and host-state coupling). | testing | 0.90 | safe_auto |
| 2 | `assets/halos-manage-certs:278-282` | `[ -f "${TRAEFIK_TLS_CONFIG}" ]` then plain `touch` is a (mild) TOCTOU: if the file is removed between check and touch, `touch` *creates* it as an empty placeholder that Traefik can't parse. The test's own comment block (lines 605-606) names exactly this hazard. `touch -c` (no-create) closes the window for free. Practical risk is near-zero (no concurrent writer for this path), but the hardening is one character. | correctness + security residual | 0.70 | safe_auto |

### P3 — Fix if straightforward

|---|------|-------|-------------|------|-------|
| 3 | `tests/test-halos-manage-certs.sh:567` | Failure message says "first-boot leaf signing" but the test pre-seeds `tls-default.yml`. True first-boot is the `_skipped_when_dynamic_config_absent` case. Cosmetic wording fix. | correctness | 0.85 | safe_auto |
| 4 | `prestart.sh:78` ↔ `assets/halos-manage-certs:89` | Filename `tls-default.yml` is hardcoded in both files, no back-reference. If a future PR renames it in prestart, halos-manage-certs silently falls into the "config does not exist" branch on every rotation and Traefik never picks up renewals until next container restart — visible only weeks later. One-line comment in prestart.sh closes the loop. | maintainability | 0.78 | safe_auto |
| 5 | `tests/test-halos-manage-certs.sh:80-89` (`_seed_traefik_tls_config`) | Helper writes a full YAML body that visually mimics prestart's output, but the script under test only calls `touch` — it never parses contents. Future reader will pay maintenance cost to keep it in sync with prestart for nothing. Adding a single-line comment ("body is illustrative; only existence matters") is cheaper than trimming. | maintainability | 0.62 | safe_auto |

### Dropped (recorded for completeness, not pursued in this PR)

- **Touch-failure WARN branch untested** (testing P2, conf 0.90). Same gap exists in the cockpit-reload battery; introducing it here would lock in a precedent inconsistency. Acceptable to leave for a future broader "failure-mode coverage" pass.
- **Hardcoded 2024-01-01 mtime sentinel** (testing P3, conf 0.70). Triggers only if host wall clock is < 2024-01-01 (RTC failure, deterministic CI epoch=0, 2023 backport). Probability < 1e-3 in our environment.
- **"Dir exists, file missing" intermediate state not explicitly tested** (testing P3, conf 0.70). Implicitly covered by every test that uses `new_scratch` (which creates the dir without seeding the file) and asserts behaviour in that state.
- **WARN message overstates fallback cost** (reliability P3, conf 0.85). Pessimistic-but-accurate; reword would add words without changing semantics.
- **Fire-and-forget — no readback that Traefik picked up the new cert** (reliability P3, conf 0.90). Inotify queue overflow could silently drop the event; mitigated by 24h timer cadence + 60d renewal threshold (multiple retries) + that any future `halos-core-containers` restart re-bumps the YAML mtime via `prestart.sh`'s unconditional heredoc rewrite. Worth filing as a follow-up issue for hardening but not a release blocker.
- **Race halos_ca_sign_leaf ↔ halos-core-containers restart** (reliability P3, conf 0.70). Eventual consistency holds; not introduced by this PR.
- **TOCTOU symlink-follow on touch as root** (security residual, theoretical). Mitigated by `/etc/halos/traefik-dynamic.d/` being root:root 0755 (only root can place entries; root already wins). `touch -c` fix above also doesn't follow non-existent symlinks.

### Learnings & Past Solutions

`docs/solutions/` has **no existing entry** for Traefik file-watcher reload semantics or the cockpit-reload pattern this PR mirrors — genuinely new institutional knowledge. Adjacent learnings flagged: `2026-05-24-set-e-cmdsubst-blind-spot.md` (relevant to future helpers in this script family), `2026-05-04-image-swap-smoke-test-without-deb-rebuild.md` (the right pattern to validate the Traefik reload on-device without `.deb` churn). Post-merge **`ce:compound`** is a natural fit: this is the cockpit-reload pattern's Traefik twin and deserves a sibling solution doc.

### Agent-native parity

PASS. Three operator surfaces (run service, journalctl, openssl s_client) all reachable by an agent via Bash/SSH. The three new log substrings (`"touching Traefik dynamic config to reload certs"`, `"Skipping Traefik cert reload"`, `"WARNING: touch "`) are stable, distinct, and locked by tests. `mtime` of `tls-default.yml` is a deliberate observability signal (the tests assert against it via `-nt`), so an agent can stat the file and discriminate "touched this run" vs "skipped" without scraping logs. Path is env-overridable for non-default layouts.

### Coverage

- Confidence suppressed: 0 findings below 0.60 dropped (lowest kept was 0.62, intentional — maintainability/2).
- Pre-existing risks: 1 (inotify reliability — predates this PR).
- Untracked files excluded: none.
- All 8 reviewers returned structured results.

---

**Recommended next action:** apply the 5 safe-auto fixes above (3 in tests, 1 in production script via `-c`, 1 cross-repo doc note), re-run `./run test`, push to this PR. After merge: file a follow-up issue for the inotify-readback hardening and run `ce:compound` to document the Traefik-twin reload pattern under `docs/solutions/`.

🤖 Generated by [Claude Code](https://claude.com/claude-code) via `ce:review`.

mairas:
## Fixes applied (review pass 1)

Pushed `7fabbbd` addressing the ce:review findings:

| # | Severity | Finding | Disposition | Commit |
|---|----------|---------|-------------|--------|
| 1 | P2 | Inline env blocks in `test_cockpit_reload_signal_skipped_when_install_fails` and `test_fingerprint_guard_exits_when_helper_fails` missed the `HALOS_TRAEFIK_DYNAMIC_DIR="$d/traefik-dynamic"` override → potential host-state pollution on dev machines | **Fixed** — added override line to both inline blocks | `7fabbbd` (`tests/test-halos-manage-certs.sh:482, 697`) |
| 2 | P3 | `[ -f ]` + plain `touch` would create an empty placeholder on TOCTOU race | **Fixed** — switched to `touch -c` (no-create) with explanatory comment | `7fabbbd` (`assets/halos-manage-certs:284`) |
| 3 | P3 | Test failure message said "first-boot leaf signing" but pre-seeds the config | **Fixed** — reworded to "initial leaf signing with pre-existing dynamic config" | `7fabbbd` (`tests/test-halos-manage-certs.sh:564`) |
| 4 | P3 | Filename `tls-default.yml` hardcoded in two files with no cross-repo back-reference | **Fixed** — added load-bearing comment in `prestart.sh` next to `TLS_CONFIG_FILE` pointing back at halos-manage-certs | `7fabbbd` (`prestart.sh:78`) |
| 5 | P3 | `_seed_traefik_tls_config` writes YAML body the production code never parses | **Fixed** — added inline comment clarifying body is illustrative, not a contract | `7fabbbd` (`tests/test-halos-manage-certs.sh:80`) |
| 6 | P2 | `touch`-failure WARN branch not exercised by any test | **Dropped** — maintainer call; same gap exists in the parallel cockpit-reload suite and won't be filed as a separate issue. |
| 7 | P3 | Hardcoded `2024-01-01` mtime sentinel — brittle if host wall clock predates that | **Dropped** — fires only on RTC failure / 2023 backports / deterministic epoch=0 sandboxes. Probability << 1e-3 in our environment. |
| 8 | P3 | "Dir exists, file missing" intermediate state not explicitly tested | **Dropped** — implicitly covered by every test that uses `new_scratch` (which creates the dir without seeding the file). |
| 9 | P3 | WARN message overstates fallback cost (any halos-core-containers restart re-bumps mtime) | **Dropped** — current wording is pessimistic-but-accurate; reword would add words without changing semantics. |
| 10 | P3 | Fire-and-forget touch — no readback that Traefik picked up the new cert | **Dropped** — maintainer call; mitigations (24h cadence + 60d threshold + prestart.sh unconditional mtime bump on any restart) judged sufficient. Not filing as a follow-up issue. |
| 11 | P3 | Race halos_ca_sign_leaf ↔ halos-core-containers restart | **Dropped** — eventual consistency holds; not introduced by this PR. |

`./run test` green (20/20 in `test-halos-manage-certs.sh`, plus full lib-ca and lib-hostnames suites). CI will re-verify on the new commit.

PR #151, opened 2026-05-26
fix(homarr): preserve originating host on post-login redirect (part of #117)

mairas:
## Summary

Removes `NEXTAUTH_URL=https://${HALOS_DOMAIN}` from the Homarr container in `docker-compose.yml`. With `trustHost: true` in the Homarr fork's NextAuth config and the per-request `reqWithTrustedOrigin` rewriter active in the fork's `[...nextauth]/route.ts`, NextAuth derives URLs from `x-forwarded-host` per request — which is exactly what multi-hostname access requires.

The OIDC issuer remains canonical via `AUTH_OIDC_ISSUER` (single-issuer constraint, R11 in the multi-hostname design).

## Root cause

Two next-auth code paths consume `AUTH_URL`/`NEXTAUTH_URL` and override request-driven URL construction even when `trustHost: true`:

1. `next-auth/lib/env.js` `reqWithEnvURL(req)` — rewrites the request URL's origin to the env URL origin *before any handler logic runs*. Called from `next-auth/index.js` on every GET/POST to `/api/auth/*`, and from `lib/index.js` `handleAuth` for middleware-style auth.
2. `@auth/core/lib/utils/env.js` `createActionURL(...)` — prefers `AUTH_URL ?? NEXTAUTH_URL` over `x-forwarded-host`/`host` headers when constructing internal URLs (session, signin, callback). Used by `getSession` for server-side `auth()` calls in RSCs.

Result: the Homarr fork's `reqWithTrustedOrigin` rewrite (header → request URL) was silently undone inside NextAuth, and every server-generated redirect landed on canonical regardless of the originating host.

## Why this is the minimal fix

- `trustHost: true` was already set in the fork — it controls host *validation*, not URL construction precedence
- The fork already builds OIDC `redirect_uri` and `redirectProxyUrl` from request headers (`createRedirectUri`)
- The only thing forcing canonical was the env override; removing it lets the existing fork machinery work as designed
- `extra_hosts: "${HALOS_DOMAIN}:host-gateway"` stays — Homarr's server-side OIDC discovery still needs to resolve the canonical issuer host
- `AUTH_OIDC_ISSUER` stays pointed at canonical — single-issuer constraint is intentional

## Test plan

- [x] `./run test` — 39/39 shell-library tests pass locally
- [ ] On a test device with a non-canonical hostname added to `/etc/halos/hostnames.conf`, log in via Homarr on the non-canonical host; verify post-login lands back on the originating host (not canonical)
[...]

PR #156, opened 2026-05-27
chore: bump version to 0.4.1

mairas:
## Why

PR #151 changed runtime behavior of the homarr container (removed `NEXTAUTH_URL` env var) but landed without a VERSION bump. CI auto-computed the next revision as `v0.4.0+2`, which is semantically wrong: `+N` is a Debian packaging revision (same upstream source, different packaging), not a vehicle for source changes. Since `v0.4.0+1` was already released as a stable, any subsequent code change requires bumping the upstream version.

## What

`0.4.0 → 0.4.1` via `bumpversion patch`. Updates `VERSION` and `.bumpversion.cfg`.

## Follow-up (not in this PR)

- Delete the stale `v0.4.0+2_pre` pre-release and tag, and the draft `v0.4.0+2` release.
- Check whether the bad `homarr-container_0.4.0+2*.deb` has been published to apt.halos.fi unstable and remove it if so.
- Add a guard in `halos-org/shared-workflows/.github/workflows/build-release.yml` that fails the build when a stable `vX.Y.Z+N` tag for the current `VERSION` already exists AND the diff vs that tag touches non-packaging files.
- Document the rule in this repo's `AGENTS.md` so future PRs surface the requirement in code review.


mairas:
Superseded — parallel session already opened a version-bump PR. Closing this duplicate.

PR #157, opened 2026-05-27
ci: use shared pr-checks workflow (wires in version-bump-check)

mairas:
## Why

This repo's \`.github/workflows/pr.yml\` only ran tests; it never called \`shared-workflows/.github/workflows/version-bump-check.yml\`. That gap let #151 land a runtime-behavior change to \`docker-compose.yml\` without bumping \`VERSION\`, producing the semantically-wrong \`v0.4.0+2\` tag. #154 fixed the VERSION after the fact, but the right time to catch it is at PR-check time.

## What

Replaces the inline test-only \`pr.yml\` with a call to the shared \`pr-checks.yml\` reusable workflow — same pattern halos-marine-containers already uses. That orchestrator runs:

1. Tests via this repo's existing \`.github/actions/run-tests\`
2. Version-bump-check from shared-workflows (the new coverage)
3. Lintian — **skipped here** (\`skip-lintian: true\`) because lintian runs as a backstop in \`build-release.yml\` before publishing, same rationale as halos-marine-containers

## Required companion

halos-org/shared-workflows#25 — stops \`version-bump-check.yml\` from silently excluding root \`docker-compose.yml\`. Without that PR, this wiring catches future cases where any other package-affecting file changes without a VERSION bump, but the specific compose-only regression that motivated all this still slips through. Both PRs should land together.

## Test plan

- [ ] CI green (the new shared workflow runs against this PR — the diff touches only \`.github/\` so version-bump-check is excluded; tests still run)
- [ ] After merge: future PRs touching package-affecting files without a VERSION bump fail PR Checks rather than landing silently

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,11 +5,9 @@ on:
     branches: [main]
 
 jobs:
-  test:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout code
-        uses: actions/checkout@v4
-
-      - name: Run tests
-        uses: ./.github/actions/run-tests
+  checks:
+    uses: halos-org/shared-workflows/.github/workflows/pr-checks.yml@main
+    with:
+      # Skip lintian in PR checks - container-packaging-tools ensures compliance.
+      # Lintian runs as a backstop in build-release.yml before publishing.
+      skip-lintian: true


mairas:
## Self-review

Small wiring change, but worth a careful look since it changes how every PR in this repo gets checked.

### Equivalence vs. previous inline `pr.yml`

Old workflow ran one job: `test` calling `./.github/actions/run-tests` (composite action: setup-python → pytest → shell tests in debian:trixie container → YAML validation → asset-server security tests → cleanup).

New workflow calls shared `pr-checks.yml`, which spawns four jobs against this repo's checkout:

| Job | Behavior here | Notes |
|-----|---------------|-------|
| `tests` | Runs `./.github/actions/run-tests` — identical composite, identical inputs (no `with:` block, action has no required inputs) | Equivalent to old `test` job |
| `version-check` | Conditional on `.github/actions/check-versions/` — **does not exist here**, so step is skipped via the `if` guard | No-op |
| `lintian` | Skipped entirely via `skip-lintian: true` input | Same rationale as halos-marine-containers — lintian runs in `build-release.yml` before publishing |
| `version-bump-check` | Calls `shared-workflows/.github/workflows/version-bump-check.yml@main` | **The new coverage.** Requires halos-org/shared-workflows#25 to catch root `docker-compose.yml` changes; without it, catches every other package-affecting file but not the specific regression that motivated this work |

### Branch protection / required checks

Checked `repos/halos-org/halos-core-containers/branches/main/protection` — `required_status_checks.contexts` and `.checks` are both empty. Renaming the `test` job to `checks / tests` does not break branch protection. ✓

### Things I did not change but considered

- **Pin to `@main` vs SHA.** halos-marine-containers also pins `@main`. Following convention; deliberate choice to let shared-workflow bug fixes propagate automatically. The security purist's "pin to SHA" is theoretically safer but inconsistent with org practice.
- **Concurrency control.** Old `pr.yml` had no `concurrency:` block; new one doesn't either. PR pushes will queue rather than cancel in-progress runs. Not in scope but a small future improvement.
- **Per-job CI latency.** Splitting one job into 3 jobs (tests, version-check no-op, version-bump-check; lintian skipped) adds parallel-startup overhead but the wall-clock-critical job (tests, with Docker) is unchanged. Acceptable.

### Sequencing with halos-org/shared-workflows#25

These two PRs are complementary:
- This PR alone catches future regressions where a non-compose package file (e.g., `prestart.sh`, `assets/*`, `debian/rules`) is changed without bumping `VERSION` after a stable release.
- shared-workflows#25 alone is wasted work in this repo because nothing here calls the workflow yet.
- Both together close the exact regression class that produced `v0.4.0+2` from #151.

No hard merge-ordering requirement (this PR's diff only touches `.github/`, so it's auto-excluded from version-bump-check itself), but they should land roughly together to avoid a window where the wiring is in but the fix isn't.

### Verdict

**Ready.** Minimal, structurally equivalent to halos-marine-containers' working setup, no branch-protection breakage. Land with shared-workflows#25.

— self-review (single inline pass; not full ce:review given the 6-line scope)

Issue #18, opened 2025-12-19
feat(sso): Create Traefik container package


mairas:
## Description

Create `apps/traefik/` with Docker provider configuration as the foundation for the SSO infrastructure.

## Implementation Requirements

**MANDATORY**: Follow `docs/IMPLEMENTATION_CHECKLIST.md` during implementation.

This includes:
- Exploration phase WITHOUT coding
- Planning with "think hard"
- Test-Driven Development (write tests first)
- Verification with subagents
- References to SSO_SPEC.md and SSO_ARCHITECTURE.md

## Implementation Details

Create the following files:
- `docker-compose.yml` with traefik:v3.6 image
- `traefik.yml` static config (Docker provider, entrypoints web/websecure)
- Create `halos-proxy-network` bridge network (owned by this container)
- `metadata.yaml` with `field::core` tag
- `config.yml` for user-configurable options (if any)
- `dynamic/authelia.yml` - ForwardAuth middleware definition (placeholder for Issue #19)

Key configuration:
- Docker provider watching `/var/run/docker.sock`
- `exposedByDefault: false`
- Network set to `halos-proxy-network`
- Entrypoints: web (80), websecure (443)

## Test Scenarios

- Container starts successfully
- Network `halos-proxy-network` is created
- Traefik dashboard accessible on port 8080 (internal only)
- Can route to a test container with Traefik labels

## Acceptance Criteria

- [ ] Package builds with container-packaging-tools
- [ ] Container starts and creates shared network
- [ ] Can route to a test container with Traefik labels
- [ ] Matches SSO_SPEC.md requirements
- [ ] Follows SSO_ARCHITECTURE.md patterns

## Dependencies

None - this is the foundation package.

## References

- SSO_SPEC.md
- SSO_ARCHITECTURE.md

PR #161, opened 2026-05-28
fix(routing): strip HSTS at per-app entrypoints

mairas:
## Why

Signal K Server's upstream defaults ([helmet](https://github.com/helmetjs/helmet)) emit `Strict-Transport-Security: max-age=31536000; includeSubDomains` on every response. HSTS state is keyed by **host** in browsers, so a single visit to any per-app port — for example the 307 from `/signalk-server/` to `https://<device>.local:4430/` — pins the entire host into HSTS-only mode for a year, with `includeSubDomains` propagating it.

Under HSTS, every TLS error on the host becomes **non-overridable**: no "Proceed to <device>.local (unsafe)" link, only the `thisisunsafe` keystroke. That bypass is itself rate-limited by Chromium (~5–15 minutes, fixed-duration, not refreshed by reloads), so the user-perceived symptom is "the cert alert keeps reappearing every ~15 minutes regardless of how many times I bypass it."

Confirmed by live diagnostic on a test device: the server-side cert bytes were byte-identical for the entire 16-minute window over which the user measured the cadence. The recurrence was the Chromium HSTS-bypass timer recycling, not anything the device did to the cert.

## What

- New `assets/traefik/dynamic/strip-hsts.yml` defines a `strip-hsts@file` middleware (`headers.customResponseHeaders.Strict-Transport-Security: ""`).
- `assets/configure-container-routing` always prepends `strip-hsts@file` to the middleware chain of every `<app>-secure` router, regardless of `auth_mode`. So forward-auth apps get `strip-hsts@file,authelia@file` (or `strip-hsts@file,authelia-<app>@file` for custom-headers apps), and no-auth apps (e.g., signalk-server) get just `strip-hsts@file`.

Scope intentionally limited to per-app entrypoints (`:4430-:4450`). The `:443` websecure routers (Homarr / Cockpit redirect) don't currently emit HSTS upstream; adding defense-in-depth there would be a separate follow-up if needed.

Also fixes a pre-existing bug in the dynamic config install loop in `prestart.sh`: the `if [ ! -f ]` guard prevented updated package-shipped configs from overwriting stale versions on upgrade. The broader prestart.sh audit is tracked in halos-org/halos#124.

## How verified

Hot-fixed on the test device:

- **Before:** `curl -ksI https://<device>.local:4430/signalk` returned `strict-transport-security: max-age=31536000; includeSubDomains` in the 200 response.
- **After** patching the on-device script and restarting `marine-signalk-server-container.service`: same request returns no `Strict-Transport-Security` header. Confirmed on both the 302 redirect at `/` and the 200 JSON at `/signalk`.

## Caveats

Browsers that already cached HSTS state for the device before this change still need to clear it once via `chrome://net-internals/#hsts` → "Delete domain security policies" → enter hostname → delete. After clearing, the lockout will not recur with this fix in place. The sibling docs PR updates the troubleshooting guide to explain this.

## Links

- Sibling docs PR: halos-org/docs#18
- Original HSTS docs from earlier today: halos-org/docs#17
- prestart.sh audit: halos-org/halos#124


mairas:
## ce:review synthesis

Multi-agent review (correctness, testing, maintainability, project-standards, agent-native, security, reliability + learnings researcher). Scope: 37 additions / 4 deletions across `assets/configure-container-routing` and `assets/traefik/dynamic/strip-hsts.yml`, base `e8592b5`.

**Verdict — Ready with fixes.**

The central fix is correct and addresses a real, well-diagnosed user-blocking bug. Empty-string `customResponseHeaders` is the documented Traefik strip mechanism. Chain ordering `strip-hsts,authelia` is correct: Traefik runs response middlewares in reverse, so `strip-hsts` touches the response last and strips HSTS from both backend and auth-denied responses.

### P2 — Moderate

1. **Boot/upgrade race on `auth_mode != forward_auth` apps.** Per-app systemd units (template at `container-packaging-tools/.../systemd/service.j2:3`) have only `After=docker.service` — no ordering against `halos-core-containers.service`. Cold-boot window is small because Traefik itself is started by halos-core-containers after its prestart installs `strip-hsts.yml`. The more plausible window is in-place upgrade: between dpkg unpacking the new `configure-container-routing` and the postinst service restart, any per-app autoheal/manual restart could write labels referencing a not-yet-installed `strip-hsts.yml`. Until now this race only mattered for `authelia@file` (forward_auth apps); it's empirically not load-bearing, but the new failure mode targets `auth_mode: none` apps (signalk-server, opencpn, etc.) that previously emitted no `@file` middleware reference. **Suggest filing as a follow-up issue** — add `After=halos-core-containers.service` to the per-app systemd template, or install dynamic configs from postinst rather than prestart. Not a blocker.
2. **Three new middleware-chain branches with zero automated coverage.** `generate_override()` now has 3 paths: forward_auth+custom-headers, forward_auth+default, other. PR verification (`curl -ksI`) only exercised SignalK's path. The seam is test-ready — `ROUTING_DIR` / `OUTPUT_DIR` are env-overridable and `tests/test-lib-hostnames.sh` is a working precedent. ~80-120 LOC of fixtures + grep assertions would lock the invariant.
3. **PR description has no `Closes #<n>` / `Refs #<n>`.** AGENTS.md mandates issue linking. Sibling docs PRs (#17, #18 in halos-org/docs) don't satisfy the rule. Either file a halos-core-containers issue and add the reference, or explicitly note this is a fly-by fix with no tracking issue.

### P3 — Low

4. **`:443` scope gap — empirically verified clear today.** Live check on halosdev.local:
   ```
   /        (Homarr)     → no HSTS
   /sso/    (Authelia)   → no HSTS
   /ca/     (ca-download)→ no HSTS
   /cockpit/(redirect)   → no HSTS
   :9090    (Cockpit dir)→ no HSTS  (outside Traefik — unreachable anyway)
   ```
   PR's "they don't currently emit HSTS upstream" assumption holds as of this commit. Future-proofing option: declare `strip-hsts@file` as a default middleware on the `websecure` entrypoint in `traefik.yml` so a future Homarr/Cockpit version that flips on helmet defaults can't silently regress this. Cheap defense-in-depth; not required.
5. **Test-device hostname `halosdev.local` in shipped asset comments.** ✅ Addressed in `10a1d3a` — replaced with `<device>.local`.
6. **`strip-hsts@file` literal duplicated across three arms.** ✅ Addressed in `10a1d3a` — hoisted to a `mw_chain` base that downstream arms append to.
7. **User-visible upgrade impact under "Caveats" rather than "Release notes".** AGENTS.md's literal trigger list (canonical-hostname / hostnames.conf / OIDC) doesn't include HSTS, so the rule doesn't strictly fire. Spirit applies — relabeling makes it changelog-greppable. Trivial.
8. **Auth-denied response strip not yet verified.** Smoke test: hit a forward_auth app port unauthenticated and confirm Authelia's 302/401 lacks `Strict-Transport-Security`. Expected to pass given Traefik's response-unwind ordering.
9. **`prestart.sh:164` `if [ ! -f ]` guard.** Existing pattern across all dynamic configs — future revisions of `strip-hsts.yml` won't reach upgraded systems without manual cleanup. Worth a tracked follow-up if a future PR refines the rule.
10. **`strip-hsts.yml` rejected-alternative framing.** Lines 11-16 explain "without this middleware, users see `thisisunsafe`" — per project convention, rejected-alternative rationale belongs in the commit message / PR description, not inline. First paragraph (host-keyed cache mechanics) earns its keep. Judgment call.

### Verification done in-review

- **AvNav HSTS check:** raw backend on `:8083` carries no HSTS; through Traefik on `:4431` the 302 was Authelia rejecting the unauthenticated request, not AvNav. To confirm whether AvNav itself emits HSTS, a smoke test needs an authenticated session AND strip-hsts temporarily out of the chain — but for the question "does the fix work in production", the unconditional per-app strip makes the emitter inventory moot.
- **Empty-string `customResponseHeaders` semantics:** Traefik 3.7 strips the header (vs setting it to empty), confirmed by the on-device curl in the PR description.

### Suggested follow-ups (not blocking)

- File an issue for finding 1 (per-app systemd ordering vs. core-containers prestart).
- File an issue for finding 4 (promote `strip-hsts` to a default `websecure` middleware as defense-in-depth).
- Optionally bundle findings 2 (tests), 7 (Release notes heading), 9 (install-loop guard) into a hygiene PR.

PR #162, opened 2026-05-28
feat(routing): proxy SK Server icons on :443

mairas:
## Why

The Homarr dashboard renders Signal K plugin card icons via path-only URLs like `/signalk-server/@signalk/app-dock/app-icon.svg`. By default these match the auto-generated `redirect-signalk-server-https@file` router (priority 100, `PathPrefix(/signalk-server/)`) and get 307-redirected to `https://<host>:4430/@signalk/app-dock/app-icon.svg`.

That redirect crosses a port boundary. Chromium-family browsers key cert exceptions by `(host, port, leaf-fingerprint)`, so a user without the device CA installed sees broken icons until they manually navigate to `:4430` and click through the cert warning — once per (hostname, app port) tuple they ever visit. Safari requires this even with the prior self-signed cert default (regression flagged in test session). Multi-hostname access (`<host>.local` + DHCP-domain FQDN) multiplies the count.

Static images aren't navigation; they don't need origin isolation from the dashboard. Same-origin serving means the dashboard's already-granted `:443` cert exception covers them.

## What

New `assets/traefik/dynamic/signalk-server-icons.yml` defining:

- Router `signalk-server-icons-on-443` matching `PathRegexp(^/signalk-server/.+\.(svg|png|ico|jpg|webp)$)` at priority 200, beating the redirect router (priority 100).
- StripPrefix middleware to map `/signalk-server/<asset>` → `/<asset>` for the backend.
- Service pointing at the SK Server backend on `http://host.docker.internal:3000` — same target the per-app `:4430` router uses.
- `strip-hsts@file` middleware (from #161) wired in so the new router doesn't reintroduce the HSTS leak.

Scope intentionally narrow: extensions cover icons/images only. HTML, JS, and CSS are excluded because a SK plugin SPA loaded under a `/signalk-server/...` prefix on the dashboard origin would silently break (webapps assume `/` as their root and embed absolute paths). Plugin navigation continues through the existing 307→`:4430` path so each app keeps its `/`-rooted origin.

## How verified

Live device:

- **Icon URL via :443:** `curl -ks -D - -o /dev/null https://halosdev.local/signalk-server/@signalk/app-dock/app-icon.svg` → `HTTP/2 200`, `content-type: image/svg+xml`, no `Strict-Transport-Security`.
- **HTML navigation regression:** `/signalk-server/admin/` → `HTTP/2 302` to `:4430`, unchanged.
- **SPA bundle regression:** `/signalk-server/admin/main.js` → `HTTP/2 302` to `:4430`, unchanged.
- **Other whitelisted extensions** (png, ico, jpg, webp): reach the backend (404 from SK Server's HTML 404 page, proving the proxy works for them too).
- **Traefik API confirms** `signalk-server-icons-on-443@file` is enabled with the expected rule and priority 200.

## Dependencies

- Built on top of #161 (HSTS strip middleware) — the new router references `strip-hsts@file`. PR base is set to that branch so the diff reads cleanly. Will retarget to `main` once #161 merges, or rebase if a different order is preferred.

## Future work

This pattern is signalk-server-specific. If other per-app backends grow similar needs (Homarr fetching status badges, etc.), generalizing through `configure-container-routing` with a `static_assets` opt-in in `routing.d/<app>.yml` would be the clean path. Out of scope here.

### DIFF
--- a/assets/traefik/dynamic/signalk-server-icons.yml
+++ b/assets/traefik/dynamic/signalk-server-icons.yml
@@ -1,0 +1,37 @@
+# Serve Signal K Server static icons / images same-origin on :443 instead
+# of redirecting them to the per-app port :4430, so the dashboard's cert
+# exception covers them and the user avoids a per-port click-through.
+# Full rationale (browser cert-exception model, scope trade-offs):
+# docs/CERTS.md → "Same-origin asset serving on :443".
+http:
+  routers:
+    signalk-server-icons-on-443:
+      # Images only (case-insensitive); navigation (HTML/JS/CSS) intentionally
+      # falls through so plugin SPAs keep a `/`-rooted origin. See docs/CERTS.md.
+      rule: 'PathRegexp(`(?i)^/signalk-server/.+\.(svg|png|ico|jpg|jpeg|gif|avif|webp)$`)'
+      entrypoints:
+        - websecure
+      tls: {}
+      # Must beat the auto-generated redirect-signalk-server-https router (100).
+      priority: 200
+      middlewares:
+        - signalk-server-icons-strip-prefix
+        - strip-hsts@file  # don't re-introduce the HSTS leak the per-app routers strip
+      service: signalk-server-icons-backend
+
+  middlewares:
+    signalk-server-icons-strip-prefix:
+      stripPrefix:
+        prefixes:
+          - "/signalk-server"
+
+  services:
+    signalk-server-icons-backend:
+      loadBalancer:
+        servers:
+          # Backend host:port must track the Signal K Server app definition
+          # (halos-marine-containers/apps/signalk-server/metadata.yaml:
+          # `routing.host_port`, currently 3000) — the same target the
+          # auto-generated per-app `:4430` router derives from routing.d.
+          # This file hardcodes it; if the port ever changes, update both.
+          - url: "http://host.docker.internal:3000"
--- a/docs/CERTS.md
+++ b/docs/CERTS.md
@@ -278,3 +278,44 @@ sudo systemctl start halos-manage-certs.service
 The auto-CA at `/var/lib/container-apps/halos-core-containers/data/halos-core-containers/certs/ca/ca.crt`
 is preserved across mode switches; reverting just re-points
 `serving-ca.crt` back at it and re-signs the leaf with the auto-CA.
+
+## Same-origin asset serving on :443 (cert-exception friction)
+
+Until a device's CA is installed, every TLS surface the user touches needs a
+manual cert exception. Browsers key those exceptions by **`(host, port)`**
+(Chromium also folds in the leaf fingerprint; Safari is per-`(host, port)`
+even against the prior self-signed default). So each distinct port a user is
+sent to costs one more click-through, and multi-hostname access
+(`<host>.local` *and* the DHCP-domain FQDN) multiplies the count again.
+
+This bites the Homarr dashboard. Each Signal K plugin card renders an icon
+via a path-only URL such as
+`/signalk-server/@signalk/app-dock/app-icon.svg`. By default that path
+matches the auto-generated `redirect-signalk-server-https@file` router
+(priority 100, `PathPrefix(/signalk-server/)`) and gets 307-redirected to
+`https://<host>:4430/...` — the per-app port. Crossing from `:443` to
+`:4430` is a new `(host, port)` pair, so the dashboard renders with broken
+icons until the user manually visits `:4430` and clicks through the warning,
+once per app port they ever encounter.
+
+**Fix:** static images are not navigation and don't need origin isolation
+from the dashboard. `assets/traefik/dynamic/signalk-server-icons.yml`
+defines a higher-priority (200) router that catches image extensions under
+`/signalk-server/`, strips the prefix, and proxies them to the Signal K
+backend on `:443` directly. Served same-origin with the dashboard, they're
+covered by the cert exception the user already granted for the dashboard —
+no extra click-through.
+
+**Why only images.** The scope is deliberately narrow (svg, png, ico, jpg,
+jpeg, gif, avif, webp, case-insensitive). HTML, JS, and CSS are excluded: a
+Signal K plugin SPA loaded under a `/signalk-server/...` prefix on the
+dashboard origin would silently break, because webapps assume `/` as their
+root and embed absolute paths. Navigation to a plugin therefore stays on the
+existing redirect-to-`:4430` path so each plugin keeps a clean `/`-rooted
+origin — accepting the one cert exception per app port for *navigation*
+while eliminating it for *icons*. (The proper long-term fix is to install
+the device CA; see the CA download endpoint above. This router only removes
+the friction for the icon case in the meantime.)
+
+The router also wires `strip-hsts@file` so it does not re-introduce the HSTS
+leak the per-app routers were patched to strip.
--- a/tests/test-traefik-signalk-icons-regex.sh
+++ b/tests/test-traefik-signalk-icons-regex.sh
@@ -1,0 +1,169 @@
+#!/usr/bin/env bash
+# Tests for the PathRegexp rule in
+# assets/traefik/dynamic/signalk-server-icons.yml
+#
+# Run from repo root:
+#   bash tests/test-traefik-signalk-icons-regex.sh
+#
+# The rule decides which /signalk-server/* requests are served same-origin
+# on :443 (static images) versus left to the priority-100 redirect router
+# (navigation: HTML/JS/CSS). A regression here silently reproduces the
+# cross-port cert-exception friction the router exists to remove, with no
+# on-device signal. This test extracts the regex from the config itself so
+# it stays coupled to the shipped rule, and evaluates it the way Traefik
+# does: against the URL path only (query/fragment already stripped).
+#
+# Pure bash (POSIX ERE via [[ =~ ]]) to match the rest of the suite and
+# avoid any external-tool dependency in CI. The rule's `(?i)` flag (which
+# ERE does not interpret) is asserted separately, and case-insensitive
+# matching is emulated by lowercasing the path before matching against the
+# all-lowercase extension alternation — equivalent for this pattern.
+
+set -u
+set -o pipefail
+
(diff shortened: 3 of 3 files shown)


mairas:
## Code Review — PR #162 (proxy SK Server icons on :443)

**Scope:** 1 commit, 1 new file (`assets/traefik/dynamic/signalk-server-icons.yml`, 68 lines / ~25 config).
**Intent:** Serve Signal K static image assets on the `:443` entrypoint (same-origin with the dashboard) so users without the device CA don't hit per-app-port cert exceptions for icons. Navigation (HTML/JS/CSS) intentionally still redirects to `:4430`.
**Mode:** Report-only (comment, no branch mutation).
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + security & adversarial (routing/auth boundary, file ≥50 lines). 6 personas, independent Sonnet contexts.

**Verdict: ✅ Ready to merge.** No P0/P1 blockers. The findings below are robustness/maintainability improvements — none block the merge. The one I'd most encourage folding in is the regex robustness (case-insensitive + `.jpeg`), because it's trivial and directly protects the PR's own goal.

One reviewer raised an **auth-bypass P2 that I'm dismissing after verification** — see "Investigated & dismissed" below. I want that transparent rather than buried.

---

### P2 — Moderate

| # | Location | Issue | Reviewer(s) | Conf | Route |
|---|----------|-------|-------------|------|-------|
| 1 | `signalk-server-icons.yml:48` | **Regex is case-sensitive and the extension list is narrow.** `^/signalk-server/.+\.(svg\|png\|ico\|jpg\|webp)$` won't match `.SVG`/`.PNG` (uppercase), nor `.jpeg`/`.gif`/`.avif`. Any such icon falls through to the priority-100 redirect router → 307 to `:4430` → reproduces exactly the cert-friction this PR removes. Real-world likelihood is low (npm/@signalk plugin icons are lowercase `.svg`/`.png` by convention), but the fix is trivial and risk-free. | correctness, adversarial, testing | 0.88 | gated |
| 2 | `signalk-server-icons.yml:68` | **Hardcoded `host.docker.internal:3000` + `/signalk-server` prefix duplicate values that `configure-container-routing` derives from `routing.d`/`metadata.yaml`.** Two-site drift risk if the SK port or app id ever changes. The PR's own "Future work" acknowledges this; the clean fix is the deferred `static_assets` generalization. | maintainability | 0.82 | advisory |
| 3 | CI (`.github/actions/run-tests`) | **`assets/traefik/dynamic/*.yml` is excluded from CI YAML-syntax validation** (only `apps/*/metadata.yaml` is checked). Pre-existing gap, widened by adding another unvalidated file — a YAML typo here fails silently at Traefik startup on-device, not at PR time. | testing | 0.85 | pre-existing → file-issue |

### P3 — Low

| # | Location | Issue | Reviewer(s) | Conf | Route |
|---|----------|-------|-------------|------|-------|
| 4 | `:443` origin | **Dashboard/Authelia session cookies are now sent to the SK backend** (same-origin side-effect of moving icons to `:443`; previously `:4430` was a separate origin). Low impact: localhost-scoped, SK ignores unknown cookies. Optionally strip `Cookie` on the request. | security | 0.65 | advisory |
| 5 | service def | **If SK is down, icon paths return 502 on `:443`.** Adversarial framed this as a regression, but pre-PR the same request 307'd to a broken-cert `:4430` — both states show a broken tile, so the *outcome* is unchanged. Noted, not actionable. | adversarial | 0.70→adj | advisory |
| 6 | `:47` & `:65` | Router and service share the identical name `signalk-server-icons-on-443`. Valid in Traefik (separate namespaces); a `-backend` suffix on the service would aid log/dashboard disambiguation. | maintainability | 0.62 | optional |
| 7 | tests | No repeatable guard that the regex matches icons but **not** HTML/JS. A tiny `re.match`/`grep -P` fixture test (should-match vs should-not-match) would plug into `./run test` cheaply. | testing | 0.65 | advisory |

---

- **`prestart.sh` install-guard `if [ ! -f "$dest" ]`** (flagged P1 by correctness + adversarial, conf 1.0): on this branch the dynamic-config install loop only copies a file when absent, so a *future* edit to a shipped config would never reach existing devices. **However, this is moot for this PR:** `origin/main` already removed the guard (in #161's vicinity), and this PR doesn't touch `prestart.sh`, so a three-way merge into `main` keeps main's guardless version. **Recommendation:** rebase this branch onto current `main` — it's behind, and rebasing both picks up the guard removal and lets the diff retarget cleanly now that #161 is merged. No separate issue needed (main is already fixed).
- **CI YAML validation gap** (finding #3): genuinely pre-existing and worth a follow-up issue. Happy to file it.

### Investigated & dismissed (transparency)

- **"Auth bypass: icons reachable on `:443` without forwardAuth" (security, P2, 0.72) — dismissed.** The premise was that the per-app `:4430` router applies `authelia@file`. It does not for this app: `halos-marine-containers/apps/signalk-server/metadata.yaml:46-47` sets `routing.auth.mode: none`. Signal K enforces its own authz internally via OIDC (`SIGNALK_OIDC_*`, default permission `readonly`). Both the existing redirect→`:4430` path and the new `:443` path reach the same backend (`:3000`) with no Traefik-level Authelia, and SK applies identical internal authz on both. **No net-new bypass.** The adversarial reviewer correctly identified `auth_mode: none`; the security reviewer's contradicting claim was wrong for this app.
- **Double-extension path trick (e.g. `…/data.json.svg`) (security, P3) — negligible.** Even if such a path reaches a plugin route, SK applies the same internal authz it would on `:4430`; no privilege change vs the status quo.

### Coverage

- Suppressed: 0 findings below 0.60 confidence.
- `project-standards`: clean. VERSION (`0.4.1`) matches latest stable tag base (`v0.4.1+3`) → no bump required (CI auto-increments `+N`); no `debian/changelog` edit; no hardcoded hostnames (`host.docker.internal` + path only).
- Residual risks worth a one-line comment in the file: priority `200` is chosen to beat the generated redirect router (priority `100`) — there's no assertion guarding against a future change that raises the redirect priority; and `strip-hsts@file` is a cross-file dependency with no compile-time check (verified present and correct).
- On-device verification in the PR description is thorough (200 on icons, 302 for HTML/JS, Traefik API confirms priority 200). Good.

🤖 Multi-agent review (6 personas) · synthesized + adversarially cross-checked

mairas:
## Review fixes applied

Force-pushed onto the feature commit (`6d10c41` → `eb0dd73`). What changed in response to the review:

| Finding | Action |
|---|---|
| **P2** Regex case-sensitive + narrow extension list | Rule is now `(?i)^/signalk-server/.+\.(svg\|png\|ico\|jpg\|jpeg\|gif\|avif\|webp)$` — case-insensitive, and adds `jpeg`/`gif`/`avif`. Uppercase or `.jpeg` icons no longer fall back through the cert-friction redirect. |
| **P3** No repeatable regex regression guard | Added `tests/test-traefik-signalk-icons-regex.sh` (9 cases). It extracts the rule **from the config** and asserts every image extension incl. uppercase matches, while root/HTML/JS/CSS and wrong prefixes fall through to `:4430`. Plugs into `./run test`; full suite 48/48 green. |
| **P2** Hardcoded backend port duplicates generated routing | Added a comment at the service `url` naming the source of truth (`signalk-server/metadata.yaml` `routing.host_port`) and the drift contract. Full DRY fix is the deferred `static_assets` generalization. |
| **P3** Router and service share a name | Service renamed `signalk-server-icons-on-443` → `signalk-server-icons-backend`. |
| **P2** CI doesn't validate `assets/traefik/dynamic/*.yml` (pre-existing) | Filed halos-org/halos-core-containers#163 (out of scope here). |

### Deliberately not changed
- **Cookie forwarding to backend (P3):** stripping the `Cookie` header would break SK's own session-based authz for any non-public asset path the regex matches (SK uses `auth_mode: none` + internal OIDC). The leak is localhost-scoped and SK ignores unrelated cookies — not worth the breakage risk. Left as documented residual risk.
- **502-when-SK-down (P3):** the pre-PR path also produced a broken icon (307 → cert error); outcome unchanged.

### Note
The original "What"/"How verified" in the description predates these fixes — the shipped rule is now the broader case-insensitive one above. Live-device curls for `.svg` still hold; the added extensions/case are covered by the new unit test rather than re-verified on-device.

The branch is still behind `main`; recommend a rebase (picks up the `prestart.sh` install-guard removal) before merge.

mairas:
## Rebased onto main + documentation relocated

Force-pushed `590a4f2` → `11c792c`. Two changes since the review-fixes push:

### 1. Rebased onto current `main`
The branch was behind `main`. Rebased cleanly (single commit, no conflicts). Consequences:
- **Install-guard finding resolved structurally.** The branch now inherits main's guardless `prestart.sh` dynamic-config install loop, so a future edit to this file deploys on upgrade. No code change needed here — it came with the rebase.
- `strip-hsts@file` (from #161) is present on the base, so the middleware reference resolves.
- `VERSION` is `0.4.2` (from main) — already ahead of the latest stable tag base, so no bump is required and `version-bump-check` passes.

### 2. Design narrative moved out of the YAML → `docs/CERTS.md`
The config file carried a 51-line rationale essay — 3.6× the densest sibling dynamic-config (`strip-hsts.yml` at 14) and 68% of the file. The same rationale was also triplicated across the commit body and PR description.

- The browser cert-exception model, Safari behavior, multi-hostname multiplication, and the images-only scope trade-off now live in a new **"Same-origin asset serving on :443 (cert-exception friction)"** section in `docs/CERTS.md`, next to the related "Chicken-and-egg" trust discussion.
- `signalk-server-icons.yml` is trimmed to **14 comment lines** — a concise purpose header + a pointer to the doc, plus terse inline anchors at the load-bearing lines (the rule's images-only/case-insensitive intent, why `priority: 200`, why `strip-hsts` is wired, and the backend host:port drift note).

### Scope now
3 files: `assets/traefik/dynamic/signalk-server-icons.yml` (config), `docs/CERTS.md` (rationale), `tests/test-traefik-signalk-icons-regex.sh` (regex guard).

### Verification
- Full suite green: 20 + 76 + 39 + 10. The regex test still extracts the live rule from the config after the comment trim.
- YAML parses.
- CI: `tests` ✓, `version-bump-check` ✓, `version-check` ✓, `lintian` skipped.

PR is rebased on `main` and ready to merge.

Issue #2, opened 2025-12-13
feat: add local icons volume mount to Homarr container


mairas:
## Summary

Add a volume mount to the Homarr container to serve local icons from `/usr/share/pixmaps`. This enables offline icon serving for the Homarr dashboard.

## Problem

Homarr currently fetches icons from external URLs (CDNs, project websites). When the device is offline (common for marine installations), icons fail to load.

## Solution

Homarr natively supports local icons via the `/app/public/icons` directory. By mounting the system's pixmaps directory, all installed container package icons become available to Homarr.

## Implementation

**File:** `apps/homarr/docker-compose.yml`

Change from:
```yaml
volumes:
  - /var/run/docker.sock:/var/run/docker.sock:ro
  - ${CONTAINER_DATA_ROOT}/data:/appdata
```

To:
```yaml
volumes:
  - /var/run/docker.sock:/var/run/docker.sock:ro
  - ${CONTAINER_DATA_ROOT}/data:/appdata
  - /usr/share/pixmaps:/app/public/icons:ro
```

## How It Works

1. Container packages install icons to `/usr/share/pixmaps/<package-name>.png`
2. Homarr serves files from `/app/public/icons/` at the URL path `/icons/`
3. homarr-container-adapter transforms file paths to URL paths:
   - `/usr/share/pixmaps/app.png` → `/icons/app.png`

## Related Issues

- hatlabs/halos-distro#48 - Epic: Local icon hosting for offline Homarr dashboard
- hatlabs/homarr-container-adapter#4 - Transform file paths to URLs
- hatlabs/halos-marine-containers#51 - Add icons to marine apps
- hatlabs/halos-imported-containers#39 - Download icons for CasaOS apps

## Workflow Reference

Follow @docs/DEVELOPMENT_WORKFLOW.md

### □ Phase 1: EXPLORE (No Coding!)
- [ ] Read this issue completely
- [ ] Read current `apps/homarr/docker-compose.yml`
- [ ] Verify Homarr documentation on local icons

### □ Phase 2: PLAN
- [ ] Confirm mount is read-only (`:ro`) for security
- [ ] Consider if any other volumes needed

### □ Phase 2.5: STOP
- [ ] Let your human partner review plan
- [ ] Get explicit approval to proceed

### □ Phase 3: IMPLEMENT
- [ ] Add volume mount to docker-compose.yml
- [ ] Bump package version in metadata.yaml

### □ Phase 4: VERIFY
[...]

Issue #160, opened 2026-05-28
Default cert mode = self-signed; auto-CA and custom-CA become opt-in


mairas:
> ⚠️ **Read this before re-diagnosing.** The recurring "~30-minute" cert warning is **RESOLVED** and was **NOT** leaf rotation / sentinel / `${fqdn}` drift. It was **HSTS (Bug B), fixed by #161**. Server cert bytes were byte-identical over the measured window; `halos-manage-certs` did not run. Do not re-attribute the recurring cadence to "Bug A" sentinel drift — that bug causes **one** re-sign per boot, not a recurring cadence. Full evidence in the comments dated 2026-05-28. The original "Problem" framing below predates the diagnosis and is kept only for history.

## Diagnostic outcome (RESOLVED 2026-05-28)

The user-observed "cert warning every ~30 minutes, even though nothing changes on the server" was **two distinct client-side causes**, neither of which is leaf rotation:

- **Bug B — HSTS (the dominant, recurring symptom). FIXED by #161.** Signal K Server's helmet defaults emitted `Strict-Transport-Security`. One visit pins the host into HSTS-only mode, which removes the "Proceed" interstitial link and forces `thisisunsafe` — and Chromium rate-limits that bypass to a fixed **5–15 minute** window, not refreshed by reloads. That recycling *is* the "~30 min" cadence. Confirmed by live diagnostic: server cert bytes byte-identical for the full 16-minute window; `halos-manage-certs` did not run. #162 additionally removed the per-port icon click-through. (See comment 2026-05-28T13:31Z.)
- **Stale / untrusted client CA (the earlier finding).** The device had regenerated its CA ~3 days after the workstation installed the old one, and the installed CA had **no macOS trust settings** (present ≠ trusted). Pure client trust-store state; zero server change.
[...]

PR #165, opened 2026-05-29
fix(certs): sticky domain resolution to stop one-re-sign-per-boot

### LINKED ISSUE #164: fix(certs): sticky domain resolution to stop one-re-sign-per-boot (Bug A)
## Problem

`halos-manage-certs` re-signs the TLS leaf exactly once per boot on devices whose DHCP domain isn't available at early boot.

`_halos_resolve_domain` (in `lib-hostnames.sh`) is stateless. The cert service runs `Before=halos-core-containers.service`, before NetworkManager settles, so `hostname -d`/`nmcli` return no domain → the `${fqdn}` line in `hostnames.conf` soft-drops → the leaf is signed with SANs `[<host>.local]` only and the sentinel is stored for that set. About 15 min later the renewal timer fires, DHCP domain is now resolvable, `${fqdn}` expands to `<host>.hal`, `halos_hostnames_hash` changes, the sentinel mismatches, and the leaf is re-signed.

Confirmed on halosdev.local: boot at 09:20 signed `[halosdev.local]`; timer at 10:22 logged "Hostname list or CA changed, re-signing" and expanded to `[halosdev.hal, halosdev.local]`; stable for 12h after (one re-sign per boot, not a recurring cadence).

The `.hal` SAN is **wanted** — it's real unicast DNS, more robust than mDNS `.local`, and resolves to a different IP. The bug is the *instability of when it's included*, not the SAN itself.

## Fix — sticky domain at the resolver layer

Persist the last-known-good resolved domain and fall back to it when live resolution returns empty:

- After resolving, if the value is non-empty and passes `_halos_domain_safe`, write it atomically to a state file.
- When live resolution is empty, read the state file and use that instead of dropping the FQDN.
- A genuinely *different* live domain overwrites the stored value → legitimate re-sign. A *transient empty* keeps the prior FQDN → no re-sign.

Fixing this in the resolver (not in cert code) stabilizes the hostname set across boots for every consumer — cert SANs, Authelia per-hostname cookies, OIDC redirect URIs, and the routing canonical.

### Behavior after the fix

First-ever provisioning still re-signs once (the domain has genuinely never been seen). Every subsequent boot signs with `.hal` immediately from the stored value → no re-sign. "One re-sign per boot" becomes "one re-sign ever".

### Accepted tradeoff

A device that *permanently* moves to a network with no DHCP domain keeps the stale `.hal` SAN indefinitely (harmless — `.local` still works, and a stale SAN is simply never matched). This is the deliberate "never drop a previously-present FQDN on a transient empty resolution" choice; distinguishing transient from permanent absence is not worth the complexity.

### Explicitly rejected

`After=network-online.target` ordering — it would push the whole Traefik/Authelia/Homarr stack behind `NetworkManager-wait-online` (~60–90s) on every offline boot, just to name a cert SAN.

## Design

- New env seam `HALOS_HOSTNAMES_DOMAIN_STATE` (state-file path), default under `/var/lib/halos/`. Stickiness engages only when the path is set and its parent is writable; otherwise resolution falls back to today's stateless behavior, so it can never break and existing tests are unaffected.
- The three resolution sources in `_halos_resolve_domain` funnel through a single sticky + cache tail. The persist *write* is gated on `_halos_domain_safe` so injected/garbage values are never stored.

## Verification

- Unit (TDD): inject `hal` (persists), then inject empty → resolver still returns `hal`. Plus: genuine change overwrites; unwritable/unset state path falls back to stateless; garbage never persisted.
- `./run test` green.
- Field: reboot a device twice; confirm `halos-manage-certs` logs no "Hostname list or CA changed" re-sign on the second boot.

## Context

Extracted from #160, whose self-signed-revert thrust was dropped (no benefit once #161/#162 fixed the user-visible HSTS symptom). This was the only live work remaining there.

mairas:
## Why

`halos-manage-certs` re-signs the TLS leaf once on every boot.

The cert service is ordered `Before=halos-core-containers.service`, so it runs before NetworkManager has settled. The domain resolver chain (`hostname -d` → `nmcli`) returns nothing, the `${fqdn}` line in `hostnames.conf` soft-drops, and the leaf is signed with `.local`-only SANs — with the sentinel stored for that set. ~15 min later the renewal timer fires, DHCP has provided a domain, `${fqdn}` now expands to `<host>.hal`, the hostname hash changes, the sentinel mismatches, and the leaf is re-signed. Verified on halosdev.local: boot signed `[halosdev.local]`, the timer logged "Hostname list or CA changed, re-signing" and expanded to `[halosdev.hal, halosdev.local]`.

The `.hal` SAN is wanted (real unicast DNS, more robust than mDNS `.local`). The bug is the *instability of when it's included*, not the SAN.

## How

Make `_halos_resolve_domain` sticky: persist the last-known-good domain and reuse it when the live chain resolves nothing. After the first provisioning, every subsequent boot signs with the full SAN set immediately → stable hostname hash → no re-sign. A genuinely different live domain overwrites the stored value, so real domain changes still produce one legitimate re-sign.

Fixing this in the resolver (not in cert code) stabilizes the hostname set for every consumer — cert SANs, Authelia per-hostname cookies, and OIDC `redirect_uris`.

- State at `/var/lib/halos/resolved-domain`; override via `HALOS_HOSTNAMES_DOMAIN_STATE`, set-empty disables (test isolation). All state I/O is best-effort and never fails resolution (consumers run under `set -e`). Persisted writes are gated on `_halos_domain_safe`, so a garbage/injected value can't be replayed into cert SANs.
- Default path is fixed (not derived from `CONTAINER_DATA_ROOT`) so all loader consumers agree without each setting the env.

**Trade-off:** a device that *permanently* moves to a network with no DHCP domain keeps its old FQDN as a stale-but-harmless cert SAN until a domain resolves again. Deliberate — distinguishing transient from permanent absence isn't worth the complexity. `After=network-online.target` was rejected (it would push the whole stack behind `NetworkManager-wait-online` on every offline boot).

## Tests

5 new tests in `test-lib-hostnames.sh`: reuse-on-transient-empty, genuine-change-overwrites, disabled-keeps-stateless, garbage-not-persisted, and a consumer-level check that the hostname hash is identical across a resolving boot and a subsequent empty-resolution boot (the outcome that gates re-signing). Full suite green.

## Verify on a device

Reboot twice and confirm `halos-manage-certs` logs no "Hostname list or CA changed" re-sign on the second boot.

Closes #164. Extracted from #160 (closed — self-signed revert dropped, no benefit once #161/#162 fixed the HSTS symptom).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -128,8 +128,13 @@ hostname list:
 function whose stdout becomes the resolved domain. Gated on `declare -F`
 so a stray environment export cannot short-circuit production resolution
 or trigger command execution. When honored, the function's output
-(including empty) is authoritative — no fall-through. Tests use this to
-pin empty-domain behavior independently of the host network.
+(including empty) replaces the live-resolution chain (`hostname -d` /
+`nmcli`) — no fall-through to it. The sticky-domain layer still runs on
+top, exactly as it does for the real chain: with persistence enabled an
+injected empty can be backfilled from the persisted domain, and an
+injected (safe) value is persisted. Tests therefore pin empty-domain
+behavior by also controlling persistence — set `HALOS_HOSTNAMES_DOMAIN_STATE`
+empty to disable it, or point it at a scratch state file.
 
 **OIDC client snippet placeholder convention.** App `prestart.sh` scripts
 that drop OIDC client snippets into `/etc/halos/oidc-clients.d/` should
--- a/assets/hostnames.conf
+++ b/assets/hostnames.conf
@@ -17,8 +17,11 @@
 # Domain resolution order (first non-empty wins):
 #   1. `hostname -d` (admin-set via /etc/hosts or hostnamectl)
 #   2. `nmcli -t -f IP4.DOMAIN device show` (DHCP option 15, when NetworkManager is present)
-# When no domain resolves, lines using ${fqdn} or ${domain} are silently
-# skipped (logged as HALOS_HOSTNAMES_SKIP); other entries continue to apply.
+# The last-resolved domain is persisted, so once a domain has ever resolved
+# it is reused at early boot before the network settles (keeps the cert SANs
+# stable across reboots). Until a domain first resolves, lines using ${fqdn}
+# or ${domain} are silently skipped (logged as HALOS_HOSTNAMES_SKIP); other
+# entries continue to apply.
 #
 # Examples:
 #   ${hostname}.local        # mDNS -- works on the LAN without configuration
--- a/assets/lib-hostnames.sh
+++ b/assets/lib-hostnames.sh
@@ -90,7 +90,14 @@ _halos_domain_safe() {
 # Resolved values are validated by _halos_domain_safe; anything that fails
 # (whitespace, NUL, shell metacharacters, non-DNS characters) is treated
 # as no-domain and the chain falls through.
-# Echoes the empty string when nothing resolves; never errors.
+#
+# Sticky behavior: the last-known-good domain is persisted to
+# $HALOS_HOSTNAMES_DOMAIN_STATE (unset → /var/lib/halos/resolved-domain;
+# set-but-empty → disabled). When the live chain resolves nothing, the
+# persisted value is reused instead of returning empty, so a domain that
+# resolved on an earlier boot stays available while the network settles.
+# All state I/O is best-effort and never fails resolution. Echoes the empty
+# string when nothing resolves and nothing is persisted; never errors.
 #
 # Cached per `halos_load_hostnames` invocation in HALOS_HOSTNAMES_DOMAIN_CACHE.
 # Note: empty-string is a valid cached value (meaning "resolver completed,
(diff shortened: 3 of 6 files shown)


mairas:
## Code review — sticky domain resolution (high effort)

Self-review via multi-angle finders + verification. 8 findings, ranked. Nothing here is a merge-blocker on its own, but #1–#3 are worth fixing before merge.

### 1. Injection seam lost its early return — contract broken + test-seam values now hit disk (correctness)
`assets/lib-hostnames.sh:116` — The funnel restructure deleted the injection branch's `return 0`. Injection now falls through into the sticky block. Two consequences, both against the documented `HALOS_DOMAIN_RESOLVER` contract in AGENTS.md ("output, including empty, is authoritative — no fall-through"):
- An injected **empty** domain is now overwritten by a persisted value when state is enabled (the production default path).
- An injected domain that passes `_halos_domain_safe` is now **written to `/var/lib/halos/resolved-domain`** — the seam was a pure read-only short-circuit.
Unit tests miss both because `_reset_state` force-disables state. Fix: short-circuit the injection path before the sticky block (keep the early return for the seam). This also removes the only way a non-chain-validated value can reach the persist branch.

### 2. Shared fixed temp path races between concurrent loader runs (concurrency)
`assets/lib-hostnames.sh:170` — `${state_file}.tmp` is a fixed name. The cert timer run (`OnBootSec=15min` / `OnUnitActiveSec=24h`), a `prestart.sh` on container restart, and a manual `reload-oidc-clients` can overlap. Two writers to the same `.tmp`, then `mv` — `rename(2)` is atomic, but the *content* at rename time can be a truncated prefix of one writer's value. A truncated-but-still-valid label (`exampl` from `example.com`) passes the read-side `_halos_domain_safe` → wrong SAN persisted → wrong cert + spurious re-sign. Fix: unique temp (`${state_file}.$$`), or only write when the value actually changed (see #4, which subsumes this).

### 3. `/var/lib/halos` is never owned/created by the package (robustness)
`assets/lib-hostnames.sh:169` — Every other state dir is created in `postinst`; this one is lazily `mkdir -p ... 2>/dev/null || true` from a sourced library, and all persist I/O is silenced. If `/var/lib` is read-only/hardened (or the dir can't be created), stickiness silently no-ops **forever** — Bug A is not actually fixed in the field, with no diagnostic. The unit tests use tmp paths so they can't catch this. Fix: declare the dir via `tmpfiles.d`/`postinst` (or `StateDirectory=halos` on the units), and log a WARN when a persist write fails instead of swallowing it.

### 4. Write-on-every-resolve wears flash (efficiency)
`assets/lib-hostnames.sh:168` — Every domain-resolving boot rewrites an identical file (`mkdir` + tmp + `mv`). On SD/eMMC that's needless write wear for a value that only changes on first boot or a genuine domain change. Fix: read-before-write — reuse the sibling branch's `cat`+`_halos_trim`, skip the write when `stored == d`. Incidentally closes the race in #2 for the steady state.

### 5. Cert-renewal timer is now a writer of the SSO-identity domain (coupling/altitude)
`assets/lib-hostnames.sh:166` — The persisted domain drives cert SANs **and** Authelia per-hostname cookies / OIDC `redirect_uris` (rendered by `prestart.sh`). The 24h cert timer now resolves-and-overwrites that shared value; a transient `nmcli` domain captured by the timer gets locked in and later read by `prestart`. Cert renewal cadence and SSO identity now share a write path but not a refresh trigger. Worth either confining boot-time stickiness to the cert service or documenting the cross-service contract explicitly.

### 6. Comment documents the bug it fixes, not the why (maintainability)
`assets/lib-hostnames.sh:153` — `Sticky domain (fixes Bug A — one re-sign per boot)` is changelog framing. Per project convention comments explain *why the code does what it does* and reflect current state, not dev history / bug IDs. Rewrite to explain why the domain is persisted and reused; move the bug reference to the commit/PR.

### 7. Test harness disables the production code path (test coverage)
`tests/test-lib-hostnames.sh:41` — `_reset_state` sets `HALOS_HOSTNAMES_DOMAIN_STATE=""` for every test, so no test exercises the unset→default-path behavior that production uses. Combined with #1, the injection-empty regression is invisible to the suite. Add at least one test on the default-path branch (point it at a tmp dir, don't disable).

### 8. `garbage_not_persisted` test under-verifies its own claim (test quality)
`tests/test-lib-hostnames.sh` — It asserts only that boot-2 returns empty, which is guaranteed by the *read-side* guard regardless of whether garbage was persisted. If the persist-side guard regressed, the test would still pass. Assert the state file is absent/empty after the garbage boot.

---
Generated by `/code-review high`. Findings are advisory — your call on which to action before merge.

mairas:
## Fix plan (addressing the review above)

**Fixing in code:**
- **#2 (temp race):** unique temp file (`${state_file}.$$`) so concurrent loader runs can't collide on a shared `.tmp`.
- **#3 (dir ownership + observability):** create `/var/lib/halos` in `debian/postinst` alongside the other state dirs; log a `HALOS_HOSTNAMES_WARN` when a persist write fails instead of silently swallowing it.
- **#4 (flash wear):** read-before-write — skip the write when the stored value already equals the resolved domain. Steady-state boots do zero writes, which also closes #2's window outside genuine changes.
- **#8 (test under-verifies):** assert the state file is empty/absent after the garbage boot, so a regression in the persist-side `_halos_domain_safe` guard is actually caught.
- **#7 (partial):** add a "persistence enabled, nothing stored yet → returns empty" test (true-first-boot), proving stickiness never fabricates a domain.

**Rejecting with reasoning:**
- **#1 — keep injection feeding the sticky layer; fix the docs, not the code.** The proposed short-circuit breaks the test design: all sticky tests drive persistence through the `HALOS_DOMAIN_RESOLVER` seam, and short-circuiting removes the only controllable resolution source for testing persist/reuse. Production risk is nil — the seam is `declare -F`-gated, so it only fires when a defined shell function exists in the environment, which no production consumer provides. The real defect is the stale AGENTS.md wording ("authoritative — no fall-through"); reconciling that.
- **#5 — document the shared-state contract; do not re-architect.** Two independently-triggered services resolving the domain predates this PR. At boot they're ordered (`halos-manage-certs` `Before=halos-core-containers`) and both read the same persisted value on empty live resolution, so this change makes them *more* consistent. Confining the write to one service would defeat the cross-consumer consistency that's the point of the fix. Documenting the contract in HOSTNAMES.md.

**#7 (full default-path coverage):** the enabled-state branch is covered via explicit state paths; the literal unset→`/var/lib/halos` default is a hardcoded string untestable without host pollution. Not adding a host-polluting test.

mairas:
## Review fixes pushed (`9f00946`)

**Fixed in code:**
- **#2 (temp race):** per-PID temp `${state_file}.$$` — concurrent loader runs can't truncate each other's write. `assets/lib-hostnames.sh` persist branch.
- **#3 (dir ownership + observability):** `mkdir -p /var/lib/halos` added to `debian/postinst` alongside the other state dirs; a failed persist now logs `HALOS_HOSTNAMES_WARN: could not persist resolved domain … (stickiness inactive — cert may re-sign each boot)` instead of silently swallowing.
- **#4 (flash wear):** read-before-write — the resolver reads the stored value and skips the write when unchanged. Steady-state boots do zero writes, which also closes #2's window outside genuine domain changes.
- **#6 (comment):** rewritten to explain why (stable domain across reboots for downstream consumers); bug-ID/dev-history framing removed. Done in the prior push.
- **#8 (test under-verifies):** `test_sticky_domain_garbage_not_persisted` now asserts the state file is empty after the garbage boot — the load-bearing check the old d2-only assertion was missing.
- **#7 (coverage):** added `test_sticky_domain_first_boot_returns_empty` (persistence enabled, nothing stored → returns empty, no file created) so the enabled branch is exercised end to end.

**Pushed back, addressed by docs instead of code:**
- **#1 (injection contract):** kept injection feeding the sticky layer — short-circuiting it would remove the only controllable resolution source the sticky tests use, and the production risk is nil (the seam is `declare -F`-gated; no production consumer defines such a function). Reconciled the stale "authoritative — no fall-through" wording in `AGENTS.md` to state that injection replaces the live chain but the sticky layer still runs on top, and that tests pin empty-domain behavior by controlling `HALOS_HOSTNAMES_DOMAIN_STATE`.
- **#5 (cert-timer/SSO coupling):** not a regression — two independently-triggered services resolving the domain predates this PR, and at boot they're ordered (`halos-manage-certs` `Before=halos-core-containers`) and both read the same persisted value, so this change makes them *more* consistent. Documented the shared-state contract and the refresh boundary (cert re-signs on change; `prestart`-rendered Authelia/OIDC config realigns on the next service restart) in `docs/HOSTNAMES.md`.

**Not done (reasoned):** the literal unset→`/var/lib/halos` default path is a hardcoded string untestable without writing to the host; the enabled-state logic is fully covered via explicit scratch paths, so no host-polluting test was added.

**Verification:** full `./run test` green — `lib-ca` 76, `lib-hostnames` 45 (6 sticky), `manage-certs` 20, `traefik-regex` 10. `shellcheck -x` clean on the changed regions of `lib-hostnames.sh` and on `postinst`. Pre-commit hooks (changelog, hostname) pass.

Issue #159, opened 2026-05-28
Effortless HaLOS CA trust on remote clients


mairas:
## Goal

Make HaLOS CA installation effortless on external clients so the per-port click-through regression introduced by the CA-signed leaf (vs. the prior self-signed leaf) is reversed for any user who completes the install. Click-through users (who decline install) retain functional dashboard access — they're not the target of this work.

## Context

Pre-#142 (extract-cert-management): self-signed leaf. Brave/Chromium treats self-signed certs as host-wide trust anchors after a single click-through.

Post-#142: CA-signed leaf. Chromium switches to per-(host, port, leaf-fingerprint) exception semantics. Cards on the Homarr dashboard redirect across port boundaries (`:4430-:4450` per `configure-container-routing`), so each visit to a new app triggers a fresh interstitial. Icon subresources silently fail until their host port has been click-throughed.

Chosen path: keep CA-signed leaves, lean on making CA install painless enough that users actually do it. Empirically confirmed in this session by hot-fixing the device to self-signed and observing one-click-covers-all behavior in Brave.

## Locked decisions

| Decision | Value |
| --- | --- |
| Platforms in scope, priority order | macOS, Android, Linux, iOS |
| Windows | Out of scope this round (manual \`.crt\` only) |
| Dashboard tile | `webapps.d` entry; no new card kind |
| Tile lifecycle | Always visible (no auto-hide) |
| Apple UX | Two-step walkthrough with trust-toggle screenshots; no claim of "one click" |
| Artifact set | `.mobileconfig` (Apple), `.deb` (Debian/Ubuntu), `.crt` (Android/other Linux) |
| Landing URL | `/ca` (HTML) |
| Existing `/halos-ca.crt` | Move under `/ca/halos-ca.crt`; recent endpoint, docs update lands together |
| Sidecar | Extend existing `ca-download` nginx sidecar (no new container) |
| User-Agent detection | Yes, expand the relevant OS section by default |
| Device-side self-trust | Out of scope here; tracked separately in #158 |

## Critical UX reality

[...]

PR #166, opened 2026-05-29
feat(certs): move CA download under /ca/ + landing scaffold (effortless CA trust, unit 1)

mairas:
## Why

Unit 1 of effortless CA trust (#159). The CA-signed leaf (since #142) means Chromium uses per-(host, port) cert exceptions, so dashboard navigation across app ports triggers repeated interstitials. The agreed path is to make CA install painless via a guided landing page rather than revert to self-signed. This unit lays the URL/serving foundation: it moves the CA download under `/ca/` so the landing page and its assets share one prefix, and ships a placeholder page. No real install UI yet (that's Unit 2).

## How

- **nginx** (`assets/ca-download/nginx.conf`): a `/ca/` location tree.
  - `= /ca/halos-ca.crt` serves the active CA (headers/guards unchanged from the old endpoint).
  - `/ca/` serves the landing dir.
  - `= /halos-ca.crt` (old path) returns **410 Gone** with `Link: </ca/halos-ca.crt>; rel="canonical"`.
  - `include mime.types` added so landing assets get correct content types.
- **compose** (`docker-compose.yml`): both routers rule on `PathPrefix(\`/ca/\`) || Path(\`/halos-ca.crt\`)` (the legacy path kept so the sidecar serves its own 410). Healthcheck probes `/ca/halos-ca.crt`. Landing mounts at **`/srv-landing`** — a sibling of the read-only `/srv` mount, not nested (see below).
- **landing** (`assets/ca-download/landing/index.html`): placeholder — heading, explanation, direct cert link. Replaced in Unit 2.
- **tests** (`tests/test-ca-download-endpoint.sh`): runs the real config in a throwaway `nginx:alpine` against a dummy cert + the real landing asset; asserts the new layout, the 410 + hint, and the read-only guard. Skips cleanly (exit 0) where Docker is unavailable.
- **docs** (`docs/CERTS.md`): endpoint URLs + a move note.

## Notes for review

- **A nested `/srv/landing` mount would break the container.** The plan specified `landing:/srv/landing:ro`, but `/srv` is itself a read-only bind mount, so Docker can't create the nested mountpoint (`read-only file system` at container init — reproduced locally). Mounted as sibling `/srv-landing` instead; nginx alias updated to match.
- **POST returns 403, not 405.** The plan's test scenario expected `405 + Allow`, but the existing `limit_except GET HEAD { deny all; }` yields 403. Preserving it (a 405 would be an unasked-for behavior change); the test matches reality.
- Local test covers the nginx layer (10/10). The Traefik `PathPrefix(/ca/)` routing is verified separately on the device.

## Release notes

The CA download URL **moves** from `https://<host>/halos-ca.crt` to `https://<host>/ca/halos-ca.crt`. The old path now returns `410 Gone` with a canonical `Link` hint. Any operator script or bookmark using the old URL must update.

## Verification

`./run test` green (ca-download 10, manage-certs 20, lib-ca 76, lib-hostnames 45, traefik-regex 10). Device verification of the Traefik routing to follow on this PR.

Refs #159

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -224,17 +224,19 @@ services:
       - proxy
     labels:
       - traefik.enable=true
-      # HTTP — /ca/ prefix plus the legacy /halos-ca.crt path (which the
-      # sidecar answers with a 410 + canonical hint), redirected to HTTPS.
-      # Priority 200 beats homarr's catch-all PathPrefix(`/`) at priority 1.
-      - traefik.http.routers.ca-download-http.rule=PathPrefix(`/ca/`) || Path(`/halos-ca.crt`)
+      # HTTP — /ca/ prefix, the bare /ca (which the sidecar 308-redirects to
+      # /ca/), plus the legacy /halos-ca.crt path (answered with a 410 +
+      # canonical hint), redirected to HTTPS. Priority 200 beats homarr's
+      # catch-all PathPrefix(`/`) at priority 1.
+      - traefik.http.routers.ca-download-http.rule=PathPrefix(`/ca/`) || Path(`/ca`) || Path(`/halos-ca.crt`)
       - traefik.http.routers.ca-download-http.entrypoints=web
       - traefik.http.routers.ca-download-http.middlewares=redirect-to-https@file
       - traefik.http.routers.ca-download-http.priority=200
-      # HTTPS — /ca/ prefix plus the legacy /halos-ca.crt path. No
-      # ForwardAuth: this is a public trust anchor; gating it would defeat
-      # the chicken-and-egg bootstrap.
-      - traefik.http.routers.ca-download-secure.rule=PathPrefix(`/ca/`) || Path(`/halos-ca.crt`)
+      # HTTPS — /ca/ prefix, the bare /ca (308-redirected to /ca/ by the
+      # sidecar), plus the legacy /halos-ca.crt path. No ForwardAuth: this is
+      # a public trust anchor; gating it would defeat the chicken-and-egg
+      # bootstrap.
+      - traefik.http.routers.ca-download-secure.rule=PathPrefix(`/ca/`) || Path(`/ca`) || Path(`/halos-ca.crt`)
       - traefik.http.routers.ca-download-secure.entrypoints=websecure
       - traefik.http.routers.ca-download-secure.tls=true
       - traefik.http.routers.ca-download-secure.priority=200
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -183,8 +183,9 @@ services:
         tag: "{{.Name}}"
 
   ca-download:
-    # Sidecar that publishes the active device CA at /halos-ca.crt so
-    # operators can install one trust anchor for all services on this device.
+    # Sidecar that publishes the active device CA at /ca/halos-ca.crt plus a
+    # trust-install landing page at /ca/, so operators can install one trust
+    # anchor for all services on this device.
     # Routed by Traefik (no host port published). prestart.sh copies the
     # active CA's PEM bytes into the mounted dir each run.
     image: nginx:1.27-alpine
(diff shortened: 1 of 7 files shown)


docker-compose.yml
```diff
@@ -219,15 +224,17 @@ services:
...
       - traefik.http.routers.ca-download-http.entrypoints=web
       - traefik.http.routers.ca-download-http.middlewares=redirect-to-https@file
       - traefik.http.routers.ca-download-http.priority=200
-      # HTTPS — exact-path match. No ForwardAuth: this is a public trust
-      # anchor; gating it would defeat the chicken-and-egg bootstrap.
-      - traefik.http.routers.ca-download-secure.rule=Path(`/halos-ca.crt`)
+      # HTTPS — /ca/ prefix plus the legacy /halos-ca.crt path. No
+      # ForwardAuth: this is a public trust anchor; gating it would defeat
+      # the chicken-and-egg bootstrap.
+      - traefik.http.routers.ca-download-secure.rule=PathPrefix(`/ca/`) || Path(`/halos-ca.crt`)
```
mairas: Confirmed on the device: a bare `/ca` (no trailing slash) returns **307 and never reaches the landing page**. `PathPrefix(`/ca/`)` only matches `/ca/…`, and nginx `location /ca/` also needs the slash — so the controlled entry points (tile URL, docs) all use `/ca/`, but a human typing `https://<host>/ca` misses the page entirely.

Minor/optional, but for a discoverability-focused trust page it's worth handling. The full fix is two-part: add `Path(`/ca`)` to **both** ca-download router rules (this one and the http one above) **and** add an nginx `location = /ca { return 308 /ca/; }` — Traefik routing alone isn't enough, since nginx `location /ca/` won't match the bare `/ca`. Your call whether it's worth it for unit 1 or a follow-up.

docs/CERTS.md
```diff
@@ -198,22 +202,22 @@ chicken-and-egg procedure above):
...
     | diff - /tmp/halos-ca.crt && echo OK
 
 # 2. Headers are correct (Content-Type, Content-Disposition).
-curl -skI https://halos.local/halos-ca.crt | grep -iE 'content-(type|disposition)'
+curl -skI https://halos.local/ca/halos-ca.crt | grep -iE 'content-(type|disposition)'
 
 # 3. HTTP redirects to HTTPS.
-curl -sI http://halos.local/halos-ca.crt | head -1   # 308 Permanent Redirect
-curl -sI http://halos.local/halos-ca.crt | grep -i ^location
+curl -sI http://halos.local/ca/halos-ca.crt | head -1   # 308 Permanent Redirect
```
mairas, line 213: Device shows this redirect is a **307 Temporary Redirect**, not 308 — so this example comment is inaccurate. Pre-existing (the redirect-to-https middleware behavior, not changed here), but the line is in this diff so worth correcting while we're here.
```suggestion
curl -sI http://halos.local/ca/halos-ca.crt | head -1   # 307 Temporary Redirect
```

mairas (commented):
Self-review of unit 1. CI is green, and I device-verified on halosdev: `/ca/` → 200 text/html (landing body present), `/ca/halos-ca.crt` → 200 `application/x-x509-ca-cert`, old `/halos-ca.crt` → 410 with the canonical `Link` header, and `ca-download` comes up **healthy** — so the `/srv-landing` sibling mount (vs the plan's nested `/srv/landing`) holds in production.

Two minor findings inline, neither a blocker. The 403-vs-405 read-only response and the sibling mount are both intentional and covered in the PR description.

docs/CERTS.md
mairas: Retracting this — the doc is correct, my evidence was confounded.

The doc example uses the **canonical host** (`halos.local`). Re-tested cleanly: `http://halosdev.local/ca/halos-ca.crt` (canonical host) → **308 Permanent Redirect** via the `redirect-to-https` middleware, exactly as documented.

My original 307 came from curling `localhost`, which hits a *separate* middleware (`localhost-redirect.yml`, localhost→mDNS canonicalization, `permanent: false`) — a different redirect than the one this line documents. **Do not apply the suggestion**; "308" is right.

docker-compose.yml
mairas: Fixed in c089ef0. Both router rules now match `Path(`/ca`)` and nginx has a `location = /ca` that 308-redirects to `/ca/`.

Device-verified on halosdev: `/ca` → 308 with a **relative** `Location: /ca/`, and following it reaches the landing page in a **single https hop** (200, num_redirects=1).

One extra catch while verifying: the redirect needs `absolute_redirect off`. nginx sits behind Traefik's TLS termination on plain `:80`, so a default absolute redirect emits `http://<host>/ca/` and bounces the client through an http downgrade before Traefik's http→https redirect catches it. With `absolute_redirect off` the Location is relative and the client stays on https. (The local docker test couldn't see this — no TLS proxy in front — so I tightened it to assert the Location is relative.)

PR #168, opened 2026-05-31
feat(certs): per-OS CA-trust landing page and dashboard tile

mairas:
## Why this exists

HaLOS serves every web surface behind a CA-signed leaf certificate (since #142). The catch: Chromium treats a CA-chained leaf with per-`(host, port, fingerprint)` exception semantics, not the host-wide trust a self-signed leaf got. The Homarr dashboard redirects across port boundaries (`:4430–:4450`), so a user who clicks through the warning on one card hits a fresh interstitial on the next, and icon subresources silently fail until each port has been click-throughed individually.

The epic (#159) keeps the CA-signed leaves and instead makes installing the device CA painless enough that users actually do it — one trust anchor, then every surface validates. Unit 1 (#166) moved the download to `/ca/` and left a placeholder. This PR ships the real guidance page and puts it on the dashboard.

## What this PR does

**Landing page (`/ca/`).** Replaces the placeholder with a self-contained per-platform walkthrough for installing the CA's `.crt`: macOS, iOS/iPadOS, Android, Linux, and Windows. Inline CSS and JS, zero external assets — it has to work on a freshly-flashed device with no upstream network. The page is built as collapsible `<details>` sections; client-side User-Agent sniffing opens the section matching the visitor's OS, and with JavaScript disabled every section is simply visible (progressive enhancement, no dead-ends).

**The honesty framing is the point, not decoration.** On every current platform, *installing* a certificate and *trusting* it for TLS are two separate actions, and users routinely stop after the first. So the page leads with "installing the certificate is not enough on its own," and each section names exactly where the trust step lives:

- **macOS** — Keychain Access, then set *Always Trust* on the cert (the `.crt` + Keychain path sidesteps the Sequoia configuration-profile trust regression entirely; the trust toggle here is the Keychain setting).
- **iOS/iPadOS** — install the profile, *then* Settings → General → About → **Certificate Trust Settings** and flip the toggle. iOS 17/18 require this second step; installing alone does nothing for Safari.
- **Android** — install via Settings, with the caveat that Android trusts user CAs in browsers only, never in apps (by Android design) — stated so users don't file "the app still warns" as a bug.
- **Linux** — `update-ca-certificates` for the system store (Chromium/curl), plus the separate Firefox NSS import, since Firefox ignores the system store.
- **Windows** — manual *Trusted Root Certification Authorities* install. #159 defers a Windows *installer*, but the raw `.crt` works today, so the cheap manual section ships rather than leaving Windows users with nothing.

**Dashboard tile.** `assets/ca-download-tile.toml` → `/etc/halos/webapps.d/ca-download.toml`, synced onto the Homarr board by `homarr-container-adapter` (same registry mechanism as `cockpit.toml`). External link (no health check), always visible, utility-band priority. Ships with a square icon (`assets/halos-ca-download.svg` → `/usr/share/pixmaps/`, served to Homarr as `/icons/...`).

## Decisions worth flagging for review

- **Text-only steps for v1.** No screenshots yet — authentic iOS/macOS Settings captures are a human asset drop, and text steps are honest and useful now. The carrier work (below) and screenshots layer in later without reworking the markup.
- **Icon is a recrop, not a redraw.** The original art (`ca_download_icon.svg`, untracked, from an earlier abandoned attempt) was 2816×1536 landscape. I measured the artwork's bounding box by rendering it and reading the non-transparent pixel bounds, then set a centered square viewBox — the geometry is unchanged, just reframed to a tile-appropriate aspect. It keeps the original slate `#516169` fill; on the dark board it may read slightly dim — easy to lighten if reviewers prefer.
- **Deferred-carrier seam.** The download button points at `/ca/halos-ca.crt` for all platforms today. The `.mobileconfig` (Apple) and `.deb` (Debian/Ubuntu) carriers are tracked separately and will replace the raw `.crt` link in their own sections as they land — the per-section structure exists precisely so that swap is local.

## Packaging detail

`debian/rules` installs an entire `assets/<dir>/` tree via glob (so compose bind-mount sources never drift). Both new files are placed *directly* under `assets/` (not in a subdir), so the glob doesn't sweep them into the container lib dir; they're installed explicitly to their host paths, mirroring the existing `hostnames.conf` handling. The block carries a comment explaining this so the next person doesn't "tidy" them into a subfolder and silently relocate them.

## Tests

`test-ca-download-endpoint.sh` gains six assertions — one unmistakable marker per platform (`Always Trust`, `Certificate Trust Settings`, `CA certificate`, `update-ca-certificates`, `Trusted Root Certification Authorities`) plus the trust-vs-install caveat. A regression to the placeholder, or a half-built page missing a platform, now fails loudly here instead of shipping a dead trust experience.

## Verification

- `./run test`: no failures across all suites; the ca-download endpoint test is 18/18.
- Tile TOML parses; `url = "/ca"` is valid per `homarr-container-adapter`'s registry URL validator.
- Landing page rendered in a real browser: layout clean, UA auto-expand fires, the install-≠-trust callout is prominent above the fold.
- **Not** locally built into a `.deb` — that needs the debtools container, and building on the dev host is fine but the package build is exercised by CI. The `debian/rules` additions are two `install -D` lines mirroring an existing pattern, so the risk is low but the build is CI-verified, not locally verified.

## Version

No `VERSION` bump. `VERSION` is `0.4.2`, the latest stable tag is `v0.4.1+3`; they already differ, so a prior PR opened this release cycle and the per-release-cycle bump is already spent. CI's `version-bump-check` agrees (passing).

Refs #159


mairas:
## Code review — PR #168 (per-OS CA-trust landing page + dashboard tile)

Tiered persona review (6 reviewers: correctness, testing, maintainability, project-standards, api-contract, adversarial). Findings merged, deduplicated, confidence-gated. Two factual claims (CI test execution, icon serving path) were verified directly against the repo, not relayed.

**Intent:** Unit 2 of #159 — replace the `/ca/` placeholder with a per-OS `.crt` walkthrough, add a Homarr tile + icon, wire into the Debian package. Stopgap `.crt` flow until the `.mobileconfig`/`.deb` carriers land.

### P1 — High

|---|------|-------|-------------|------|-------|
| 1 | `assets/ca-download/landing/index.html` (iOS section) | The iOS steps describe a "Profile Downloaded → VPN & Device Management → Install" flow that likely **does not occur as written**. The cert is served with `Content-Disposition: attachment` (nginx.conf) plus the `download` attribute on the anchor — on iOS Safari that routes the file to the Files app instead of triggering the profile-install prompt the steps assume. Two reviewers independently flagged iOS as the weakest section. Real-device verification was deferred (text-only v1), so this is unconfirmed but plausible and user-facing on an in-scope platform. | correctness, adversarial | 0.80 | manual / human · **verify on device** |
| 2 | `tests/test-ca-download-endpoint.sh` | **The new content assertions never run in CI.** Verified: `.github/actions/run-tests/action.yml` runs the shell suite inside a `debian:trixie` container with no Docker socket, so this test hits its `command -v docker` guard and self-skips. The six per-platform markers (and all pre-existing HTTP assertions) are dead in CI — a rollback to the placeholder would merge green. | testing | 0.97 | see fix below |

**On #1:** options — (a) verify the exact flow on a current iPhone and correct the wording; (b) add the Files-app fallback step; (c) mark the iOS section provisional until the `.mobileconfig` carrier lands. A header change to fix this lives in `nginx.conf` (Unit 1's file), so it may be out of this PR's scope — verification first.

**On #2:** the *contained* fix is to move the per-marker content checks into a **Docker-free** test that greps `assets/ca-download/landing/index.html` on disk — those checks need no sidecar and would then actually run inside trixie. The broader gap (the HTTP endpoint tests — status/headers/308/403 — never run in CI) is **pre-existing** since Unit 1 and should be a tracked issue, not fixed here. See Pre-existing below.

### P2 — Moderate

|---|------|-------|-------------|------|-------|
| 3 | `assets/ca-download-tile.toml:9` | `url = "/ca"` lacks the trailing slash. Every other path-only tile uses one (`cockpit.toml` → `/cockpit/`). `external = true` means no health-check breakage today, but the tile link incurs an extra `308 /ca → /ca/` hop, diverges from the canonical landing URL, and is a latent trap if `external` is ever flipped to `false` (ping target would 308, reporting the app down). | api-contract | 0.82 | safe_auto — change to `/ca/` |
| 4 | `assets/ca-download-tile.toml` | No test asserts the tile TOML parses, has the required fields, and that `icon_url` matches the SVG's install path. The plan explicitly called for this, and it's Docker-free so it would run in CI. | testing (echoed by 3) | 0.88 | gated_auto — add parse/field test |
| 5 | `assets/ca-download-tile.toml` → `ca-download.toml` | Source is named `ca-download-tile.toml` but installs as `ca-download.toml`; grepping the installed name finds no source. The SVG keeps its name end-to-end. Consider renaming source to `assets/ca-download.toml` (the glob matches directories only, so a flat file is safe). | maintainability | 0.82 | discretionary |

### P3 — Low

|---|------|-------|-------------|------|-------|
| 6 | `index.html` (UA script) | UA auto-expand can misfire: iPadOS reports `MacIntel` and relies on `maxTouchPoints > 1`, which Brave's fingerprint protection can force to 0 → falls through to the macOS section. `navigator.platform` is also deprecated. Low stakes — sections stay clickable; only the auto-open is wrong. | correctness, maintainability, adversarial | 0.82 | advisory |
| 7 | `tests/test-ca-download-endpoint.sh:107` | Android marker `"CA certificate"` is not section-specific — it also appears in the download button text (`Download the HaLOS CA certificate`), so deleting the Android section wouldn't fail the test. Assert section-unique strings or the structural `<details id="…">` markers (which the JS also depends on). | testing, maintainability | 0.85 | safe_auto |
| 8 | `docs/CERTS.md` | "with JavaScript off, every section is visible" is inaccurate — `<details>` render **collapsed** by default; with JS off the user must click each. (The PR description has the same overstatement.) Reword to "accessible — click any heading to expand," or add `open` to all sections and have JS close the non-matching ones. | adversarial | 0.82 | safe_auto |
| 9 | `debian/rules` | No check that the `.deb` actually contains the TOML and SVG at their install paths. A destination typo passes lintian and deploys without the tile/icon. Add `dpkg-deb --contents … | grep`. | testing | 0.75 | gated_auto |

| Issue | Route |
|-------|-------|
| The HTTP endpoint tests in `test-ca-download-endpoint.sh` (status codes, headers, 308, 403) have **never executed in CI** — the Docker-gated suite has self-skipped inside the trixie container since Unit 1. This is infrastructure, not introduced here, but it means the ca-download endpoint has had no CI coverage. | **file-issue** — should be tracked before relying on these tests as a gate |

### Advisory / residual (no action needed)

- The "flat file directly under `assets/` escapes the `assets/*/` install glob" convention is load-bearing but enforced only by a comment. A future contributor adding an asset under a new subdir would silently misroute it. (maintainability)
- If `ca-download.toml` ever fails to parse, the adapter logs a warning and drops just that tile (correct skip-and-continue), but the tile vanishes with no operator-visible error. (adversarial)
- Per-platform steps live in both the HTML and CERTS.md and will drift as OS settings move. (maintainability)

### Coverage

- Suppressed: low-confidence noise below 0.60 dropped during merge.
- **project-standards: clean** — VERSION policy reasoning confirmed correct (`0.4.2` already differs from `v0.4.1`, no bump needed); `debian/changelog` untouched; no hard-coded hostnames in added lines.
- **api-contract: icon chain verified working** — `/usr/share/pixmaps` is bind-mounted read-only into the Homarr container and `transform_icon_url` maps to `/icons/halos-ca-download.svg`; the new icon will render.
- Untested: UA-detection JS (no headless-browser test), `.deb` install paths, tile parse via the real adapter.

---

### Verdict: Ready with fixes

Nothing is P0; the package builds and the page renders. Recommended before merge:

1. **iOS (#1)** — verify on a real device or mark the section provisional. Don't ship instructions that strand users.
2. **CI coverage (#2)** — make the content-marker checks Docker-free so they actually run; file an issue for the pre-existing endpoint-test skip.
3. **`url` → `/ca/` (#3)** and **add the TOML parse test (#4)** — both trivial, both run in CI.

#6–#9 are discretionary. The pre-existing CI-skip (PE) should get a tracking issue regardless.

mairas:
Addressed in 132615f:

- **#2 (tests dead in CI)** — moved the per-OS content guard out of the Docker-gated endpoint test into a Docker-free pytest (`tests/test_landing_content.py`) that reads the HTML directly and runs in CI via the pytest step. Markers are now section-unique (fixes **#7**).
- **#4** — added `tests/test_ca_download_tile.py`: parses the tile, checks required fields, and asserts `icon_url` matches the SVG install path in `debian/rules`.
- **#3** — tile `url` `/ca` → `/ca/`.
- **#8** — corrected the CERTS.md no-JS claim (sections are collapsed-but-accessible).
- `./run test` now also runs pytest locally.

Still open, tracked as issues:
- **#1 iOS accuracy** → #169 (needs real-device verification)
- **Pre-existing endpoint-test CI skip** → #170

#5 (source filename), #6 (UA-detection edge), #9 (.deb content check) left as discretionary.

PR #173, opened 2026-05-31
feat(certs): Apple .mobileconfig CA-trust carrier

### LINKED ISSUE #169: Verify and correct iOS CA-install steps on the /ca/ trust landing page
Surfaced by review of #168 (two reviewers independently flagged the iOS section).

## Problem

The iOS section of the trust landing page (`assets/ca-download/landing/index.html`) describes:

1. "iOS shows *Profile Downloaded*"
2. Settings → General → VPN & Device Management → tap the downloaded profile → Install
3. Settings → General → About → Certificate Trust Settings → enable toggle

Two concerns about whether step 1–2 happen as written for the current `.crt` download:

- The cert is served with `Content-Disposition: attachment` (`assets/ca-download/nginx.conf`) plus a `download` attribute on the anchor. On iOS Safari that can route the file to the **Files app** instead of triggering the OS profile-install prompt the steps assume — the user never sees "Profile Downloaded" and gets stuck at step 2.
- The exact "Profile Downloaded" wording / VPN & Device Management path should be confirmed against current iOS (17/18) behavior for a raw `.crt`.

This was not caught earlier because real-device testing was deferred (text-only v1).

## Options

- Verify on a current iPhone and correct the wording.
- Add the Files-app fallback step ("open Files, tap the downloaded `.crt`, then…").
- Reconsider the `Content-Disposition: attachment` header for the cert MIME type (lives in the Unit 1 nginx config) — the `application/x-x509-ca-cert` type alone may trigger the install prompt without forcing a download.
- Or mark the iOS section provisional until the `.mobileconfig` carrier lands (the carrier is the real fix).

## Acceptance

iOS section steps match what a user actually sees on a current iPhone, end to end, OR the section is explicitly marked provisional with a working fallback path.

Refs #159

mairas:
## What

Adds the Apple `.mobileconfig` carrier (Phase 2, Unit 5 of #159) and rewires the Apple landing-page sections around it. Closes #169.

`halos-manage-certs` now generates `halos-ca.mobileconfig` next to the public CA PEM on every run (new `halos_ca_publish_mobileconfig` helper in `lib-ca.sh`, aux-failure semantics — a generation failure logs WARN and never blocks cert management). The ca-download sidecar serves it at `/ca/halos-ca.mobileconfig`.

## Why this fixes #169

The raw `.crt` is served `Content-Disposition: attachment`; on iOS Safari that routes the file to the Files app and the profile installer never fires, so the documented iOS steps ("Profile Downloaded" → VPN & Device Management) never happened. The `.mobileconfig` is served **inline** with the `application/x-apple-aspen-config` MIME and no attachment disposition, and the landing link has no `download` attribute — so Safari hands it to the profile installer. The iOS section now leads with the profile; the corrected raw-`.crt` fallback (which honestly says the file lands in Files) moves into an "Advanced" disclosure.

## Approach notes

- **Profile**: single `com.apple.security.root` payload with the active CA's DER, `PayloadDisplayName = HaLOS Device CA (<hostname>)`. Unsigned (no Apple Developer ID) — cosmetic "Not Verified" notice, called out in the copy. Fixed UUIDs / identifiers so a CA rotation replaces the profile instead of stacking a duplicate; output is byte-deterministic for a given CA + hostname.
- **Install ≠ trust** stays honest: macOS Keychain "Always Trust" / iOS Certificate Trust Settings toggle remain, on the page and in docs.
- **Inline plist heredoc** (not a separate template asset): `lib-ca.sh` installs to `/usr/lib/<pkg>/` while `assets/` trees install into the container-app data dir, so an external template would couple the helper to a second install path. Mirrors the existing inline leaf-ext heredoc.
- **macOS uses the raw `.crt`, not the profile** (device-tested): a profile-delivered root isn't trusted for SSL on macOS Ventura+, doesn't appear in Keychain Access, and macOS has no Certificate Trust Settings toggle — so the profile is a dead end there. macOS keeps the raw-`.crt` + Keychain "Always Trust" flow; the profile is iOS-only.

## Tests

- `test-lib-ca.sh` (+8): DER round-trips out of the generated plist (fingerprint match), display-host embedded, byte-deterministic, missing-CA/non-cert → rc1 + preserve-existing, missing-args → rc2.
- `test-halos-manage-certs.sh` (+2): `.mobileconfig` published and well-formed after a run.
- `test-ca-download-endpoint.sh` (+3, Docker-gated): `/ca/halos-ca.mobileconfig` → 200, aspen MIME, **no** attachment disposition.
- `test_landing_content.py` (+4): both Apple sections link to the profile; raw-`.crt` demoted to Advanced; iOS Files-app fallback present.
- Verified on macOS: generated profile passes `plutil -lint`; Apple's plist parser round-trips the embedded DER back to the real CA (`CA:TRUE`).

## Release notes

- New public URL: `https://<device>/ca/halos-ca.mobileconfig` (Apple configuration profile carrying the device CA).
- macOS/iOS dashboard trust page now offers the profile as the primary install path.

## Versioning

Bumps `VERSION` 0.4.2 → 0.4.3 — this PR opens the next release cycle after the 0.4.2 stable cut (`v0.4.2+8`). Later carrier units in this cycle won't bump again.

Refs #159
Closes #169


mairas:
## Code review — PR #173 (tiered persona review)

**Scope:** `feat/ca-mobileconfig-carrier` vs `main` (1 commit, 11 files). **Intent:** generate + serve an unsigned Apple `.mobileconfig` carrying the device CA, served inline to fix the iOS Files-app routing (#169), with Apple landing sections rewired.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + security, reliability, adversarial (cert/trust-anchor, public endpoint, 160 LOC). 7 of 7 returned.

### P1 — High

|---|------|-------|-----------|------|
| 1 | `assets/lib-ca.sh` | **Fleet collision: fixed `PayloadUUID`/`PayloadIdentifier` are identical across every device.** Apple keys profiles by identifier/UUID, so on a Mac/iPhone that already trusts device A, installing device B's profile *silently replaces* device A's CA — device A's HTTPS then fails validation with no warning. `display_host` only feeds `PayloadDisplayName`, which has no effect on identity. This directly breaks the multi-device/fleet use case in #159. | security, adversarial | 0.95 |

**Fix (applying in this PR):** derive both UUIDs and identifiers from the CA's SHA-256 fingerprint (per-device, high-entropy → collision-free across devices; stable for a given CA so same-device behaviour is unchanged). Trade-off: a CA *rotation* then yields a new profile rather than replacing the old one in-place — acceptable, since rotation already requires operators to re-trust, and an orphaned old entry beats silently evicting a *different device's* trust.

### P2 — Moderate

|---|------|-------|-----------|------|
| 2 | `assets/lib-ca.sh` | `der_b64="$(base64 < "$der_tmp" \| tr -d '\n')"` captures `tr`'s exit status, not `base64`'s — a mid-read failure embeds truncated DER past the empty-only guard. Same path leaks `der_tmp` on the empty-return branch (every other branch cleans it). | correctness, reliability, maintainability | 0.85 |
| 3 | `tests/test-lib-ca.sh` | No `directory-at-output` test for the mobileconfig helper (the `[ -d "$out_file" ]` guard is untested; `publish_public` has the parallel test). | testing | 0.95 |
| 4 | `tests/test-halos-manage-certs.sh` | Integration test asserts only `HaLOS Device CA (` — passes even if the hostname arg is dropped (`...()`); comment says `device.local` but the harness uses `fixture.test`. | testing, maintainability | 0.82 |

**Fixes (this PR):** sequential `base64` capture with explicit status + unconditional `der_tmp` cleanup; add the directory-at-output and a no-debris-on-success test; tighten the integration assertion to `HaLOS Device CA (fixture.test)` and fix the comment.

### P3 — Low

|---|------|-------|-----------|------|
| 5 | `landing/index.html` | iOS "Advanced" fallback claims the raw `.crt` shows a *Profile Downloaded* prompt — that wording is `.mobileconfig`-only; a raw cert from Files shows a different dialog. | adversarial | 0.82 |
| 6 | `landing/index.html` | macOS step assumes the Ventura+ "VPN & Device Management" pane; pre-Ventura uses "System Preferences → Profiles". | adversarial | 0.68 |
| 7 | `tests/*`, `nginx.conf` | A couple of comments narrate the change ("the #169 fix") rather than the invariant. | project-standards | 0.67 |
| 8 | `tests/test-lib-ca.sh` | No no-debris-on-success test for the mobileconfig helper. | testing, maintainability | 0.65 |

**Fixes (this PR):** reword the iOS fallback to match real raw-cert behaviour; add the pre-Ventura parenthetical; light comment touch-ups; add the no-debris test.

### Verified safe (not findings)
- The profile carries only the CA cert (DER), never a key (helper reads `HALOS_CA_ACTIVE_CRT`). Mode 0644 correct for a public trust anchor.
- `display_host` XML-injection: `halos_canonical_hostname` is constrained by the DNS-label regex (`[A-Za-z0-9.-]`), so `<>&"'` can't reach the plist (security-reviewer confirmed). No escaping added (source-constrained).
- nginx exact-match precedence, `disable_symlinks on`, `limit_except`, no path traversal (alias is a constant). Inline-no-attachment is the correct #169 fix (verified: generated profile passes `plutil -lint` and round-trips the DER on macOS).

### Coverage / residual
- First-boot ordering: `halos-manage-certs` runs `Before=` the container stack, so the profile exists before the sidecar serves; the healthcheck still probes `/ca/halos-ca.crt` only (same as before) — a missing profile is degraded, not unhealthy. Accepted.
- No `Cache-Control` on the profile (a rotation could serve stale via an intermediary) — same posture as the `.crt`; CA rotation is rare. Accepted.

**Verdict: Ready with fixes.** Addressing #1 (P1) and the P2/P3 items in-branch, then re-running the suite and re-verifying on device before merge.

mairas:
## Review findings addressed (amended into the commit)

**P1 — fleet collision (#1):** UUIDs and identifiers are now derived from the CA's SHA-256 fingerprint (`fi.halos.ca-trust.<fp12>` / per-fingerprint UUIDs) instead of fixed constants. Distinct devices → distinct profile identity, so installing a second device's profile no longer evicts the first's CA. Same-device reinstall still overwrites cleanly; a CA rotation now produces a new profile (acceptable — re-trust is already required on rotation). Verified on macOS: two CAs → two distinct `PayloadUUID`/`PayloadIdentifier`, `plutil -lint` OK, DER still round-trips. New test `test_publish_mobileconfig_uuid_is_per_ca` locks the invariant.

**P2:**
- #2 `base64` now captured directly (no pipe) so its exit status is the one tested; truncated output can't slip past, and `der_tmp` is cleaned on every path.
- #3 added `test_publish_mobileconfig_refuses_directory_at_output`.
- #4 tightened the integration assertion to `HaLOS Device CA (fixture.test)` and fixed the stale `device.local` comment.

**P3:**
- #5 iOS "Advanced" fallback reworded — no longer claims a raw `.crt` shows *Profile Downloaded*; describes the actual allow-certificate dialog.
- #6 added the pre-Ventura "System Preferences → Profiles" note.
- #7 reworded the two change-narration comments to describe the invariant.
- #8 added `test_publish_mobileconfig_no_new_debris_on_success`.

**Not changed (verified safe):** `display_host` XML escaping — `halos_canonical_hostname` is DNS-regex-constrained, so `<>&"'` can't reach the plist; no escaping added (source-constrained, YAGNI).

Suite: lib-ca 87, manage-certs 21, endpoint 15 (Docker), pytest 18 — all green. Next: deploy the CI build from trixie-unstable to the test device and verify the served MIME + profile before merge.

mairas:
## Correction: macOS uses the raw `.crt`, not the profile (device-tested)

Testing the profile on a real Mac surfaced a bug in the original macOS rewiring. On macOS (Ventura+):

- A root delivered by a *manually-installed* configuration profile is **not** trusted for SSL automatically, and macOS has **no** *Certificate Trust Settings* toggle (that screen is iOS-only).
- The profile-delivered cert does **not** appear in Keychain Access (confirmed on the test Mac), so there is no UI to grant it SSL trust — a dead end. ([Apple Developer Forums 724327](https://developer.apple.com/forums/thread/724327).)

So on macOS the profile is strictly worse than the raw `.crt`: an extra Device Management install step *and* no reachable trust. The raw `.crt` → System keychain → *Always Trust* is the only working macOS path.

**Change:** the macOS section reverts to the raw-`.crt` + Keychain "Always Trust" flow (no profile button); iOS keeps `.mobileconfig` as primary, where it genuinely helps (Certificate Trust Settings toggle exists, and it avoids the Files-app routing in #169). `docs/CERTS.md` and the helper comments now state the iOS-only rationale. Landing-content tests are section-aware: `test_macos_uses_raw_crt_not_profile` asserts macOS offers no `.mobileconfig`; `test_ios_leads_with_mobileconfig` asserts iOS does.

Verified live on `halosdev.local` (devtest build): served macOS section links only `/ca/halos-ca.crt`; iOS section links the profile.

Issue #176, opened 2026-06-01
Device-identifying auto-CA (hostname in CN, filename, and /ca instructions)


mairas:
Part of epic #159 (effortless CA trust). Follows the merged `.mobileconfig` carrier (#173). The old `.deb` carrier (Unit 6) was dropped.

# Device-identifying auto-CA

**Target repo:** `halos-core-containers`. New unit on the open epic #159 (effortless CA trust). The `.mobileconfig` carrier merged as #173; the `.deb` carrier (old Unit 6) was dropped.

## Overview

Each HaLOS device generates its own auto-CA, but the CA's subject is the constant `/CN=HaLOS Device CA` and the published file is always `halos-ca.crt`. Across a fleet, every device's CA is therefore indistinguishable — in OS trust stores (Keychain Access, the iOS Profiles list, the Windows store) and in the Downloads folder. The pain is worst not at install but at **management**: an operator trying to delete a stale device's CA sees several identical `HaLOS Device CA` rows with no way to tell them apart.

This plan makes the auto-CA device-identifying by reflecting the device hostname/fqdn in (a) the CA's subject CN, (b) the downloaded filename, and (c) the `/ca` landing-page instructions. The CN change is gated by an **adoption** model so it never orphans an already-installed CA.

## Problem Frame

The auto-CA is deliberately long-lived (20 y) and is **not** regenerated on hostname change, because regenerating orphans every trust anchor an operator already installed. So we cannot simply bake the current hostname into the CN and re-issue on rename. The insight that unlocks this: regenerating the auto-CA is *free of orphan risk* until the CA has been downloaded at least once — before first download, nothing is installed anywhere to orphan. So the CN can be kept fresh by regenerating while **unadopted**, and frozen permanently at **first external download**.

The filename and the on-page instructions are simpler: the `/ca` page is served from the very device being accessed, so client-side JavaScript already knows the hostname via `window.location.hostname` and can set a device-specific download filename and name the device in the steps — with no server-side hostname injection and no URL-path change.

## Requirements Trace

- **R1.** The auto-CA subject CN reflects the device hostname (`HaLOS Device CA (<fqdn>)`), so devices are distinguishable in trust stores for both install and deletion. (see origin: #159)
- **R2.** The downloaded certificate filename reflects the hostname, so multiple devices' files are distinguishable in a Downloads folder.
- **R3.** The `/ca` instructions name the device and stay correct for **both** older certs (generic CN) and new certs (hostname CN).
[...]

Issue #178, opened 2026-06-01
Unit 2: Hostname CN + unadopted-refresh regeneration


mairas:
> Sub-issue of #176 (device-identifying auto-CA); part of epic #159. See #176 for the full design context — decisions, requirements R1–R8, the lifecycle/serving diagrams, risks. This issue carries just the unit spec.

**Goal:** New auto-CAs carry `HaLOS Device CA (<hostname>)`; while unadopted, a hostname change refreshes the CN by regenerating the CA; once adopted (or for pre-feature CAs) the CN is frozen.

**Requirements:** R1, R4 (policy half), R6, R7, R8.

**Dependencies:** #177 (the adoption sentinel and its initialization).

**Files:**
- Modify: `assets/lib-ca.sh` — `halos_ca_ensure_auto` accepts the hostname and uses it in the CN on creation, and initializes the sentinel to `pending`; `halos_ca_select_active` threads the hostname through to the auto branch only. Custom-CA branch unchanged.
- Modify: `assets/halos-manage-certs` — pass `HALOS_DOMAIN` into selection; add the CN-refresh gate (regenerate the auto-CA when the sentinel reads `pending` and the resolved hostname differs from the CN's embedded hostname). The regenerated CA must be the one every downstream consumer of the active CA sees in the same run — see the end-of-run consistency invariant in Approach.
- Test: `tests/test-lib-ca.sh` (CN embeds hostname on creation; sentinel initialized `pending`; custom CA untouched), `tests/test-halos-manage-certs.sh` (a `pending` CA whose hostname changed refreshes CN with all artifacts consistent at run end — see the end-of-run consistency scenario; an `adopted` CA is frozen; a bare-CN legacy CA is stamped `adopted` and frozen; self-heal/expiry regen still fires independent of adoption).

**Approach:**
- The CN-refresh gate is a new branch distinct from the expiry/skew path; both can lead to regeneration but for different reasons, and only the CN-refresh branch is adoption-gated (sentinel `pending`).
- The compare hostname comes from parsing the CN of the active auto-CA (a bare legacy CN yields no embedded hostname, which always "differs" — only reachable if the sentinel was reset to `pending`, so a reset legacy CA upgrades to a device-identifying CN).
- Regeneration reuses the existing delete-then-bootstrap behavior of `halos_ca_ensure_auto`; the leaf re-signs because the sentinel's CA-fingerprint component changes.
[...]

PR #181, opened 2026-06-01
Unit 1: ca-download busybox httpd + adoption sentinel

### LINKED ISSUE #177: Unit 1: Serving swap to busybox + CGI — single adoption sentinel, liveness healthcheck

**Goal:** Replace the nginx sidecar with `busybox httpd` serving the landing page statically and the cert/profile endpoints via CGI, plus a static `/healthz` liveness endpoint. The cert/profile CGIs flip the single adoption sentinel `pending`→`adopted` in place after sending the body; the cert-manager ensures the sentinel exists with a correct initial value. After this unit the sentinel reaches `adopted` on the first *completed* external download; the refresh gate that consumes it lands in #178.

**Requirements:** R4 (detection half), R5. (Serving model: Decision 8; sentinel: Decisions 2–3.)

**Dependencies:** None.

**Files:**
- Replace: `assets/ca-download/nginx.conf` → `assets/ca-download/httpd.conf` (busybox httpd config: `*.crt:/bin/sh` + `*.mobileconfig:/bin/sh` suffix rules, MIME types) plus **three** CGI scripts placed at the docroot public paths — cert (attachment), profile (**inline aspen, #169**), and the `/halos-ca.crt` 410. The `/ca`→`/ca/` redirect is busybox-native (no CGI). A static `/healthz` returns `ok` (liveness). The cert and profile CGIs end by flipping `${adoption}` `pending`→`adopted` **in place** (no-op if already `adopted`). Constraint: no *static* `.crt`/`.mobileconfig` files in the docroot (the suffix rule would execute them).
- Modify: `docker-compose.yml` — swap `image: nginx:1.27-alpine` → `busybox:1.37` (pin the digest); set the `httpd -f -p 80 -u <uid> -h <docroot>` command (drop privileges after binding :80, to the uid that owns the sentinel); mount `httpd.conf` and the CGI scripts read-only; bind-mount the **single** sentinel file `${AUTO_CA_DIR}/adoption` read-write (the file, not a directory), owned by `<uid>`; keep `/srv` and `/srv-landing` read-only; point the healthcheck at the static `/healthz` (liveness only — no longer the cert path). `init: true` already present.
- Modify: `assets/halos-manage-certs` / `assets/lib-ca.sh` — ensure the sentinel exists *before the container stack starts* with the correct initial value (new-version CA → `pending`; pre-existing CA, no sentinel → classify by CN: bare → `adopted`, new-format → `pending`); a read helper ("is the CA adopted?"). All writes **in place** (open/truncate/rewrite), an explicit exception to the `tmp`+`mv` convention (single-file bind mount; see Decision 3). Aux-failure semantics (WARN, continue) except that a missing sentinel must be created (the container depends on it existing). Raise `halos_ca_publish_public` failure to an ERROR-severity journal line (it is the only signal that the served cert is missing, now that the healthcheck no longer probes it — R5).
- Test: `tests/test-ca-download-endpoint.sh` (sentinel flipped to `adopted` only after a *completed* GET of cert + profile; profile served **inline** with the aspen type and **no** `Content-Disposition: attachment` — #169 guard; cert served as attachment; `/ca`→`/ca/` native 302 with a relative Location; `/halos-ca.crt` 410; `/healthz` returns 200 and never touches the sentinel; GET/HEAD only), `tests/test-halos-manage-certs.sh` / `tests/test-lib-ca.sh` (sentinel initialized `pending` for a new CA, `adopted` for a seeded bare-CN legacy CA; idempotent; writes are in place / inode preserved; a publish failure logs at ERROR).

**Approach:**
- The healthcheck is liveness-only (static `/healthz`); it never touches the cert or the sentinel, so it cannot trip adoption (Decision 4). A missing published cert is surfaced by `halos_ca_publish_public` logging at ERROR, not by the healthcheck.
- The flip is the CGI's last action, after the body is flushed, so a torn download (SIGPIPE) leaves the sentinel `pending`.
- One always-present file is the durable truth (Decision 2/3): no separate state mount, no promotion. In-place writes keep the container's single-file bind mount coherent (Decision 3); an unrecognized value reads as `adopted`.
- The sidecar can rewrite only the one sentinel file; it cannot create/delete in `AUTO_CA_DIR` and `ca.key` is never mounted.

**Patterns to follow:** the read-only-mount posture of the sidecar, the current `nginx.conf` as the behavioral spec the CGIs must match (especially the #169 inline-profile rationale), `set -e` cmd-subst guards in the CGI shell. (Note: the sentinel write deliberately departs from the repo's atomic `tmp`+`mv` convention — see Decision 3.)

**Test scenarios:**
- Happy path: a completed external GET of `/ca/halos-ca.crt` flips the sentinel to `adopted`; a `pending` sentinel before, `adopted` after.
- Happy path: a completed GET of `/ca/halos-ca.mobileconfig` also flips it, and the response is inline aspen with no attachment header (#169).
- Edge case: `/healthz` returns 200 (liveness) and leaves the sentinel unchanged.
- Edge case: a client that aborts mid-body (SIGPIPE to the CGI) leaves the sentinel `pending` — no adoption on a torn download.
- Edge case: a second download with the sentinel already `adopted` is a no-op.
- Init: a new-version CA creation leaves `adoption=pending`; a seeded bare-CN CA with no sentinel is stamped `adopted` on the next cert-manage run; the sentinel exists before the container starts.
- Error: a failed `halos_ca_publish_public` logs at ERROR and does not block the run.
- Integration: the cert-manager's in-place sentinel rewrite is visible to the running container (inode unchanged).

**Verification:** On a test device, fetching the cert externally flips `${AUTO_CA_DIR}/adoption` to `adopted` (and the running container and host agree on the contents); the sidecar healthcheck reports healthy via `/healthz`; iOS still shows the profile-install prompt from `/ca/halos-ca.mobileconfig` (#169 preserved).

mairas:
First unit of the device-identifying auto-CA work (#176). Swaps the `ca-download` sidecar from nginx to **busybox httpd + CGI** and introduces the **adoption sentinel** — the infrastructure later units use to keep the auto-CA's subject CN device-identifying without ever orphaning an installed trust anchor.

## Why busybox

The CN can be safely refreshed only until the CA has been downloaded once; after that it must freeze. Recording "has ever been downloaded" needs a deliberate per-request side effect, which stock `nginx:1.27-alpine` cannot express. busybox httpd runs the cert/profile endpoints as CGI shell scripts that emit the response and then flip a single `pending`→`adopted` sentinel. Net image footprint also drops (~50 MB nginx, used nowhere else, → ~4 MB busybox).

## What changed

- **`assets/ca-download/httpd.conf` + docroot `assets/ca-download/www/`** — `*.crt`/`*.mobileconfig` suffix rules route those paths through `/bin/sh`; three CGIs (cert as attachment, profile **inline** per #169, legacy `/halos-ca.crt` → 410). `/ca`→`/ca/` is busybox-native (302, relative Location). `/healthz` is a static liveness file. The landing page moved into `www/ca/index.html` (busybox maps URL→docroot strictly, no `alias`/rewrite).
- **`docker-compose.yml`** — digest-pinned `busybox:1.37`; `httpd` drops to `nobody` after binding :80; the single sentinel file is bind-mounted read-write (the file, not a directory — `ca.key` is never mounted); healthcheck is **liveness-only** (the old cert-path probe could only detect, not fix, a missing cert, and made Traefik 3.7 drop the container into a homarr fallthrough).
- **`assets/lib-ca.sh`** — `halos_ca_adoption_init` (classifies a sentinel-less CA by CN: bare legacy → `adopted`/frozen, device-CN or just-created → `pending`) and `halos_ca_is_adopted`; `halos_ca_ensure_auto` now exposes a just-created flag. Writes are in place (the bind mount is inode-pinned at container start).
- **`assets/halos-manage-certs`** — creates the sentinel *before* the container stack starts and chowns it to the sidecar uid; a failed public-CA publish now logs at **ERROR** (the only signal left, now that the healthcheck no longer probes the cert).

The adoption flip is the CGI's last action, after the body is flushed, so a torn download (SIGPIPE) does not adopt.

## Tests

- `tests/test-ca-download-cgi.sh` (new, Docker-free): method routing, torn-download → no-adopt, in-place inode preservation, 503 on missing cert, 410.
- `tests/test-ca-download-endpoint.sh` (rewritten for busybox): full URL layout, headers, native `/ca` redirect, inline-profile #169 guard, and host↔container adoption coherence.
- `tests/test-lib-ca.sh` / `tests/test-halos-manage-certs.sh`: sentinel init/classification, legacy bare-CN → adopted, preserve-on-rerun, ERROR-on-publish-failure.

All suites green locally (`./run test`): lib-ca 100, manage-certs 25, endpoint 23, cgi 9, hostnames 45, traefik 10, pytest 19.

## Notes

- **VERSION** not bumped: `VERSION` (0.4.3) is already ahead of the latest stable tag (v0.4.2+8), so the cycle is open and CI walks `+N`.
- **Fly-by:** the first commit removes the repo-local "Git Workflow Policy" section from `AGENTS.md` (duplicated workspace-level guidance; intentional, separate atomic commit).
- On-device iOS profile-install verification (#169) remains a manual check per the issue's verification block.

Refs #176
Closes #177


mairas:
## Tiered review (security · correctness · adversarial)

Ran three persona reviewers over the diff. Security found **no exploitable issues** (`ca.key` is never mounted; CGIs `printf` only literal headers and dispatch `REQUEST_METHOD` by exact match; the suffix rule only executes files in the read-only `/www` docroot, which holds no static `.crt`). Correctness/adversarial surfaced these, triaged below.

### Fixed in a412f41
- **Torn-download protection was over-claimed (high).** busybox httpd buffers CGI stdout, so a ~2–4 KB cert flushes in one buffer and the script adopts on *serve*, not on confirmed receipt — the SIGPIPE "no adopt on disconnect" guarantee only holds for payloads larger than the socket buffer. Corrected the CGI comment and the test's framing. This is safe: premature adoption only forgoes a future CN refresh, it never orphans (the model freezes the CN precisely to avoid orphaning).
- **Directory at the sentinel path aborted the cert-manager forever (high).** If a `compose up` beat the cert-manager, Docker creates a *directory* at the bind-mount source; `adoption_init` then hit `printf > <dir>` under `set -e` and the script could never recover. It now self-heals (removes the non-regular file and recreates). Test added.
- **Device-CN glob matched empty `()` (medium).** Tightened to require a non-empty hostname, so a malformed `HaLOS Device CA ()` fails safe to `adopted`. Test added.
- **Stale `docs/CERTS.md` landing path (low).** Fixed after the `landing/ → www/ca/` move.

### Accepted by design (not changed)
- **`halos_ca_is_adopted` has no caller yet.** Intentional — this unit ships the sentinel infrastructure; the CN-refresh gate that consumes it lands in Unit 2 (#178), as stated in the commit/PR body. The init guard preserves any `pending` value an upgrade leaves, so the hand-off is safe.
- **Liveness-only healthcheck "silences" a publish failure.** Deliberate (plan Decision 4 / R5): the former cert-presence probe could only *detect* a missing cert, not fix it, and made Traefik 3.7 drop the container into a homarr fallthrough. Making the publish abort the cert-manager would block the whole container stack over an auxiliary step — worse. The failure surfaces as a journal `ERROR`.
- **Concurrent-download race.** Both writers converge on `adopted`; a crash between truncate and write leaves an empty file, which `is_adopted` reads as adopted (fail-safe).
- **Endpoint test uses sentinel mode 0666 (no chown).** Cross-platform constraint (macOS dev can't `chown 65534`); the `-u 65534:65534` privilege drop is asserted via the rendered compose config instead.

All suites green locally: lib-ca 102, manage-certs 25, endpoint 23, cgi 9, hostnames 45, traefik 10, pytest 19.

PR #182, opened 2026-06-01
Unit 2: device-identifying auto-CA CN + unadopted refresh

### LINKED ISSUE #178: Unit 2: Hostname CN + unadopted-refresh regeneration

mairas:
Second unit of the device-identifying auto-CA work (#176), building on the adoption sentinel from #177. Makes the auto-CA's subject CN device-identifying and keeps it fresh until first download, then freezes it.

## Behavior

- New auto-CAs carry `CN = HaLOS Device CA (<hostname>)` (matching the `.mobileconfig` display string), so fleet devices are distinguishable in Keychain Access / the iOS Profiles list / the Windows store — for both install and **deletion** (the original pain point).
- While the CA is **unadopted** (sentinel `pending`, never downloaded), a hostname change regenerates the CA with the new CN. Regenerating before first download is orphan-free — nothing is installed anywhere yet.
- Once **adopted** (or for a pre-feature bare-CN CA, stamped `adopted` on first sight in #177), the CN is **frozen**. A later rename leaves a cosmetically-stale CN but never orphans the installed anchor; TLS keeps working via the leaf SANs.

## How

- `assets/lib-ca.sh`: `halos_ca_ensure_auto` takes an optional hostname and builds the device CN (bare CN when omitted — the one-arg form still works for back-compat); `halos_ca_select_active` threads it to the auto branch only (custom CAs untouched, **R7**); `halos_ca_cn_hostname` extracts the embedded hostname (empty for a bare/non-device CN, which always "differs" — so a reset or Unit-1-era `pending` bare-CN CA upgrades to a device CN on its next refresh).
- `assets/halos-manage-certs`: passes `HALOS_DOMAIN` into selection and adds the **CN-refresh gate** — regenerate when the auto-CA is unadopted and its CN host differs from the resolved hostname. The gate runs **before any consumer of the active CA** (public publish, `.mobileconfig`, leaf re-sign), so every run-end artifact reflects the one regenerated CA. The changed fingerprint re-signs the leaf via the existing `.domain` sentinel.

The CN-refresh is a distinct trigger from the expiry/clock-skew self-heal regen, which stays **un-gated** by adoption and now also picks up the current hostname in its new CN (**R6**).

## End-of-run consistency

The key correctness contract: after any run, the served CA (`PUBLIC_CA_DIR/halos-ca.crt`), the CA in the `.mobileconfig`, the leaf's issuer, and the sentinel's fingerprint all reflect the **same** CA. Tested directly: after an unadopted-rename run the leaf verifies against the *published* `.crt`, the published CA fingerprint equals the auto-CA's, and the profile names the new host.

## Tests

- `tests/test-lib-ca.sh`: CN embedding, CN round-trip (no churn), `cn_hostname` for bare/custom CNs, hostname threading, custom-CN-untouched. (109 total)
- `tests/test-halos-manage-certs.sh`: first-boot CN names the device; unadopted rename refreshes CN + re-signs leaf + end-of-run consistency; no-churn when unchanged; adopted rename frozen; legacy bare-CN frozen; expiry self-heal picks up the hostname even when adopted; custom CA untouched. (33 total)

All suites green (`./run test`): lib-ca 109, manage-certs 33, endpoint 23, cgi 9, hostnames 45, traefik 10, pytest 19.

## Notes
- VERSION not bumped (cycle already open; CI walks `+N`).
- Requirements covered: R1, R4 (policy half), R6, R7, R8 (escape hatch = reset the sentinel to `pending`, documented for Unit 4).

Refs #176
Closes #178

### DIFF
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -98,7 +98,7 @@ mkdir -p "${CERTS_DIR}"
 # silently falling back — operators who installed a custom CA can SSH to
 # diagnose, and silent fallback would orphan the trust anchor they
 # distributed.
-halos_ca_select_active "${CUSTOM_CA_DIR}" "${AUTO_CA_DIR}" "${SERVING_CA}"
+halos_ca_select_active "${CUSTOM_CA_DIR}" "${AUTO_CA_DIR}" "${SERVING_CA}" "${HALOS_DOMAIN}"
 CA_CRT="${HALOS_CA_ACTIVE_CRT}"
 CA_KEY="${HALOS_CA_ACTIVE_KEY}"
 echo "Active CA: ${CA_CRT} (mode=${HALOS_CA_ACTIVE_MODE})"
@@ -112,6 +112,36 @@ echo "Active CA: ${CA_CRT} (mode=${HALOS_CA_ACTIVE_MODE})"
 # tests run unprivileged.
 halos_ca_adoption_init "${ADOPTION_FILE}" "${AUTO_CA_DIR}/ca.crt" "${HALOS_CA_AUTO_CREATED:-0}"
 chown "${HALOS_CA_DOWNLOAD_UID}:${HALOS_CA_DOWNLOAD_UID}" "${ADOPTION_FILE}" 2>/dev/null || true
+
+# CN-refresh gate: keep the auto-CA's subject CN tracking the current hostname
+# while it is still unadopted (never downloaded), then freeze it. This is a
+# distinct trigger from the expiry/clock-skew self-heal regen inside
+# halos_ca_ensure_auto, which still fires regardless of adoption. Auto mode
+# only — operators own custom-CA CNs. The regen runs here, before any consumer
+# of the active CA (public publish, .mobileconfig, leaf re-sign), so every
+# run-end artifact reflects the one regenerated CA; a CA whose fingerprint
+# changes re-signs the leaf via the existing sentinel mismatch.
+# Skip when the hostname is unresolved: an empty domain cannot produce a stable
+# CN, and refreshing toward it would churn a new CA on every flap.
+if [ "${HALOS_CA_ACTIVE_MODE}" = "auto" ] && [ -n "${HALOS_DOMAIN}" ] \
+   && ! halos_ca_is_adopted "${ADOPTION_FILE}"; then
+    CURRENT_CN_HOST="$(halos_ca_cn_hostname "${CA_CRT}")"
+    if [ "${CURRENT_CN_HOST}" != "${HALOS_DOMAIN}" ]; then
+        # Re-check adoption immediately before the destructive regen: a download
+        # can flip the sentinel between the gate's first check and here, and
+        # regenerating an adopted CA would orphan the just-installed anchor.
+        if halos_ca_is_adopted "${ADOPTION_FILE}"; then
+            echo "Adoption raced ahead of the CN refresh; leaving the CA frozen."
+        else
+            echo "Auto-CA unadopted and its CN host (${CURRENT_CN_HOST:-none}) differs from ${HALOS_DOMAIN}; refreshing CA CN..."
+            rm -f "${AUTO_CA_DIR}/ca.crt" "${AUTO_CA_DIR}/ca.key"
+            if ! halos_ca_ensure_auto "${AUTO_CA_DIR}" "${HALOS_DOMAIN}"; then
+                echo "halos-manage-certs: CN-refresh regeneration failed; cert generation cannot continue" >&2
+                exit 1
+            fi
+        fi
+    fi
+fi
 
 # Publish the active CA's PEM bytes into the public dir the ca-download
 # sidecar bind-mounts. Refreshed unconditionally so the published bytes
--- a/assets/lib-ca.sh
+++ b/assets/lib-ca.sh
@@ -232,8 +232,14 @@ halos_ca_leaf_needs_renewal() {
 # healthy CA was reused unchanged. Read by halos_ca_adoption_init to decide a
 # sentinel-less CA's initial state — a CA born this run starts "pending"
 # (refresh-eligible), a pre-existing one is classified by its CN.
+#
+# <hostname> (optional) makes the subject CN device-identifying:
+# "HaLOS Device CA (<hostname>)" so devices are distinguishable in trust stores.
+# Omitted/empty yields the bare legacy CN; callers that don't care (tests) can
+# keep the one-arg form. Only affects a CA this call generates — an existing
+# healthy CA is reused with whatever CN it already has.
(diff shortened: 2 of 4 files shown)


mairas:
## CI fix + tiered review

### CI failure (root-caused, fixed)
`test_pending_rename_refreshes_cn` failed on the runner but passed on macOS. Root cause: the test detected CA regeneration by **inode change**, but the CN-refresh gate deletes `ca.crt` before regenerating, and Linux ext4/overlay **reuses the just-freed inode number** — so the inode looked unchanged (false negative). macOS APFS doesn't reuse it, hiding the bug locally. Reproduced in a `debian:trixie` container (OpenSSL 3.5). Fixed by detecting regeneration via **fingerprint** (content-based, filesystem-agnostic) across the four CA-regen tests. All bash suites now pass in trixie (lib-ca 109, manage-certs 33, cgi 9).

### Review (correctness + adversarial)
Two real findings, both fixed:
- **Churn on unstable hostname resolution (high).** If `HALOS_DOMAIN` flaps empty↔value, the gate would regenerate an RSA-4096 CA every timer run. Added an empty-domain guard so the gate is skipped when the hostname is unresolved.
- **Download-flip race (orphan, net-new).** The daily timer could pass the `is_adopted` check, then a download adopts, then the timer's `rm` deletes the just-installed CA. Added an adoption re-check immediately before the destructive regen, narrowing the window to near-zero.
- **Minor:** `_halos_ca_subject_cn` masked openssl's exit status behind `sed`; now captured separately so failures propagate.

### Accepted by design
- A Unit-1 device that never downloaded its bare-CN CA is stamped `adopted` on the first Unit-2 run (so it keeps the bare CN, losing the device-CN benefit on that device). This is the deliberate migration-safety trade-off — a sentinel-less bare-CN CA *might* have been installed, so it's frozen rather than risk an orphan. The escape hatch (reset the sentinel to `pending`) recovers it; documented in Unit 4.
- The expiry self-heal regen stays un-gated by adoption (R6) and now picks up the current hostname — intended.

The end-of-run consistency is verified directly: after an unadopted-rename run the leaf verifies against the *published* `.crt`, the published CA fingerprint equals the auto-CA's, and the profile names the new host.

PR #183, opened 2026-06-01
Unit 3: device-specific download filename + on-page device name

### LINKED ISSUE #179: Unit 3: Device-specific download filename and on-page device name

**Goal:** The `/ca` page offers the cert as `halos-ca-<hostname>.crt` and names the device in the install steps, correct for both old and new CNs.

**Requirements:** R2, R3.

**Dependencies:** None (client-side; independent of #177–#178).

**Files:**
- Modify: `assets/ca-download/landing/index.html` — extend the inline JS to read `window.location.hostname`, set the cert download anchor(s)' `download` attribute to a sanitized `halos-ca-<hostname>.crt`, and fill the device name into the intro and the per-OS steps. Keep the stable `HaLOS Device CA` prefix in trust-store-search wording.
- Test: `tests/test_landing_content.py` (download anchors carry a `download` attribute / placeholder; the cert link path is still `/ca/halos-ca.crt`; the "search for HaLOS Device CA" wording is present so old CNs still match).

**Approach:**
- Pure client-side; no server change. The URL path stays `/ca/halos-ca.crt`; only the suggested save name changes.
- Because the page can't know whether a given device's cert has the new or old CN, instructions anchor on the stable prefix and present the hostname as context ("…it will show your device name, `<hostname>`, in parentheses").
- Sanitize the hostname for filesystem-safe filenames; fall back gracefully when accessed by raw IP.

**Patterns to follow:** the existing inline UA-detection script in `index.html`; the section-aware content guards already in `tests/test_landing_content.py`.

**Test scenarios:**
- Happy path: the macOS/Windows/Android/Linux download anchors target `/ca/halos-ca.crt` and carry a `download` attribute (static guard for the attribute's presence; the hostname value is set at runtime).
- Edge case: instructions retain the `HaLOS Device CA` search term so a generic-CN cert is still findable.
- Edge case (manual/JS): accessed via `https://<ip>/ca/`, the filename degrades to a safe default rather than embedding `:` or brackets.

**Test expectation: limited** — most behavior is runtime DOM from `location.hostname`; static guards cover the structural invariants, the rest is manual device verification.

**Verification:** Opening `/ca/` on the device, the download saves as `halos-ca-<hostname>.crt` and the steps name the device; the cert URL is unchanged.

mairas:
Third unit of the device-identifying auto-CA work (#176). Makes the `/ca` landing page name the device, so several HaLOS devices' certificates are distinguishable in a Downloads folder (**R2**) and in the OS trust store (**R3**). Independent of Units 1–2 — pure client-side, no server change.

## What changed (`assets/ca-download/www/ca/index.html`)
- The page is served *from the device being accessed*, so `window.location.hostname` is the device hostname — no server-side injection.
- The three cert-download anchors get a `data-cert-download` hook (and keep a static `download="halos-ca.crt"` fallback for JS-off). The inline JS sets each to a sanitized **`halos-ca-<hostname>.crt`**. The URL path stays `/ca/halos-ca.crt`; only the saved filename changes.
- The install steps name the device as *context* for the trust-store entry, while keeping the stable **`HaLOS Device CA`** prefix searchable — so the copy is correct for both pre-feature bare-CN certs and the new `HaLOS Device CA (<hostname>)` form.
- Sanitization: `[^A-Za-z0-9._-] → -` (dots/hyphens pass for DNS names and IPv4; IPv6 colons collapse). The on-page device *name* falls back to "this device" for raw-IP access, since the CN embeds a DNS name, never an IP.

## Verification
Browser-checked both branches:
- via `localhost`: anchors save as `halos-ca-localhost.crt`, steps read "localhost".
- via `127.0.0.1`: filename `halos-ca-127.0.0.1.crt`, device name degrades to "this device".

## Tests
`tests/test_landing_content.py` static guards: the download hook + attribute on every cert anchor, the unchanged `/ca/halos-ca.crt` URL, the searchable `HaLOS Device CA` prefix, the device-name placeholder, the filename-building JS, and the IP fallback. (21 pass.) Runtime DOM behavior is browser-verified above (the issue scopes static tests as "limited").

VERSION not bumped (cycle already open).

Refs #176
Closes #179

### DIFF
--- a/assets/ca-download/www/ca/index.html
+++ b/assets/ca-download/www/ca/index.html
@@ -109,9 +109,17 @@
       </p>
 
       <p>
-        <a class="download" href="/ca/halos-ca.crt" download="halos-ca.crt"
+        <a class="download" href="/ca/halos-ca.crt" download="halos-ca.crt" data-cert-download
           >Download the HaLOS CA certificate</a
         >
+      </p>
+
+      <p class="caveat">
+        In your operating system's trust store this CA appears as
+        <strong>HaLOS Device CA</strong>. Newer devices also show this device's
+        hostname, <code data-device-host>this device</code>, in parentheses —
+        handy when you manage several HaLOS devices and need to tell their
+        certificates apart (or remove a stale one).
       </p>
 
       <div class="note">
@@ -125,9 +133,9 @@
       <details id="macos">
         <summary>macOS</summary>
         <ol>
-          <li>Download <a href="/ca/halos-ca.crt" download="halos-ca.crt"><code>halos-ca.crt</code></a> (the button at the top does the same).</li>
+          <li>Download <a href="/ca/halos-ca.crt" download="halos-ca.crt" data-cert-download><code>halos-ca.crt</code></a> (the button at the top does the same).</li>
           <li>Double-click the downloaded file. Keychain Access opens and adds the certificate to the <strong>System</strong> keychain (choose System if it asks which keychain); enter an admin password if prompted.</li>
-          <li>It's installed but not yet trusted. In Keychain Access, select the <strong>System</strong> keychain, find <em>HaLOS Device CA</em> under <em>Certificates</em>, and double-click it.</li>
+          <li>It's installed but not yet trusted. In Keychain Access, select the <strong>System</strong> keychain, find <em>HaLOS Device CA</em> under <em>Certificates</em> (on newer devices it reads <em>HaLOS Device CA (<span data-device-host>this device</span>)</em>), and double-click it.</li>
           <li>Expand <strong>Trust</strong> and set <em>Secure Sockets Layer (SSL)</em> (or <em>When using this certificate</em>) to <strong>Always Trust</strong>. Close the window and authenticate. <span class="caveat">macOS keeps install and trust separate — without this step the warnings stay.</span></li>
           <li>Quit and reopen your browser.</li>
         </ol>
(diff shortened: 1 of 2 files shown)


mairas:
## Post-merge review (correctness + client-side input safety) — Unit 3 (#179)

Ran a correctness/security pass on the landing-page diff. **No exploitable issues and no behavioral bugs** — the substantive concerns checked out clean:

- **No XSS / no path traversal.** The `download` filename is set via `setAttribute` (an opaque save-name hint, not HTML) and the device name via `textContent` (never interprets HTML); the sanitizer strips everything but `[A-Za-z0-9._-]` (removing `/`, `\`, null, shell metachars). Browsers further sanitize the `download` hint.
- **IP detection is correct** for every form `window.location.hostname` returns (IPv4 anchored regex; IPv6 — incl. IPv4-mapped `::ffff:…` — caught by the colon check, since `location.hostname` returns IPv6 without brackets).
- **No-JS degradation works:** every cert anchor keeps a static `download="halos-ca.crt"` and every device-name slot reads "this device".

Two notes:

1. **(low, cosmetic)** A hostname with a trailing dot (FQDN form some browsers keep, e.g. `halos.local.`) yields `halos-ca-halos.local..crt`; a lone `.` yields `halos-ca-..crt`. Valid filenames everywhere, just ugly. One-line fix if we care: strip leading/trailing `[._-]` from the sanitized host before building the name.
2. **(verified — NOT a bug)** A reviewer flagged the test's `re.findall(r"<a\b[^>]*>", HTML)` as missing the multi-line primary download button (no `re.DOTALL`). I checked: Python's negated class `[^>]` matches newlines, so all **3** cert anchors — including the multi-line button — are found and validated. No change needed.

Net: clean; the cosmetic trailing-dot nit is optional.

PR #185, opened 2026-06-01
docs(certs): fix adoption sentinel path + download-header accuracy

mairas:
Fixes the valid review findings from #184 (device-identifying CA docs). Docs only.

## F1 — wrong adoption sentinel path (was breaking)
`docs/CERTS.md` gave the sentinel as `.../halos-core-containers/certs/ca/adoption` in both the display block and the escape-hatch `tee` command. The real path is `.../halos-core-containers/**data/halos-core-containers/**certs/ca/adoption` — `CONTAINER_DATA_ROOT` (from `env.defaults`) already ends in `/data`, and `AUTO_CA_DIR=${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/ca`. As written, the escape-hatch command wrote `pending` to a nonexistent path and silently no-op'd. Now matches the other (correct) paths in the same doc.

## F2 — Content-Disposition table was misleading
The table showed `filename="halos-ca-<host>.crt"`, but the CGI sends the **static** `filename="halos-ca.crt"`; the device-specific name is applied client-side via the `download` attribute. The table now shows the real header, with the client-side rename explained in prose (so `curl -I` matches the doc).

## F3 / F4 — wording
- Note the escape hatch is a **no-op** when the CA already names the current host (the gate fires only on a name mismatch).
- Sharpen the busybox-CGI solutions entry's premature-firing wording (it read ambiguously about whether SIGPIPE can abort the side effect for small payloads).

Verified the corrected path against `env.defaults` + `halos-manage-certs` and the pre-existing correct paths in CERTS.md. The #179/Unit-3 review (PR #183) found only an optional cosmetic nit — not included here.

Refs #176

### DIFF
--- a/docs/CERTS.md
+++ b/docs/CERTS.md
@@ -162,7 +162,7 @@ CA is **unadopted**, and frozen permanently at **first download**.
 
 Adoption is one bit of state in a single file:
 
-    /var/lib/container-apps/halos-core-containers/certs/ca/adoption   # "pending" | "adopted"
+    /var/lib/container-apps/halos-core-containers/data/halos-core-containers/certs/ca/adoption   # "pending" | "adopted"
 
 - **`pending`** — never downloaded. On each `halos-manage-certs` run, if the
   resolved hostname differs from the CA's embedded name, the auto-CA is
@@ -199,7 +199,7 @@ new one** — reset the sentinel to `pending` and re-run cert management:
 
 ```
 echo -n pending | sudo tee \
-  /var/lib/container-apps/halos-core-containers/certs/ca/adoption
+  /var/lib/container-apps/halos-core-containers/data/halos-core-containers/certs/ca/adoption
 sudo systemctl start halos-manage-certs.service
 ```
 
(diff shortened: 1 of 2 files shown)


mairas:
## Review — all four fixes verified correct and complete

Independent pass (traced paths/headers against the code, not just the diff):

- **F1 path:** the chain `CONTAINER_DATA_ROOT=${LIB_DIR}/data` → `AUTO_CA_DIR=${CONTAINER_DATA_ROOT}/${PACKAGE_NAME}/certs/ca` resolves to exactly the corrected path, and it matches the `docker-compose.yml` sentinel bind-mount and the other `data/halos-core-containers/` paths already in CERTS.md. The escape-hatch command now works end to end.
- **F2 table:** the CGI emits exactly `attachment; filename="halos-ca.crt"`; the table now matches, and the prose correctly attributes the device-specific name to the client-side `download` attribute.
- **F3 caveat:** matches the gate (`CURRENT_CN_HOST != HALOS_DOMAIN`) — a reset is genuinely a no-op when the name already matches.
- **F4 wording:** the reworded premature-firing note matches the CGI's buffered-stdout behavior.

No markdown breakage, no new stale paths or claims. Docs-only; version checks pass.

Ready to merge on your call.

PR #186, opened 2026-06-01
fix(ca-download): device-specific cert filename must be server-side

mairas:
## Bug
The device-specific download filename from #179 (`halos-ca-<host>.crt`) never worked. The cert endpoint serves `Content-Disposition: attachment; filename="halos-ca.crt"`, and **a server-sent `Content-Disposition` filename overrides an HTML `download` attribute** (documented same-origin browser behavior). So the page's client-side `download` value was always ignored and browsers saved `halos-ca.crt`.

Caught on the test device after a sentinel reset: clicking download on `halosdev.local/ca` saved `halos-ca.crt`, not `halos-ca-halosdev.local.crt`.

The original plan's "client-side is enough / no server-side injection needed" was wrong, and my #179 Playwright check only asserted the `download` *attribute value*, not the effective saved name with the server header present — that's the verification gap.

## Fix
Build the filename **in the cert CGI** from the request `Host` header:
- strip the port, sanitize to `[A-Za-z0-9._-]`, drop CR/LF, trim leading/trailing separators;
- the sanitization doubles as a header-injection guard (a crafted `Host` can't inject a header);
- empty/unresolved Host → bare `halos-ca.crt`; raw-IP access → `halos-ca-<ip>.crt`.

This is authoritative for browsers and `curl -OJ`. The landing JS no longer sets a download name (it would be overridden anyway) — it keeps only the device-name text fill; the now-dead `data-cert-download` hooks are removed.

## Tests
- `test-ca-download-cgi.sh`: device filename from Host, port strip, no-Host fallback, malicious-Host sanitization, and CRLF-injection guard.
- `test-ca-download-endpoint.sh`: asserts the real busybox server emits `filename="halos-ca-127.0.0.1.crt"` from the curl `Host` header (proves `HTTP_HOST` reaches the CGI).
- `test_landing_content.py`: dropped the client-side-filename assertions.

## Docs
`docs/CERTS.md`: the filename now correctly documented as server-side (Host header), with the download-attribute-override rationale — corrects the earlier client-side claim (incl. the one from #185).

All suites green (`./run test`): cgi 14, endpoint 23, lib-ca 109, manage-certs 33, hostnames 45, traefik 10, pytest 24. VERSION unchanged (cycle open).

Refs #176, #179

---
**Update:** also makes the on-page install instructions show the device-specific filename (every per-OS reference + the Linux `cp` command), and hardens the CGI host parsing (bracket-aware IPv6 port strip, 64-char length cap). Verified in a browser that the shown name equals the server's actual `Content-Disposition` filename.

### DIFF
--- a/assets/ca-download/www/ca/halos-ca.crt
+++ b/assets/ca-download/www/ca/halos-ca.crt
@@ -19,12 +19,37 @@ set -u
 CA_CRT="${HALOS_CA_PUBLIC_CRT:-/srv/halos-ca.crt}"
 ADOPTION="${HALOS_ADOPTION_FILE:-/adoption}"
 
+# Device-specific saved filename, derived from the Host the client used, so
+# several devices' certs are distinguishable in a Downloads folder. This MUST be
+# set server-side: a Content-Disposition filename from the server overrides an
+# HTML `download` attribute, so a client-side name never takes effect. The Host
+# is client-controlled, so sanitize to [A-Za-z0-9._-] (also trimming leading/
+# trailing separators) — this both yields a safe filename and blocks header
+# injection. Empty/again-unresolved Host falls back to the bare name.
+_host="${HTTP_HOST:-}"
+# Strip the port: a bracketed IPv6 host ([::1] or [::1]:443) keeps the address
+# inside the brackets; otherwise host[:port] drops at the first colon. (A bare,
+# unbracketed IPv6 Host is malformed per RFC 3986 and not handled specially.)
+case "$_host" in
+    \[*\]*) _host="${_host#\[}"; _host="${_host%%\]*}" ;;
+    *)      _host="${_host%%:*}" ;;
+esac
+# Sanitize (the Host is client-controlled), cap the length so the result stays a
+# valid <=255-byte filename component, then trim leading/trailing separators.
+_safe="$(printf '%s' "$_host" | tr -d '\r\n' | sed 's/[^A-Za-z0-9._-]/-/g')"
+_safe="$(printf '%.64s' "$_safe" | sed -e 's/^[._-]*//' -e 's/[._-]*$//')"
+if [ -n "$_safe" ]; then
+    CERT_FILENAME="halos-ca-${_safe}.crt"
+else
+    CERT_FILENAME="halos-ca.crt"
+fi
+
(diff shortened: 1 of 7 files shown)


mairas:
## Review + scope update

### Adversarial review — security core is solid
Verified with real bytes: **CRLF/header-injection is blocked** (CR/LF stripped before sanitizing; `"`/`/` collapse to `-`; the value only ever goes in a header, never a path) and **path traversal is impossible** (`../` → `etc`, and it's never used as a filesystem path). Format string is `%s`, not user-controlled.

### Review findings fixed
- **IPv6 Host mangled.** `${HTTP_HOST%%:*}` turned `[::1]:443` into `[` (→ generic fallback) and `[2001:db8::1]` into `2001`. Now bracket-aware: a bracketed IPv6 keeps the address (`[2001:db8::1]:443` → `halos-ca-2001-db8--1.crt`).
- **No length cap.** A long Host produced a >255-byte filename that the OS would truncate (possibly masking `.crt`). Capped to 64 chars.
- **Dead markup.** Removed leftover `data-cert-download` hooks.
- Strengthened the CRLF test to assert the exact `Content-Disposition` value; added IPv6, length-cap, and HEAD-with-Host tests.

### New requirement folded in (no separate issue)
Per your note — the on-page instructions referenced `halos-ca.crt` while the download now saves `halos-ca-<host>.crt`, so "tap the downloaded `halos-ca.crt`" and the Linux `cp` command named the wrong file. I templated **every** visible reference (the per-OS `<code>` mentions and both filenames in the `cp` command) and the JS fills them with the same derived name. I folded this into this PR rather than opening an issue because the feature would otherwise ship an inconsistency it introduced itself.

### Verified the *effect*, not a proxy
In a real browser against the busybox server: all 11 on-page filename references show `halos-ca-127.0.0.1.crt`, and a live `HEAD` of the cert endpoint returns `Content-Disposition: filename="halos-ca-127.0.0.1.crt"` — **shown name == served name**. (The original bug slipped through because it only checked the `download` *attribute*, not the saved file — captured as a `docs/solutions/` entry and a global verification note.)

All suites green: cgi 17, endpoint 23, lib-ca 109, manage-certs 33, hostnames 45, traefik 10, pytest 25.

PR #187, opened 2026-06-01
fix(ca-download): cert download filename from the CA's CN

mairas:
Follow-up to #186. The download filename should reflect the **cert's identity (its CN)** — the name the trust store shows — not the `Host` the page was reached by.

## Why #186's Host-based name was wrong
`Host` is only a proxy for the cert, and they diverge exactly where it matters:
- **bare-CN / legacy CA** → Host gives `halos-ca-halosdev.local.crt`, but the cert is the generic `HaLOS Device CA`; it should be `halos-ca.crt`.
- **adopted-then-renamed device** → CN is frozen at the old name, Host has the new one; the file should match the cert.
- **accessed by IP / alias** → same cert, different Host → different filenames, making one cert look like several.

## Fix (CN-based)
- `lib-ca.sh` — `halos_ca_download_filename` derives `halos-ca-<host>.crt` from the CA's CN (generic `halos-ca.crt` for a bare-legacy or operator-custom CA).
- `halos-manage-certs` — publishes that name to `${PUBLIC_CA_DIR}/download-filename` (the busybox sidecar has no `openssl` to read the CN itself).
- cert CGI — reads the name file for `Content-Disposition`, validated to a safe `halos-ca[-<safe>].crt` shape with a generic fallback; the `Host` logic is gone.
- landing JS — learns the name **authoritatively** via a `HEAD` of the cert endpoint and fills both the filename references and the device-name context from it. No client-side re-derivation that could drift from the server.

## Verification
- Browser, served by busybox with a CN-derived name file: instructions show `halos-ca-halosdev.local.crt` and device name `halosdev.local`, equal to the server's `Content-Disposition` (`HEAD`).
- Docker endpoint test asserts the CGI emits the published name; lib-ca tests cover device/bare/custom CN → filename; manage-certs tests assert the published `download-filename`; CGI tests cover the name-file read, generic fallback, and rejection of a tampered/unsafe name.

All suites green: lib-ca 112, manage-certs 35, cgi 13, endpoint 23, hostnames 45, traefik 10, pytest 25. VERSION unchanged (cycle open).

Refs #176, #179

### DIFF
--- a/assets/ca-download/www/ca/halos-ca.crt
+++ b/assets/ca-download/www/ca/halos-ca.crt
@@ -19,29 +19,26 @@ set -u
 CA_CRT="${HALOS_CA_PUBLIC_CRT:-/srv/halos-ca.crt}"
 ADOPTION="${HALOS_ADOPTION_FILE:-/adoption}"
 
-# Device-specific saved filename, derived from the Host the client used, so
-# several devices' certs are distinguishable in a Downloads folder. This MUST be
-# set server-side: a Content-Disposition filename from the server overrides an
-# HTML `download` attribute, so a client-side name never takes effect. The Host
-# is client-controlled, so sanitize to [A-Za-z0-9._-] (also trimming leading/
-# trailing separators) — this both yields a safe filename and blocks header
-# injection. Empty/again-unresolved Host falls back to the bare name.
-_host="${HTTP_HOST:-}"
-# Strip the port: a bracketed IPv6 host ([::1] or [::1]:443) keeps the address
-# inside the brackets; otherwise host[:port] drops at the first colon. (A bare,
-# unbracketed IPv6 Host is malformed per RFC 3986 and not handled specially.)
-case "$_host" in
-    \[*\]*) _host="${_host#\[}"; _host="${_host%%\]*}" ;;
-    *)      _host="${_host%%:*}" ;;
-esac
-# Sanitize (the Host is client-controlled), cap the length so the result stays a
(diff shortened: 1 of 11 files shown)


mairas:
## Adversarial review

**Security: clean.** Injection through the published name file is blocked — the CGI strips CR/LF then a `case` rejects anything outside `[A-Za-z0-9._-]`, and the fallback is a literal constant, so no quote-breakout or header split. `/srv` is mounted **read-only** (only the root cert-manager writes `download-filename`); the rw adoption mount is a separate single file. The page's `HEAD` fires before any user action and does **not** record adoption; on fetch failure / missing header the JS degrades to the static `halos-ca.crt` / "this device" fallbacks. The `filename="…"` and `^halos-ca-(.+)\.crt$` regexes can't mis-parse the values the CGI actually emits.

### Two findings, both fixed
- **(low) Non-atomic name-file write.** `download-filename` used a plain `>` truncate while `publish_public` writes atomically — a millisecond window on a CN refresh where the new cert could serve the old name. Now staged to `.new.$$` + `mv`, matching the repo pattern.
- **(info) Validation gap.** The CGI `case` accepted `halos-ca-.crt` (a dash-only device part the cert-manager can't produce and `/srv`-ro blocks anyway). Tightened the glob to require an alphanumeric right after the prefix; added a test.

All suites green: cgi 14, endpoint 23, manage-certs 35, lib-ca 112, hostnames 45, traefik 10, pytest 25.

PR #188, opened 2026-06-01
feat(certs): self-trust device CA in host trust store

### LINKED ISSUE #158: Device self-trusts its own CA via system trust store
## Problem

When a HaLOS device runs its own desktop (HALPI2 desktop variants etc.), the desktop browser doesn't trust the device's own CA out of the box. Opening the local dashboard on the device's own screen shows cert warnings until the operator manually installs the CA into the system trust store — a regression of the "expected default" state.

## Solution

Extend `halos-manage-certs` to additionally publish the active CA to `/usr/local/share/ca-certificates/halos-ca.crt` and run `update-ca-certificates` after the existing `halos_ca_publish_public` step. Fires on the same triggers the existing publish step does: boot, CA rotation, sentinel change, the 24h renewal-check timer.

## Scope

- `assets/halos-manage-certs` — system-trust publish step after `halos_ca_publish_public`
- `assets/lib-ca.sh` — possibly a `halos_ca_publish_system_trust` helper for symmetry
- Tests covering: first-install, CA rotation, idempotent re-run

## Out of scope

- Browser-profile prepopulation (Firefox NSS, Chromium policy file) — separate concern, would touch `halos-desktop-branding` if pursued
- Cross-device CA sharing — each device's CA is per-device by design

## Rationale

A device showing untrusted-cert warnings on its own screen is a quality bug independent from any user-facing install UX work for remote clients. Splitting from the user-facing CA-install umbrella to keep PR scope tight.

## Acceptance

After a fresh flash, the device's own Chromium/Brave session reaches the dashboard with no cert warning and no operator action.

mairas:
## Why

A HaLOS device that runs its own desktop (HALPI2 desktop variants) didn't trust its own CA. Opening the local dashboard on the device's own screen showed cert warnings until the operator manually installed the CA — a regression of the expected default state.

Closes #158.

## What

`halos-manage-certs` now installs the active CA into `/usr/local/share/ca-certificates/halos-ca.crt` and runs `update-ca-certificates`, right after the existing public-CA publish. The device's own browser and CLI tools then validate its HTTPS endpoints with no operator action.

New helper `halos_ca_publish_system_trust` in `lib-ca.sh` mirrors `halos_ca_publish_public`: atomic stage+mv, X.509-parse validation, preserve-the-old-copy-on-failure. It **skips the `update-ca-certificates` rebuild when the store already holds the active CA**, so the 24 h renewal timer never churns the host bundle — only a CA rotation re-triggers it. Aux-failure semantics match the sibling publish steps: a failure logs `WARNING` and never blocks cert management.

Fires on the same triggers as the existing publish step: boot, CA rotation, sentinel change, the 24 h renewal timer.

## Scope notes

- Runs on **every variant**, not gated to desktop — installing the device's own CA is benign (and useful for local `curl`) on headless devices too. The issue's scope listed no variant gating.
- Touches `docs/CERTS.md` (new "Host self-trust" subsection) to keep the operator reference current — slightly beyond the issue's file list, called out here.

## Tests

- `tests/test-lib-ca.sh` — 9 new unit tests for the helper: happy path, skip-when-current, rebuild-on-rotation, missing args (rc 2), missing/non-cert source (rc 1), preserve-on-failure, directory-at-output, no `.new` debris.
- `tests/test-halos-manage-certs.sh` — 3 integration tests via the env-stub seam (`HALOS_SYSTEM_TRUST_DIR`, `HALOS_UPDATE_CA_CMD`): first boot installs + rebuilds, idempotent rerun skips the rebuild, CA rotation re-installs + rebuilds.

Full suite green (lib-ca 121, halos-manage-certs 38, pytest 25, all sibling suites). shellcheck clean (only pre-existing info-level findings).

### DIFF
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -90,6 +90,11 @@ COCKPIT_WS_CERTS_DIR="${HALOS_COCKPIT_WS_CERTS_DIR:-/etc/cockpit/ws-certs.d}"
 COCKPIT_LEAF_OVERRIDE="${COCKPIT_WS_CERTS_DIR}/99-halos.cert"
 TRAEFIK_DYNAMIC_DIR="${HALOS_TRAEFIK_DYNAMIC_DIR:-/etc/halos/traefik-dynamic.d}"
 TRAEFIK_TLS_CONFIG="${TRAEFIK_DYNAMIC_DIR}/tls-default.yml"
+# Host CA trust store: the active CA is installed here so the device's own
+# desktop browser/CLI self-trust its HTTPS endpoints (#158). update_cmd is
+# overridable so tests can stub the bundle rebuild.
+SYSTEM_TRUST_DIR="${HALOS_SYSTEM_TRUST_DIR:-/usr/local/share/ca-certificates}"
+UPDATE_CA_CMD="${HALOS_UPDATE_CA_CMD:-update-ca-certificates}"
 
 mkdir -p "${CERTS_DIR}"
 
@@ -154,6 +159,19 @@ fi
 # the served CA is missing or stale.
 if ! halos_ca_publish_public "${HALOS_CA_ACTIVE_CRT}" "${PUBLIC_CA_DIR}"; then
     echo "ERROR: failed to publish CA copy at ${PUBLIC_CA_DIR}/halos-ca.crt; the /ca/halos-ca.crt download endpoint may serve stale or no CA until the next successful run." >&2
+fi
+
+# Install the active CA into the host system trust store so the device's own
+# desktop browser and CLI tools validate its HTTPS endpoints without operator
+# action (#158). The helper skips the costly update-ca-certificates rebuild
+# when the store already holds this CA, so the renewal timer doesn't churn the
+# bundle; only a CA rotation re-triggers it.
+#
+# Auxiliary: failure does NOT abort. A device that can't self-trust is a
+# quality regression, not a stack-blocking fault — the container leaf and its
+# CA are already in place by this point.
+if ! halos_ca_publish_system_trust "${HALOS_CA_ACTIVE_CRT}" "${SYSTEM_TRUST_DIR}" "${UPDATE_CA_CMD}"; then
+    echo "WARNING: failed to install the active CA into the host trust store at ${SYSTEM_TRUST_DIR}; the device's own browser/CLI may show cert warnings until the next successful run." >&2
 fi
 
 # Generate the Apple Configuration Profile (.mobileconfig) carrying the active
--- a/assets/halos-manage-certs
+++ b/assets/halos-manage-certs
@@ -91,8 +91,8 @@ COCKPIT_LEAF_OVERRIDE="${COCKPIT_WS_CERTS_DIR}/99-halos.cert"
 TRAEFIK_DYNAMIC_DIR="${HALOS_TRAEFIK_DYNAMIC_DIR:-/etc/halos/traefik-dynamic.d}"
 TRAEFIK_TLS_CONFIG="${TRAEFIK_DYNAMIC_DIR}/tls-default.yml"
 # Host CA trust store: the active CA is installed here so the device's own
-# desktop browser/CLI self-trust its HTTPS endpoints (#158). update_cmd is
-# overridable so tests can stub the bundle rebuild.
+# desktop browser/CLI self-trust its HTTPS endpoints (#158). HALOS_UPDATE_CA_CMD
+# is overridable so tests can stub the bundle rebuild.
 SYSTEM_TRUST_DIR="${HALOS_SYSTEM_TRUST_DIR:-/usr/local/share/ca-certificates}"
 UPDATE_CA_CMD="${HALOS_UPDATE_CA_CMD:-update-ca-certificates}"
 
--- a/assets/lib-ca.sh
+++ b/assets/lib-ca.sh
@@ -786,8 +786,12 @@ halos_ca_publish_public() {
 # rebuilding the system bundle each time would be pure churn. Only a CA
 # rotation (content change) re-triggers the rebuild.
 #
-# Failure semantics: on any failure leaves the pre-existing trust copy (if any)
-# untouched, removes the .new sibling, returns non-zero.
+# Failure semantics: a stage/cp/parse/mv failure leaves the pre-existing trust
+# copy (if any) untouched and removes the .new sibling. An update_cmd failure
+# happens after the new cert is installed, so the helper removes that file too —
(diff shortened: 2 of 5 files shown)


mairas:
## Code review — tiered persona pass

**Scope:** 5 files (`assets/lib-ca.sh`, `assets/halos-manage-certs`, `docs/CERTS.md`, two test files), ~99 LOC production shell.
**Intent:** Self-trust the device CA by installing it into the host trust store + running `update-ca-certificates`. Idempotent (no bundle churn on the 24h timer), rebuild-on-rotation, non-blocking on failure.
**Reviewers (7):** correctness, testing, maintainability, project-standards (always-on) + security, reliability, adversarial.

**Verdict: Ready with fixes.** One genuine P1 (self-trust silently breaks forever after a single `update-ca-certificates` failure), flagged independently by 4 of 7 reviewers. Fix is one line + the test that would have caught it. Everything else is P2/P3 hardening.

### P1 — Critical

|---|------|-------|-----------|------|
| 1 | `lib-ca.sh:849`/`:811` | **Rebuild permanently skipped after an `update-ca-certificates` failure.** `mv` (843) installs the trust file *before* `update_cmd` runs. If `update_cmd` fails, the function returns 1 but the file is already on disk matching the active CA. The next run's `cmp -s` (811) matches → returns 0 → `update-ca-certificates` is never retried. `/etc/ssl/certs` stays without the CA; the device shows cert warnings indefinitely with no further signal, and a 20-year auto-CA never rotates to break the deadlock. The idempotency check guards the wrong invariant (file bytes, not the bundle outcome). | correctness, reliability, maintainability, adversarial | ~0.97 |

**Fix:** `rm -f "$trust_file"` immediately before `return 1` on the `update_cmd` failure path. The next run then sees the file missing, reinstalls, and retries the rebuild — self-healing. Simpler than a sentinel file (KISS).

### P2 — High

|---|------|-------|-----------|------|
| 2 | `lib-ca.sh:828` | **X.509 validation doesn't check `basicConstraints CA:TRUE`.** The parse check (`openssl x509 -noout`) accepts any cert, so a non-CA cert (e.g. the device leaf, via a future wrong code path) could be installed as a host-wide trust anchor. `halos_ca_validate_pair` already does this check; this helper installs into the *system* store, so the bar is higher. Not exploitable today (the active CA is always a CA by construction) — defense-in-depth. | security | 0.82 |
| 3 | `test-lib-ca.sh`, `test-halos-manage-certs.sh` | **No test for the `update_cmd`-failure path** — neither the unit-level (file installed, rc=1) nor the integration-level aux-failure (WARNING logged, script continues, leaf still produced). This is exactly the test that would have caught finding #1. No test asserts the WARNING text either (the codebase tests this for the sibling `publish_public` ERROR). | testing, correctness, reliability, adversarial | 0.95 |
| 4 | `lib-ca.sh:797+` | **Stage+mv+parse boilerplate now duplicated** across `publish_public` and `publish_system_trust` (~28 lines). Third near-identical helper in the file. *Optional* — the repo already tolerates this for `publish_public`/`publish_mobileconfig`, and the two diverge (cmp-skip, `update_cmd`, no `chmod 0755`). Extract `_halos_ca_atomic_install` only if you agree the threshold is crossed. | maintainability | 0.82 |

### P3 — Low

|---|------|-------|-----------|------|
| 5 | `lib-ca.sh:849` | `update_cmd` output fully suppressed (`>/dev/null 2>&1`). On failure the only journal line is the static WARNING — the actual `update-ca-certificates` error (dpkg lock, broken hook) is lost. Let stderr flow: keep `>/dev/null`, drop `2>&1`. | correctness, reliability, maintainability, adversarial | 0.85 |
| 6 | `halos-manage-certs:94` | Comment says "update_cmd is overridable" but the var is `HALOS_UPDATE_CA_CMD` — a grep for the seam lands nowhere. | maintainability | 0.88 |
| 7 | `halos-manage-certs:97` | `HALOS_UPDATE_CA_CMD` read from the environment then executed as root. Requires root-equivalent to abuse (already root), so not escalation — but it carries a test seam into production. Could pass only via the function arg. | security | 0.63 |
| 8 | `lib-ca.sh:811` | `cmp` reads `trust_file` before the symlink-free guarantees; a symlink planted at the path (root-only) could force a false "current" match. Low value given root-only. | security | 0.61 |
| 9 | `test-lib-ca.sh:1230` | Second helper call uses `\|\| return 1` instead of capturing rc — weaker diagnostic than the rest of the suite. | testing | 0.75 |
| 10 | `test-halos-manage-certs.sh` | `sleep 1` in the idempotency test is unnecessary — `cmp` is content-based, not mtime-based. | correctness, maintainability | 0.7 |

### Residual risks (no action required)

- **No trust-store cleanup on package removal** — `halos-ca.crt` lingers in `/usr/local/share/ca-certificates` after purge. Benign on a single-owner device; operator-surprise on a repurposed one. Consider a `postrm` hook in a follow-up.
- **No `TimeoutSec=` on the oneshot** — a hung `update-ca-certificates` (ordered `Before=` the container stack) would block startup. Low probability on a Pi.
- **Orphan hash-symlink after rotation** — low-confidence (0.75): `update-ca-certificates` should handle same-filename content change, but an interrupted run combined with finding #1 could leave `/etc/ssl/certs` inconsistent. Fixing #1 mitigates.

### Standards

Clean. `VERSION` correctly **not** bumped (0.4.3 cycle already open vs latest stable `v0.4.2+8`); no `debian/changelog` edit; no hard-coded hostnames; `docs/CERTS.md` update appropriate.

---
Applying #1 (+ its test), #5, #6 now; #2 recommended. Pushing fixes to the branch.

🤖 tiered review via `mairas-workflow:review` (7 personas on Sonnet)

mairas:
## Review disposition

Fixes pushed (`9654d8f`, `bcd6696`):

- **P1 #1** (self-trust deadlock after `update-ca-certificates` failure) — fixed: helper now removes the trust file on `update_cmd` failure so the next run retries. Added the unit + integration failure-path tests that were missing.
- **P3 #5** (stderr suppressed) — fixed: `update_cmd` stderr now flows to the journal.
- **P3 #6** (stale comment naming) — fixed.
- **P2 #4** (duplication) — applied: extracted shared `_halos_ca_atomic_install`; existing publish tests pass through it unchanged.
- **P2 #2** (CA:TRUE check) — **skipped** by maintainer decision. Not exploitable today (the active CA is always a CA by construction); recorded as residual risk.

Remaining P3s (env-seam, symlink guard, test `sleep 1`) and residuals (postrm trust cleanup, `TimeoutSec=`) left as noted — none blocking.

Full suite green: lib-ca 122, halos-manage-certs 39, ca-download-cgi 14, ca-download-endpoint 23, lib-hostnames 45, icons 10, pytest 25.

PR #193, opened 2026-06-05
docs(certs): one-paste macOS CA trust that clears the stale cert

mairas:
Improves the `/ca` device-trust walkthrough's command snippets: fixes a silent re-trust failure on macOS, removes redundant manual download steps, and adds copy buttons.

## macOS — one-paste trust that clears the stale cert

- After a reflash the device regenerates its CA (same CN, new key). Running `security add-trusted-cert` over the old entry silently fails to update trust — the browser keeps warning until the user manually deletes the previous cert. The block now runs `security delete-certificate -c <CN>` first.
- The CN is read from the just-downloaded cert (the same `openssl … -nameopt RFC2253` + `sed` extraction `lib-ca.sh` uses), so the delete matches exactly what `add-trusted-cert` stores — rather than reconstructing it from the filename, which sanitizes the hostname and could diverge.
- `curl` fetches the cert into `/tmp`, folding the download into the command.

## Linux — fold the download into the install command

- System-trust path is now a single `sudo curl … -o /usr/local/share/ca-certificates/<file>` + `update-ca-certificates`, replacing download-then-`cp`. Re-running overwrites the file, so no stale-cert step is needed (unlike the macOS keychain).
- Firefox still needs the file for its GUI import, so that note now points at the top download button.

## Copy buttons on all command snippets

- A copy-to-clipboard button is injected (by JS, so a no-JS reader never sees a dead control) into each `<pre>`.
- Clipboard access needs a secure context, which a cert-warning click-through may not grant on the first visit. Three tiers: `navigator.clipboard.writeText` → hidden-textarea `execCommand('copy')` → select the snippet text for manual ⌘C.

## Shared mechanism

A new `[data-cert-origin]` span is filled client-side from `location.origin`, mirroring the page's existing filename-fill pattern, so the `curl` URLs point at the right device.

## Notes

- `test_landing_content.py`: macOS assertion updated for the new shape; added a guard that the copy button keeps its `execCommand` fallback. 22 passed.
- No VERSION bump: the `0.4.4` cycle is already open (latest stable is `v0.4.3+10`), so the repo-level check passes without one.
- With JS disabled the `curl` host shows a `https://this-device` placeholder; the snippets were already JS-dependent for the device-specific filename, so this matches existing behavior.


mairas:
## Code review — PR #193

**Scope:** `assets/ca-download/www/ca/index.html` (macOS/Linux snippets + copy-button JS), `run` (bumpversion target), `tests/test_landing_content.py`, VERSION/`.bumpversion.cfg`, solution doc.
**Intent:** macOS one-paste trust that clears the stale cert; Linux curl-fold; copy buttons with secure-context fallback; add the `./run bumpversion` target + required cycle bump.
**Reviewers (6):** correctness, testing, maintainability, project-standards (always-on) + security, adversarial (security-adjacent CA-trust commands, >50 LOC).

### Applied in this review (commit `a068234`)

| # | File | Issue | Reviewer(s) | Conf |
|---|------|-------|-------------|------|
| 1 | `run:191` | `command -v bumpversion` resolves the **shadowing function**, not the binary — the install-guard was dead and could never fire. Switched to `type -P`. | adversarial | 0.92 |
| 2 | `index.html` `flash()` | Re-clicking Copy within 1.5s captured the transient label as the restore value, leaving the button stuck on "Copied". Restore to fixed "Copy" + clear pending timer; select span explicitly (drops `lastChild` coupling). | correctness, maintainability | 0.85 |
| 3 | `tests` | The macOS **`delete-certificate`** step — the headline fix — had no assertion. Added presence + ordering (delete precedes add). | testing | 0.90 |
| 4 | `tests` | Linux curl-fold had no section-scoped guard; added one (`curl` + `update-ca-certificates`, no `cp`) and fixed a stale "Linux cp command" comment. | testing | 0.78 |
| 5 | `tests` | `execCommand("copy")` assertion coupled to exact quote style; loosened to `execCommand`. | testing | 0.70 |

### Advisory — not changed, with rationale

| # | File | Issue | Decision |
|---|------|-------|----------|
| 6 | `index.html:156` | CN-extraction `sed 's/.*CN=\([^,]*\).*/\1/p'` truncates at a comma (RFC2253 escapes `\,`). Adversarial rated this P1. | **Unreachable today:** the CN host segment is hostname-validated (`HALOS_HOSTNAMES_DNS_RE` — no commas), and the regex **matches `lib-ca.sh`'s own helper**. Calibrated to P3. Revisit only if the CA CN template ever gains punctuation — then both the page and `lib-ca.sh` need the `-nameopt multiline` form. |
| 7 | `index.html` | No-JS visitors see the literal `https://this-device` placeholder in the curl commands (regression from the relative download link). | Acknowledged tradeoff — snippets were already JS-dependent for the device filename. Optional: a `<noscript>` hint. |
| 8 | `index.html:156` | `delete-certificate -c <CN>` could remove the wrong anchor if two devices share the bare-legacy CN `HaLOS Device CA`. | Low, recoverable by re-running install. Tighter option: delete by `-Z <sha1>` of the downloaded cert. |

### Residual risks (report-only)

- **`curl -k` trust-bootstrap:** fetching the anchor over the not-yet-trusted channel with no out-of-band fingerprint check. Inherent to device-local CA distribution; defensible given the physical-ownership threat model. Lowest-noise mitigation if ever wanted: publish the CA SHA-256 on a separately-trusted channel (device label / Cockpit), not scarier curl flags.
- **`/tmp/halos-ca.crt` fixed path:** symlink/MITM surface on a shared macOS host; `mktemp` would tighten it. Low given the trust model.

### Coverage notes

- **1 false positive dropped:** project-standards flagged `.bumpversion.cfg` as skipping 0.4.4. Verified the base cfg was `0.4.4`; the bump was a clean `0.4.4 → 0.4.5`.
- **Process incident:** a read-only reviewer **executed** `./run bumpversion patch` while inspecting the target; with `commit = True` it created a spurious `0.4.5 → 0.4.6` commit locally. Caught, reverted, not pushed. Worth noting the new target is executable and side-effecting.
- **Injection check:** the CN flows through `"$(...)"` and is hostname-validated upstream — no shell-injection path; `location.origin`/filename use `textContent`, not `innerHTML`. No security defect found.

### Verdict: **Ready to merge.** Actionable findings fixed; remainder are advisory/acknowledged tradeoffs.

PR #195, opened 2026-06-14
feat(hostnames): resolve HALOS_DOMAIN once via a producer service

### LINKED ISSUE #194: Add halos-resolve-domain producer service; make core + routing consume it
Part of the cross-repo HALOS_DOMAIN architecture — tracking plan: halos-org/container-packaging-tools#206. **This must ship before** the generator change and downstream rebuilds, because consumer units will order against / depend on the producer introduced here.

## Goal

Introduce a single producer that resolves the canonical hostname once and publishes it, so every container-app systemd unit can load `HALOS_DOMAIN` via `EnvironmentFile` instead of each prestart re-resolving it. Convert core's own units and `configure-container-routing` from resolvers to consumers.

## Unit A1 — Producer service

**Files**
- Create: `debian/halos-core-containers.halos-resolve-domain.service` (oneshot)
- Create: small ExecStart helper (e.g. `assets/halos-resolve-domain` installed under `/usr/lib/halos-core-containers/`)
- Modify: `debian/postinst` (enable unit; `/run/halos` already created at postinst:45, or use `RuntimeDirectory=halos`)
- Modify: `debian/rules` / install wiring as needed
- Test: shell test mirroring `tests/test-lib-hostnames.sh` seams

**Approach**
- Source `/usr/lib/halos-core-containers/lib-hostnames.sh`, `halos_load_hostnames`, write `HALOS_DOMAIN=$(halos_canonical_hostname)` to `/run/halos/domain.env`. Recompute on every start (no `[ -z ]` guard).
- `[Unit] After=local-fs.target`, `Before=halos-core-containers.service` (mirror `halos-manage-certs.service`). `Type=oneshot`.
- The lib always yields a valid value (canonical or `<short>.local` fallback), so `/run/halos/domain.env` is always well-formed.

**Test scenarios**
- Happy path: fixture `hostnames.conf` with an fqdn-first canonical + injected resolver -> `/run/halos/domain.env` contains the canonical fqdn.
- Edge: no/empty domain resolves -> `${fqdn}` soft-drops -> file contains `<short>.local`.
- Error: `hostnames.conf` absent -> file contains `<short>.local` fallback; exit 0 under `set -e`.

**Verification:** booting writes `/run/halos/domain.env=HALOS_DOMAIN=<canonical>` before `halos-core-containers.service` starts.

## Unit A2 — core + routing become consumers

**Files**
- Modify: `debian/halos-core-containers.service` (add `After=`/`Wants=halos-resolve-domain.service` + `EnvironmentFile=-/run/halos/domain.env`)
- Modify: `prestart.sh` (drop the `halos_canonical_hostname` resolution; inherit `$HALOS_DOMAIN`; still write `HOSTNAME`)
- Modify: `assets/configure-container-routing` (remove the `[ -z HALOS_DOMAIN ]` `.local` guard at lines ~32-36; consume inherited value)

**Approach**
- After this, nothing in core writes its own `HALOS_DOMAIN` back into a file that is later re-read as an `EnvironmentFile` — structurally killing the stale-pin feedback loop.

**Test scenarios**
- core `runtime.env` no longer self-derives `HALOS_DOMAIN`; the value comes from the producer.
- `configure-container-routing` uses the inherited `HALOS_DOMAIN` (no `.local` fallback path) when the producer has run.

**Verification:** core container + routing labels reflect the producer's canonical value; no `HALOS_DOMAIN` line emitted by core prestart.

## Out of scope
- Live-reload of `hostnames.conf` (explicitly excluded from the whole effort).
- Changes to `lib-hostnames.sh` resolution logic (consumed differently, not changed).

mairas:
Unit 1 of the cross-repo HALOS_DOMAIN refactor (halos-org/container-packaging-tools#206). **Ships first** — it introduces the producer that the generator change and downstream rebuilds order against. Closes #194.

## Why

`HALOS_DOMAIN` (the canonical hostname / OIDC issuer) was resolved independently in several places, and core's `prestart.sh` wrote it into `runtime.env`, which the systemd unit re-loads as an `EnvironmentFile` — so a stale value pinned itself across restarts (the root cause behind halos-org/halos-marine-containers#177).

## What

- **New `halos-resolve-domain.service`** (oneshot) + `assets/halos-resolve-domain`: sources `lib-hostnames.sh`, resolves the canonical hostname, and publishes `HALOS_DOMAIN` to `/run/halos/domain.env`. Ordered `Before=halos-core-containers.service`.
- **Core unit** gains `After=`/`Requires=halos-resolve-domain.service` and `EnvironmentFile=-/run/halos/domain.env` (listed last, so it overrides any stale `HALOS_DOMAIN` a pre-upgrade `runtime.env` still carries).
- **`prestart.sh`** no longer derives `HALOS_DOMAIN` or writes it to `runtime.env`; it inherits it from the unit environment. It still sources the loader for the hostname *list* (`HALOS_HOSTNAMES_DNS[]`) that the Authelia per-hostname cookies block and OIDC `redirect_uris` expansion consume — `domain.env` carries only the canonical name.
- **`configure-container-routing`** drops its own `${hostname}.local` fallback and reads `HALOS_DOMAIN` from the environment (used only for diagnostic logging).

No consumer writes `HALOS_DOMAIN` back into a file it later re-reads, so the `runtime.env`-as-`EnvironmentFile` feedback loop is structurally eliminated. The producer is `Type=oneshot` without `RemainAfterExit`, so an explicit service restart re-runs it and republishes `domain.env` (picking up `hostnames.conf` edits) — same pattern as `halos-manage-certs.service`. Live-reload via a file watcher is explicitly out of scope.

## Tests

- `tests/test-halos-resolve-domain.sh` — producer behavior: literal canonical, `.local` fallback when no config, output-dir creation, and rerun-republish.
- `tests/test-prestart-loader-contract.sh` — asserts `prestart.sh` sources the loader and that every `halos_*`/`_halos_*` symbol it invokes is defined by `lib-hostnames.sh`. This guards the over-deletion class that review caught during implementation (removing the loader source while keeping calls to it would abort prestart under `set -e` and fail core boot).
- All 8 shell suites pass; pre-commit hooks (changelog, hardcoded-hostname) pass.

## Design notes

- `Requires=` (not `Wants=`) on the producer is deliberate: a failed resolver means `HALOS_DOMAIN` is genuinely unavailable, so failing loud beats booting core with a broken OIDC issuer. Matches `halos-manage-certs.service`.
- `VERSION` not bumped — mid release cycle per repo policy.

- `systemctl status halos-resolve-domain.service` → `active (exited)` / clean success before core starts.
- `cat /run/halos/domain.env` → `HALOS_DOMAIN=<canonical>` (the configured fqdn, or `<short>.local` fallback).
- Core containers come up; Homarr/Authelia OIDC issuer reflects the canonical hostname.
- Edit `/etc/halos/hostnames.conf`, `systemctl restart halos-core-containers.service`, confirm `domain.env` updates and OIDC follows.
- Failure signal: `halos-resolve-domain.service` failed → core blocked by `Requires=`. Mitigation: check `journalctl -u halos-resolve-domain.service`; the lib always yields a value unless the lib file itself is missing (packaging fault).
- Rollout ordering: this package (the producer) must be installed/upgraded before any downstream container app that `Requires=` it — enforced by the versioned `Depends` the generator adds in #206.


mairas:
Pushed two follow-up commits:

- **`92dc209` Bump VERSION 0.4.5 → 0.5.0.** The open cycle (since stable `v0.4.3`) already carries minor-level features (mobileconfig CA carrier, CA-page copy buttons, tile anchoring) plus this producer, but `VERSION` had only been patch-bumped. Escalating to `0.5.0` so the release number reflects the cycle's actual content — and so #206's downstream `Depends: halos-core-containers (>= 0.5.0)` floor is clean.
- **`be4580e` docs(agents): highest-level-wins version-bump rule.** The per-cycle policy gated only on *whether* a bump happened, not its level, so a minor change in a patch-opened cycle would ship under a patch number. Added the exception that justifies the bump above.

The workspace and marine `AGENTS.md` carry the same policy text and get the same clause in a separate cross-repo docs change.

Note: this repo's `AGENTS.md` version-bump wording predates the workspace's rewritten (GitHub-Releases-based) policy — a pre-existing drift worth reconciling separately; the highest-level-wins clause was added to the current local phrasing.

PR #202, opened 2026-08-06
Fix and trigger the OIDC client registration reload

### LINKED ISSUE #200: reload-oidc-clients is non-functional: wrong data root, and restarts Authelia without its environment
Found while debugging an OIDC login failure on a test device. `/usr/bin/reload-oidc-clients` cannot complete on any device. Two independent defects, both present on `origin/main` (`7d97178`), not just in the installed `0.6.0-1`.

## 1. Wrong data root — the tool aborts immediately

```sh
CONTAINER_DATA_ROOT="/var/lib/container-apps/${PACKAGE_NAME}"   # line 10
AUTHELIA_DATA="${CONTAINER_DATA_ROOT}/authelia"                 # line 11
```

The real value, from the generated `env.defaults` on a device:

```
CONTAINER_DATA_ROOT="/var/lib/container-apps/halos-core-containers/data"
```

The `/data` segment is missing, so `${CONTAINER_DATA_ROOT}/authelia` points at a directory that does not exist. Observed:

```
$ sudo reload-oidc-clients
ERROR: Authelia secrets not found. Run halos-core-containers first.
```

```
/var/lib/container-apps/halos-core-containers/authelia        absent
/var/lib/container-apps/halos-core-containers/data/authelia   EXISTS
```

And the running container confirms which one is real:

```
$ docker inspect authelia --format '{{range .Mounts}}{{.Source}} -> {{.Destination}}{{"\n"}}{{end}}'
/var/lib/container-apps/halos-core-containers/data/authelia -> /config
```

The abort is actually the *lucky* outcome. `AUTHELIA_OIDC_FILE` derives from the same variable, so had the secrets check passed, the tool would have written the client registration to a path Authelia never reads — succeeding silently while changing nothing.

**Fix:** line 11 → `AUTHELIA_DATA="${CONTAINER_DATA_ROOT}/data/authelia"`, or better, source `env.defaults` rather than reconstructing the path. Line 154's `cd /var/lib/container-apps/${PACKAGE_NAME}` is correct as-is — the compose file does live at the app root, which is what makes the inconsistency easy to miss.

## 2. The Authelia restart runs without the compose environment

```sh
cd /var/lib/container-apps/${PACKAGE_NAME}   # line 154
docker compose restart authelia              # line 155
```

The systemd unit supplies `CONTAINER_DATA_ROOT`, `HALOS_DOMAIN`, `REDIS_PASSWORD`, `AUTH_SECRET` and others via `EnvironmentFile`. Run standalone — which is the documented usage — none of them are set. With defect 1 patched locally, this is what happens:

```
Restarting Authelia...
WARN The "CONTAINER_DATA_ROOT" variable is not set. Defaulting to a blank string.
WARN The "HALOS_DOMAIN" variable is not set. Defaulting to a blank string.
WARN The "REDIS_PASSWORD" variable is not set. Defaulting to a blank string.
decoding failed due to the following error(s):
error decoding 'services[homarr].extra_hosts': bad host name ''
```

The re-render succeeds and the restart does not, so the tool leaves a correct registration on disk that Authelia is not running. Nothing in its output distinguishes this from success — the failure is on compose's stderr and the script does not check the exit status.

**Fix:** source `${ETC_DIR}/env.defaults` (and `env`) before the compose call, or restart via `systemctl restart halos-core-containers.service`, or use plain `docker restart authelia`, which needs no variable resolution. `set -e` is already on but does not help here, because the failing command is the last one.

## Reproduction

On any device with the package installed:

```sh
sudo reload-oidc-clients          # defect 1: aborts on the secrets check
```

Patch line 11 to add `/data`, rerun, and defect 2 appears at the restart step.

## Impact

This is the only mechanism that transfers a client secret into Authelia's registration. While it is broken, any app whose `oidc-secret` is regenerated authenticates with a secret Authelia does not know, and OIDC login fails with `invalid_client` at token exchange. See the companion issue on the missing trigger — together they make the failure silent and indefinite.

## Verified fix (on-device, package binary untouched)

A corrected copy run from `/tmp`, then `docker restart authelia`, restored login. Confirmed by probing the token endpoint directly: the real secret returns `invalid_grant` (client accepted, only the deliberately-invalid code rejected), while a wrong secret still returns `invalid_client`, so the probe is not vacuous.

### LINKED ISSUE #201: Nothing invokes reload-oidc-clients, so a rotated client secret silently breaks OIDC forever
Companion to #200. That issue is why the tool cannot run; this one is why nobody finds out.

## The gap

App prestarts **write** an OIDC client snippet and stop there. From a generated marine app prestart:

```sh
mkdir -p /etc/halos/oidc-clients.d
cat > /etc/halos/oidc-clients.d/signalk.yml << 'SNIPPET'
client_id: signalk
client_secret_file: /var/lib/container-apps/marine-signalk-server-container/data/oidc-secret
...
SNIPPET
```

Nothing then calls `reload-oidc-clients`, which is the only thing that hashes those secrets into Authelia's `oidc-clients.yml` and restarts it.

The script's own header says otherwise:

```sh
# Called automatically when files in /etc/halos/oidc-clients.d/ change
```

It is not. On a device with the package installed:

```
$ systemctl list-unit-files | grep -i watcher
(nothing)
```

`metadata.yaml:47` lists `reload-oidc-clients` under `system_bin` — "system binaries provided for other containers to use" — not under a watcher. And no app prestart calls it:

```
$ grep -rn 'reload-oidc-clients' halos-marine-containers/apps/*/prestart.sh
(no matches)
```

So the registration only ever advances when a human runs the tool by hand.

## What this produces

Observed on a test device. Timeline reconstructed from file mtimes, independent of any journal:

| Artifact | When |
|---|---|
| Authelia container started | Jul 5 17:39 |
| `data/authelia/oidc-clients.yml` last rendered | Jul 5 18:39 |
| Signal K's `data/oidc-secret` regenerated | **Aug 2 22:05** |

The app package's postinst generates a client secret when none exists. On Aug 2 it did exactly that, and the snippet was rewritten on the next start — but the rendered registration still held the hash of the Jul 5 secret, and Authelia had not restarted since Jul 5.

Result: Signal K sent a secret Authelia did not know. Every login died at token exchange:

```
OIDC callback error: OIDCError: Token exchange failed: invalid_client -
  Client authentication failed (e.g., unknown client, no client authentication
  included, or unsupported authentication method).
```

The redirect to Authelia works, the user authenticates, the authorization code comes back — and only the final exchange fails. It looks like an app bug, not a registration bug. It went unnoticed for four days on a device in daily use, and would have persisted indefinitely.

## Why the failure is silent

- The snippet is rewritten on every start, so it always *looks* current.
- The rendered registration carries no reference to the secret it was built from, so the two cannot be compared without hashing.
- Nothing warns that a snippet is newer than the registration.
- With #200 unfixed, even a manual run appears to do nothing or — once past the abort — reports success while restarting nothing.

## Fix direction

Pick one trigger and make it real:

1. **Declare the file watcher** the header already claims. `container-packaging-tools` has a watcher feature (`has_file_watchers`), which is what the comment appears to have been written against; wiring `/etc/halos/oidc-clients.d` to `reload-oidc-clients` would match the documented behaviour.
2. **Call it from the app prestart** after writing the snippet. Simple, but runs on every start and restarts Authelia from an app's `ExecStartPre` — likely worse.
3. **Call it from the app postinst**, which is where a secret is actually generated, and is the narrowest trigger.

Whichever is chosen, add a staleness check so this cannot fail silently again: compare each snippet's `client_secret_file` mtime against the rendered registration and warn (or re-render) when a snippet is newer.

Fix #200 first — the trigger is worthless while the tool it triggers cannot run.

mairas:
`reload-oidc-clients` could not complete on any device, and nothing invoked it. Together those made a rotated client secret break OIDC login silently and indefinitely — observed on a test device, four days unnoticed.

## What was wrong

The tool rebuilt `CONTAINER_DATA_ROOT` from the package name instead of reading the env files `postinst` writes, landing one directory above the volume compose mounts. It aborted on the secrets check; past that check it would have written the registration where Authelia never reads and reported success. Its Authelia restart went through `docker compose`, which cannot resolve `docker-compose.yml`'s variables without the unit's `EnvironmentFile` and so fails exactly when the tool is run the documented way — by hand.

Nothing then called it: no watcher, no prestart, no postinst, despite the script header claiming otherwise. The registration only advanced when someone ran the tool manually.

## Approach

The merge now lives once, in `assets/lib-oidc-clients.sh`, sourced by both `prestart.sh` and the tool. The two copies were identical when written; only one carried the path bug, so the drift is the actual defect and one implementation removes it rather than this instance of it.

`halos-oidc-clients-reload.path` watches `/etc/halos/oidc-clients.d/` and runs the tool on change. Apps rewrite their snippet on every start, so the trigger fires far more often than the registration changes: the merger digests its inputs — snippet text, referenced secret contents, hostname list, signing material — into a stamp beside the rendered file and returns early when they match. Without that, every app restart would bounce SSO.

A failed hash (it runs the Authelia image, so it fails for reasons unrelated to the snippets) now keeps the previous registration and drops the stamp instead of stamping a registration with the client missing — the same silent-staleness class the rest of this fixes. `prestart.sh` treats that as a warning rather than refusing to boot.

## Verification

23 new tests in `tests/test-lib-oidc-clients.sh` cover the merge, change detection, the failure paths, and both #200 defects as regressions; the full suite passes.

Systemd path semantics and the real data root were checked on a test device, using a throwaway probe client so no app's secret was touched:

- new snippet appears → watcher ran, 4 clients merged, Authelia restarted
- app rewrites an unchanged snippet → watcher ran, `already current`, Authelia **not** restarted
- secret rotates, then the app rewrites the same snippet (the #201 failure) → registered hash changed, Authelia restarted
- probe removed → watcher ran, registration back to 3 clients, Authelia healthy

The device was restored to its packaged state afterwards.

Closes #200
Closes #201


mairas:
## Code review — 8 personas

**Scope:** `fix/oidc-client-reload` vs `7d97178`, 12 files, +941/−282
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) · security (client-secret handling, SSO trust path) · reliability (restart semantics, boot path, packaging) · adversarial (941-line diff touching auth) · api-contract (the snippet format is a contract other repos write against)

Three findings below were reproduced directly rather than taken on the reviewers' word; those are marked **verified**.

### P1 — must fix before merge

|---|---|---|---|---|---|
| 1 | `tests/test-lib-oidc-clients.sh:409` | **CI is red.** `stat -f '%Lp'` is probed before `stat -c '%a'`, but GNU `stat -f` is `--file-system`: it prints the filesystem blob to stdout, exits non-zero, and the `\|\|` fallback appends `600`. Passes on macOS, fails in the trixie CI container. **Verified**: 22 passed, 1 failed in `debian:trixie`. Repo precedent (`test-halos-manage-certs.sh:146`) probes GNU first. | testing | 0.97 | safe_auto |
| 2 | `assets/lib-oidc-clients.sh` (stamp write) | **The stamp latches a registration that was never applied.** The stamp is written when the render succeeds, before the caller restarts Authelia. If `docker restart authelia` fails, the file on disk is new, Authelia still holds the old client list, and every later trigger matches the stamp and exits 0 doing nothing. **Verified**: run 1 with a failing restart → rc=1; run 2 → "already current", zero docker calls. That is #201 reintroduced through the mechanism meant to close it, and the exit-0 no-op erases the failed-unit signal. | correctness, reliability, testing | 0.90 | gated_auto |
| 3 | `debian/…-reload.path:15` | **Snippet writes during a run are dropped.** systemd's `path_enter_running()` unwatches the directory for the duration of the triggered unit, and re-arms without re-checking (PathModified is edge-triggered). A run takes seconds (one `docker run` per client). During a bulk app install, app B's snippet write lands inside app A's run and is lost — B is never registered until something else writes a snippet. | correctness, adversarial | 0.82 | manual |
| 4 | `debian/…-reload.service:10` | **No `TimeoutStartSec`, and dpkg starts it synchronously.** `Type=oneshot` disables the start timeout by default; the generated postinst runs `deb-systemd-invoke start` on both units without `--no-block`. An unresponsive docker daemon hangs `apt` indefinitely holding the dpkg lock. A hung oneshot also leaves the path unit permanently in `running`, so no reload ever fires again. On a fresh install the oneshot runs before Authelia's secrets exist and lands in `systemctl --failed`. | reliability, testing, adversarial | 0.80 | gated_auto |
| 5 | `assets/lib-oidc-clients.sh:24` | **The digest omits the renderer's own code.** `HALOS_OIDC_RENDER_VERSION` is a hand-maintained integer whose documented bump rule ("any change to the rendered output format") does not cover `halos_expand_oidc_redirect_uri` in `lib-hostnames.sh`, which builds every `redirect_uris` entry. A future change there alters the output while every device's stamp still matches → registrations frozen at the pre-upgrade expansion, silently. Same failure class this PR exists to kill, relocated one file over. | adversarial, maintainability, api-contract | 0.75 | manual |

### P2 — moderate

|---|---|---|---|---|---|
| 6 | `assets/lib-oidc-clients.sh` (render) | The heredoc write is unchecked and `mv` runs unconditionally. `_halos_oidc_render` is invoked as an `if` condition, which disables errexit for its whole body, so ENOSPC on an SD card replaces a good registration with a truncated one, stamps it, and restarts Authelia onto it. Traefik's `depends_on: authelia condition: service_healthy` then holds the whole web stack down. | correctness, security | 0.80 | gated_auto |
| 7 | `assets/lib-oidc-clients.sh` (merge) | No mutual exclusion between prestart's merge and a path-triggered one. `PartOf=` does not propagate to automatic `Restart=on-failure`, and prestart writes `homarr.yml` into the watched directory during `ExecStartPre`. Interleaved mv/stamp pairs leave a stamp describing a file the other run replaced. | correctness, reliability, security, adversarial | 0.70 | manual |
| 8 | `assets/reload-oidc-clients:14` | `HALOS_OIDC_ROOT` is an ungated env seam that relocates the shell libraries this root process **sources**. AGENTS.md documents the opposite convention for `HALOS_DOMAIN_RESOLVER` ("gated on `declare -F` so a stray environment export cannot short-circuit production resolution or trigger command execution"), and the repo's own `halos-manage-certs` uses narrow per-path overrides instead. Concrete bad case: `CONTAINER_DATA_ROOT` inherited from the environment survives the missing-`env.defaults` guard, so a fake root writes the empty placeholder over the real registration and restarts Authelia onto it. | security, adversarial, project-standards, maintainability | 0.75 | manual |
| 9 | `assets/lib-oidc-clients.sh` (secret read) | Gate is `[ -f ]`, never `[ -s ]`: a zero-byte secret file registers the client with the hash of the empty string and stamps it current. Newly reachable — the path unit fires the moment a snippet appears, which can precede the app writing its secret. | testing | 0.85 | gated_auto |
| 10 | `prestart.sh:429` | The comment justifying the tolerated merge failure ("the reload path unit retries on the next snippet write") is false for the core stack's own client: `homarr.yml` is written by prestart itself, while the path unit is not yet running (`After=`/`WantedBy=` the service, which is active only after `ExecStartPre`). On a core-only device a transient hash failure disables SSO until someone restarts the stack, signalled by one WARNING line. | reliability, adversarial | 0.75 | manual |
| 11 | `assets/reload-oidc-clients:49` | The reload re-resolves hostnames live and rewrites only `oidc-clients.yml`; `configuration.yml`'s `session.cookies` block is written only by prestart. A DHCP domain appearing at runtime therefore applies half a hostname change and bounces Authelia — which also falsifies the "a routine app restart no longer bounces SSO" claim, since the digest covers a network-derived input. | adversarial | 0.70 | manual |
| 12 | `docs/SSO_ARCHITECTURE.md` | The rewritten section now contradicts three later passages that still describe the prestart-only model (`:499`, `:508`, `:515`); the config tree at `:235` names `authelia-container/data` while the command added at `:261` uses the real path; the merged-output example still shows the old header. Both AGENTS.md and the doc also claim secret rotation registers automatically — the unit watches only the snippet directory; rotation is picked up because apps rewrite their snippet on start, which is an unstated requirement on other repos' packages. | project-standards, api-contract, maintainability | 0.80 | gated_auto |
| 13 | `assets/reload-oidc-clients:57` | `metadata.yaml` advertises the tool under `system_bin` "for other containers to use", and it now exits 0 doing nothing whenever the digest matches — with no `--force`. The only documented escape is `rm` of an internal implementation file. | api-contract | 0.85 | manual |
| 14 | `debian/…-reload.service` | No `StartLimitIntervalSec=0`. The default 5-starts-in-10s limit is reachable during a bulk install, and `path_trigger_notify_impl()` propagates `start_limit_hit` by putting the **path** unit into failed state — the trigger silently stops existing. | correctness, reliability | 0.65 | gated_auto |

| File | Issue | Route |
|---|---|---|
| `docker-compose.yml:155` | Homarr bind-mounts all of `/var/lib/container-apps` read-only, which contains Authelia's `secrets.env` and `oidc_private_key.pem`. The container runs as uid 0 with no userns remap, so 0600 root-owned modes do not help: an RCE in the network-facing dashboard yields the IdP's signing key and session secret. **P1, file issue.** | file-issue |
| `assets/lib-oidc-clients.sh:37` | The plaintext client secret is passed as `--password` on the docker command line, i.e. into world-readable `/proc/<pid>/cmdline`. Now runs unattended on every snippet change. | fix-here |
| `assets/lib-oidc-clients.sh` (parser) | Shell YAML parsing takes *every* matching line, so a duplicated top-level key injects arbitrary top-level Authelia config; flow-style or zero-indent `redirect_uris` render as `redirect_uris: null`, which Authelia rejects at load → crash loop with `restart: unless-stopped`, all SSO down, and the stamp records it as current. | fix-here (guard) + file-issue (pre-apply validation) |
| `assets/lib-oidc-clients.sh:37` | Hashing uses floating `authelia/authelia:4.39` while compose pins `4.39.19` — a second image to pull, and an implicit registry dependency on a boat. | fix-here |
| `assets/lib-oidc-clients.sh:44` | `halos_oidc_hash_secret` returns the pipeline's status (always 0), so output without a `Digest:` line yields an empty hash that renders as `client_secret: ''`. | fix-here |
| `prestart.sh:306` | First-boot regeneration is gated on `secrets.env` alone but writes `oidc_private_key.pem` seconds later, after `openssl genrsa 4096`. Power loss between the two wedges every subsequent boot (prestart aborts under `set -e`, stack never starts). | file-issue |
| container-packaging-tools | `postrm.j2` deletes `{app_id}.yml` but `oidc_snippet.py` writes `{client_id}.yml`. They differ today for signalk-server (`app_id: signalk-server`, `client_id: signalk`), so purging leaves an orphan snippet — and this PR's docs now promise removal auto-applies. | file-issue (cross-repo) |
| docs.halos.fi | Public docs still describe prestart-only merging and "regenerates on next restart". | file-issue (cross-repo) |

### Accepted trade-off

A hash failure aborts the whole render and keeps the previous registration, rather than skipping the failing client. api-contract flags that one unhashable client blocks every other app's updates. That is deliberate: dropping a client from the registration breaks exactly the logins this mechanism protects, and the alternative silently ships a registration missing a client. The blast radius is worth stating in the docs.

### Coverage

- Suppressed below 0.60 confidence: 6 findings.
- Testing gaps worth pinning: no test asserts the system converges after a failed restart (finding 2 ships green); no test parses the rendered YAML — a mutation breaking the jwks indentation keeps all 23 tests passing; the hmac/private-key half of the digest is unpinned (deleting it keeps the suite green); nothing asserts the new units actually land in the `.deb`, which is the exact trap `debian/rules` documents in its own comment; nothing couples `PathModified=` to the tool's `OIDC_CLIENTS_DIR`.
- The device verification in the PR description exercised single, well-spaced snippet writes. It does not cover findings 2, 3, 4, or 7, all of which need concurrency, failure injection, or the packaging path.

mairas:
## Review findings addressed

All 14 active findings are fixed in 336220f. Five reported elsewhere as issues.

| # | Finding | Resolution |
|---|---|---|
| 1 | CI red: BSD `stat` probed before GNU | Probe order inverted, matching `test-halos-manage-certs.sh`. Verified in `debian:trixie`. |
| 2 | Stamp latched a registration that was never applied | Recording the merge is now `halos_oidc_commit`, called by the caller only once the render is live. Test: a failed restart is retried on the next run. |
| 3 | Path unit drops snippet writes during a run | The tool re-merges until nothing changes (bounded at 3). Test: the stub adds a client during the restart; the run picks it up. |
| 4 | No `TimeoutStartSec`; dpkg started the one-shot synchronously | `TimeoutStartSec=300`, `StartLimitIntervalSec=0`, and `--no-start` for the service. It gains `WantedBy=halos-core-containers.service`, so it runs once per stack start — which is also the retry finding 10 said did not exist. |
| 5 | Digest omitted the renderer's own code | `HALOS_OIDC_RENDER_VERSION` is gone; the digest covers both shell libraries, so a change to `halos_expand_oidc_redirect_uri` invalidates the stamp on its own. Test appends to `lib-hostnames.sh` and expects a re-render. |
| 6 | Unchecked render writes under suppressed errexit | Every write checked, temp file removed via a `RETURN` trap. |
| 7 | No serialization between the two callers | `flock` around the merge (util-linux, so always present on the target; a dev machine without it runs unserialized). |
| 8 | `HALOS_OIDC_ROOT` relocated sourced libraries | Replaced by per-path overrides, following `halos-manage-certs`. `CONTAINER_DATA_ROOT` is cleared before the env files are read, so an inherited value cannot redirect the registration. Test pins that. |
| 9 | Empty secret file registered the empty-string hash | `-s`, not `-f`. |
| 10 | prestart's "the path unit retries" was false for the core stack's own client | True now — see finding 4. |
| 11 | redirect_uris expanded against a live hostname list | Expanded against `hostnames.snapshot`, the list prestart used for the `session.cookies` block, so the two halves of Authelia's config cannot disagree between stack starts. |
| 12 | Stale and contradictory doc sections | Config tree, generated-file example, install/removal flows, the design principle, the unimplemented "supported optional fields", and the force recipe all corrected. Both docs now state that the trigger is a snippet write and that an app must rewrite its snippet after rotating its secret. |
| 13 | No force escape hatch | `reload-oidc-clients --force`. |
| 14 | Start-limit could fail the path unit permanently | `StartLimitIntervalSec=0`. |

Pre-existing, fixed here because the diff already touches the code: the plaintext secret now travels in the environment rather than docker's argv, hashing uses the pinned `4.39.19` image compose already runs instead of a floating tag, an exit-0 hash with no digest is an error rather than an empty `client_secret`, and a snippet whose `redirect_uris` the shell parser cannot read is skipped instead of rendering an empty key that Authelia rejects at load.

Filed rather than fixed here: #203 (Homarr can read Authelia's signing key through the `/var/lib/container-apps` mount), #204 (validate the rendered registration before restarting), #205 (first boot interrupted between `secrets.env` and the signing key wedges the stack), container-packaging-tools#243 (`postrm` removes `{app_id}.yml` but snippets are `{client_id}.yml`), docs#36.

**Tests:** 34 in `test-lib-oidc-clients.sh` (was 23), plus a new `test-packaging-contract.sh` that pins each unit to its `dh_installsystemd --name=` call — the trap `debian/rules` documents, and how the trigger came to be missing in the first place — and couples `PathModified=` to the directory the tool merges. `test-prestart-loader-contract.sh` now checks the library's four caller globals and that the merge call stays guarded.

**On-device:** re-verified on the test hardware — a changed library re-renders and restarts Authelia (~3.2s), a second run is a no-op, `--force` re-applies, and Authelia returns to healthy. The device was restored to its packaged state.

PR #209, opened 2026-08-10
Fix boot-time ordering cycle that kept the core stack from starting

### LINKED ISSUE #208: Core stack does not start at boot: ordering cycle via halos-oidc-clients-reload.path
`halos-oidc-clients-reload.path` introduces a systemd ordering cycle that keeps the entire core stack from starting at boot. Traefik, Authelia and Homarr are down on a freshly flashed device until someone starts the unit by hand.

Reproduced on a HALPI2 flashed with `HaLOS Desktop Marine HALPI2 2026-08-08.0` (`halos-core-containers 0.7.0-1`):

```
$ systemctl is-active halos-core-containers.service
inactive
$ journalctl -b | grep halos-core-containers
systemd[1]: halos-core-containers.service: Found ordering cycle on halos-manage-certs.service/start
systemd[1]: halos-core-containers.service: Found dependency on basic.target/start
systemd[1]: halos-core-containers.service: Found dependency on paths.target/start
systemd[1]: halos-core-containers.service: Found dependency on halos-oidc-clients-reload.path/start
systemd[1]: halos-core-containers.service: Found dependency on halos-core-containers.service/start
systemd[1]: halos-core-containers.service: Job halos-manage-certs.service/start deleted to break ordering cycle starting with halos-core-containers.service/start
```

systemd resolves the cycle by dropping the `halos-manage-certs.service` job, and `halos-core-containers.service` — which `Requires=` it — never gets a job at all: `journalctl -u halos-core-containers.service -b` is empty. `halos-oidc-clients-reload.service` still gets pulled in by its `WantedBy=` and fails twice with `ERROR: Authelia secrets not found ... Start halos-core-containers first`, which is how the breakage first surfaces in `systemctl --failed`.

## Where the cycle comes from

`systemd.path(5)`: unless `DefaultDependencies=no` is set, a path unit implicitly gets `Before=paths.target` (plus `After=sysinit.target` and the shutdown conflicts). Combined with the explicit `After=halos-core-containers.service` added in #202, the edges close:

```
halos-core-containers.service
  → After= halos-manage-certs.service      (explicit)
  → After= basic.target                    (service default dependency)
  → After= paths.target                    (basic.target ordering)
  → After= halos-oidc-clients-reload.path   (path unit default Before=paths.target)
  → After= halos-core-containers.service    (explicit, #202)
```

Any service ordered after `basic.target` — i.e. every ordinary service — closes this loop with a path unit that declares `After=` on it. Nothing about our unit is special; the `After=` on a default-dependencies path unit is the defect.

## Scope

Introduced by #202 (`halos-oidc-clients-reload.path`, released in v0.7.0+1 on 2026-08-08). Every device running 0.7.0-1 boots without the core stack.

## Fix direction

Keep the intent of the `After=` (don't watch the snippet directory before the stack is up) while removing the implicit `Before=paths.target` edge: set `DefaultDependencies=no` on the path unit and restore the shutdown ordering it would otherwise lose (`Conflicts=shutdown.target`, `Before=shutdown.target`). Verification is a reboot: no `Found ordering cycle` in the boot journal, and `halos-core-containers.service` active with Traefik/Authelia/Homarr up.

mairas:
`halos-oidc-clients-reload.path` closed a systemd ordering cycle that kept the entire core stack from starting at boot: on a device flashed with the 2026-08-08 image, Traefik, Authelia and Homarr were simply not running, and the only visible symptom was `halos-oidc-clients-reload.service` failing with "Authelia secrets not found".

Fixes #208

## Why it cycled

A path unit gets `Before=paths.target` from its default dependencies, and `halos-core-containers.service` is ordered after `basic.target`, which is after `paths.target`. The `After=halos-core-containers.service` added in #202 — there so the watcher does not fire before the stack exists — closed the loop:

```
halos-core-containers.service → halos-manage-certs.service → basic.target
    → paths.target → halos-oidc-clients-reload.path → halos-core-containers.service
```

systemd broke the cycle by deleting the `halos-manage-certs.service` job. That unit is a `Requires=` of the stack, so `halos-core-containers.service` never got a job at all — `journalctl -u halos-core-containers.service -b` was empty on the affected boot.

## The fix

`DefaultDependencies=no` on the path unit, which drops the `paths.target` edge while keeping the `After=` that the watcher wants, plus the `Conflicts=shutdown.target` / `Before=shutdown.target` the defaults would otherwise have provided.

## Verification

Installed the changed unit on a HALPI2 running `HaLOS Desktop Marine HALPI2 2026-08-08.0` (`halos-core-containers 0.7.0-1`) and rebooted. Before: `is-active` = `inactive`, `Found ordering cycle` in the boot journal. After:

```
$ journalctl -b | grep -i "ordering cycle"    # (none)
$ systemctl is-active halos-core-containers.service halos-oidc-clients-reload.path
active
active
$ journalctl -u halos-core-containers.service -b | grep Attaching
docker[4078]: Attaching to authelia, authelia-valkey, autoheal, ca-download, homarr, traefik
$ journalctl -u halos-manage-certs.service -b | tail -1
Finished halos-manage-certs.service - Provision and maintain HaLOS device CA + leaf TLS certificate.
$ curl -sk -o /dev/null -w '%{http_code}\n' https://localhost/
302
```

`halos-oidc-clients-reload.service` also completed on its own for the first time ("OIDC client registration already current - Authelia left running").

`VERSION` opens a new release cycle at 0.7.1 — the previous stable tag is `v0.7.0+1` and this PR changes package payload.


mairas:
# Code review — PR #209

**Scope:** `6358a4b..c714e58` (branch `fix/oidc-path-ordering-cycle`), 4 commits.
**Intent:** make the core stack start at boot again by removing the ordering cycle the OIDC snippet watcher introduced, without losing the watcher's "don't fire before the stack is up" ordering or its shutdown behavior.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on); reliability (the diff is boot/shutdown unit ordering); adversarial (the unit gates Authelia's OIDC registration reload, so a wrong fix breaks SSO silently).

## P1 — High

|---|---|---|---|---|---|
| 1 | `tests/test-packaging-contract.sh` | The fix shipped with no regression guard, in a repo that already pins exactly this class of silent packaging breakage. `systemd-analyze verify` cannot catch it — a cycle exists only in a job transaction — so a static assertion over the unit files is the practical guard. | testing, reliability | 1.00 | **fixed here** |

## P2 — Moderate

|---|---|---|---|---|---|
| 2 | `debian/…halos-oidc-clients-reload.path:13` | `DefaultDependencies=no` also drops `After=`/`Requires=sysinit.target`, not just the `paths.target` edge. The comment accounted for the shutdown pair only, so the sysinit ordering went silently. Harmless today solely because `WantedBy=halos-core-containers.service` is the unit's only activation route — an unstated dependency on a different unit's defaults. | maintainability, correctness, reliability, testing | 0.82 | **fixed here** |
| 3 | `container-packaging-tools` `systemd/path.j2:4` | The generator emits `PartOf=` + `After=<pkg>.service` with default dependencies — byte-for-byte the pre-fix shape. Latent only because no shipped app declares `file_watchers` yet. | correctness, adversarial, testing | 0.95 | pre-existing → halos-org/container-packaging-tools#248 |
| 4 | `prestart.sh:447` | First boot runs `authelia/authelia:4.39` for the admin hash while compose and `lib-oidc-clients.sh` use `4.39.19`; nothing pre-pulls it, and failure exits 1 out of `ExecStartPre`, producing the same operator-visible symptom as #208 from a different cause. | adversarial | 0.70 | pre-existing → #210 |
| 5 | `assets/reload-oidc-clients:119` | The merge loop hits `max_rounds=3`, warns, `break`s into `echo "OIDC clients reloaded successfully"` and exits 0. Snippets written after the last merge stay unregistered and nothing retries. Reachable by a bulk app install. | adversarial | 0.70 | pre-existing → #211 |

## P3 — Low

|---|---|---|---|---|---|
| 6 | `debian/…halos-oidc-clients-reload.path:11` | The comment narrated the incident in past tense and named `halos-manage-certs.service` as "the one it picked" — an arbitrary outcome encoded as mechanism. The repo's other units use the counterfactual voice. | maintainability | 0.66 | **fixed here** |
| 7 | `debian/…halos-oidc-clients-reload.path:15` | No record of why the blunt opt-out beat simply dropping the `After=`. This repo documents rejected alternatives in the unit itself ("Type=oneshot WITHOUT RemainAfterExit", "Deliberately NOT Persistent=true") so the next maintainer does not re-litigate. | maintainability | 0.62 | **fixed here** |
| 8 | `docs/solutions/` | The generalizable rule — any `.path` unit with `After=` on an ordinary service closes this cycle — lived only in the issue and one unit comment, where the person writing the next path unit will not see it. | maintainability, project-standards | 0.78 | **fixed here** |
| 9 | `debian/halos-core-containers.service:4` | A failed `Requires=` oneshot cancels the stack's start job; `Restart=on-failure` does not cover a cancelled job, and the certs timer has no edge back to the stack. Same silent signature as #208. | reliability | 0.75 | pre-existing → #212 |
| 10 | `debian/…halos-oidc-clients-reload.path:6` | `PartOf=` does not propagate systemd's own auto-restart, so during a stack crash-loop the watcher stays armed while prestart rewrites the snippet every 10 s, with `StartLimitIntervalSec=0` removing the brake. | adversarial | 0.65 | pre-existing → #211 |
| 11 | `.github/scripts/check-changelog-edits.sh:32` | The hook tells authors to run `./run bumpversion`, claiming it "Uses dch to update debian/changelog"; in this repo it only moves `VERSION`, and the changelog is CI-generated. | project-standards | 0.65 | pre-existing → #213 |

## Applied fixes

- **#2, #6, #7** — `After=sysinit.target` restored explicitly, and the comment recast: counterfactual voice, no dependence on which job systemd happened to delete, a full accounting of what the opt-out drops, and one sentence on why dropping the `After=` was not the chosen fix. Folded into `944ca32` rather than appended, so history never carries the incomplete version.
- **#1** — `tests/test-packaging-contract.sh` grew a third check: every `debian/*.path` with an `After=` on a `.service` must carry `DefaultDependencies=no`, `Conflicts=shutdown.target` and `Before=shutdown.target`. Verified as a real reproducing test — it fails against the pre-fix unit at `6358a4b` and passes on HEAD (11/11).
- **#8** — `docs/solutions/2026-08-10-path-unit-after-service-ordering-cycle.md`, with the general rule, the boot-journal diagnostic signature, and why `systemd-analyze verify` is the wrong tool for it.

Re-verified on hardware after the changes: reinstalled the revised unit on the HALPI2, rebooted, and confirmed `DefaultDependencies=no` with `After=sysinit.target halos-core-containers.service` as loaded, no `Found ordering cycle` in the boot journal, stack and watcher both active, all six containers up (`authelia, authelia-valkey, autoheal, ca-download, homarr, traefik`), and 302 on both `http://localhost/` and `https://localhost/`.

## Coverage

- No findings suppressed by the 0.60 confidence gate; all six reviewers returned.
- **Residual risk — recovery of already-broken devices.** Devices on 0.7.0-1 have no core stack; recovery depends on `apt` reaching them and on `deb-systemd-invoke restart` bringing the stack up without a reboot. The mechanism reads correct in the generated maintainer script but was not exercised — the test device was patched by hand, not upgraded from the broken state.
- **Untested: the restored shutdown ordering.** `Conflicts=`/`Before=shutdown.target` is hand-written and was exercised only by two reboots; nothing confirmed the watcher stops before `shutdown.target` rather than lingering to a job timeout.
- **Untested: cold first boot.** All verification ran on a warm device (images cached, secrets and RSA key present, digest stamp current). #210 lives on the cold path.
- **Lost-event window (pre-existing, unchanged):** the watch is armed after the stack goes active, and `PathModified` has no start-time catch-up, so a snippet write landing between the once-per-start reload and the watch is dropped until the next write.

## Verdict

**Ready to merge.** Every pre-existing finding routed to `file-issue` now has one (#210, #211, #212, #213, halos-org/container-packaging-tools#248); nothing actionable is left unowned.

Issue #210, opened 2026-08-10
First boot pulls a second Authelia image tag prestart alone needs


mairas:
`prestart.sh` hashes the initial admin password with a different Authelia image than the one the stack actually runs:

```
prestart.sh:447            docker run --rm authelia/authelia:4.39 authelia crypto hash generate argon2 ...
docker-compose.yml:41      image: authelia/authelia:4.39.19
assets/lib-oidc-clients.sh:26  HALOS_OIDC_AUTHELIA_IMAGE="authelia/authelia:4.39.19"
```

`4.39` appears nowhere else. Nothing pre-pulls images into the image build, so on a freshly flashed device first boot must fetch a second Authelia image that only this one line needs — over whatever link the boat has, before the stack can start.

The failure is not graceful. An empty `INITIAL_HASH` hits `exit 1` (prestart.sh:450-453), which fails `ExecStartPre`, which fails `halos-core-containers.service` — Traefik, Authelia and Homarr all down. That is the same operator-visible symptom as #208 from an unrelated cause, which makes the field diagnosis of either one harder.

The service also sets no `TimeoutStartSec`, so the whole start phase — 4096-bit `genrsa`, two image pulls, two container runs — lives inside systemd's 90 s default on a Raspberry Pi. On timeout the run is SIGTERMed mid-render and `_halos_oidc_render`'s `mktemp` file, which holds the HMAC secret and signing key, is left behind in Authelia's bind-mounted `/config` because the cleanup runs from a `RETURN` trap that never fires.

## Suggested direction

- Pin the admin-hash `docker run` to the same tag the rest of the stack uses (reuse `HALOS_OIDC_AUTHELIA_IMAGE` rather than a second literal), so first boot fetches one image instead of two.
- Give `halos-core-containers.service` an explicit `TimeoutStartSec` sized for a cold first boot.
- Verify on a freshly flashed image, not a warm device: the warm path has the images cached, the secrets present and the stamp current, so it exercises none of this.

mairas:
Verified on the cold first boot this issue asked for. #214 deliberately did not close it, because the fix had only been exercised on a warm device.

Device: HALPI2 (CM5 Lite), freshly flashed with `Halos-Desktop-Marine-HALPI2_2026-08-10.0`, the first image built after 0.7.1 reached `trixie-stable`.

```
$ head -1 /boot/issue.txt
HaLOS Desktop Marine HALPI2 2026-08-10.0
[...]

PR #214, opened 2026-08-10
Stop first boot from needing a second Authelia image

mairas:
First boot fetched a second Authelia image that only one line needed, and could be killed by systemd's default start timeout while doing it. Both failures land as "the whole web stack is down", the symptom that points nowhere.

Issue #210 is addressed here but deliberately left open — its stated verification is a cold boot on freshly flashed hardware, which this branch has not had.

## Changes

**The admin-password hash uses the pinned image.** `prestart.sh` hard-coded `authelia/authelia:4.39` while `docker-compose.yml` and `lib-oidc-clients.sh` both run `4.39.19`. The library already carries the constant and states the intent — "Pinned to the tag docker-compose.yml runs, so the device never has to pull, retain, or reach a registry for a second image" — and prestart sources that library, so it now uses `HALOS_OIDC_AUTHELIA_IMAGE`. Docker's own stderr is kept rather than discarded: without it an unreachable registry, a rate-limited pull and a stopped daemon all reduce to one generic line. The plaintext moves from argv to the environment, matching `_halos_oidc_hash_secret`.

**`TimeoutStartSec=600` on the stack.** `Type=simple` counts the unit started once `ExecStart` forks, so the timeout bounds `prestart.sh`. A first boot generates a 4096-bit signing key and runs Authelia's hashing container, pulling the image on a cold cache — none of which fits the 90 s default on a Pi.

**Temp renders survive neither a signal nor a later run.** `_halos_oidc_render` cleaned its `mktemp` file from a `RETURN` trap, which does not fire when the shell is killed — a shutdown mid-merge left a 0600 file holding the HMAC secret and the signing key in the directory bind-mounted as Authelia's `/config`. The trap now covers `INT TERM HUP`, and a sweep at the top of `halos_oidc_merge_clients` — under the lock, before the digest short-circuit — catches what SIGKILL and power loss leave behind. Temp renders get a `.tmp.` infix so the sweep pattern cannot reach the `.lock` or `.stamp` beside them.

## Verification

`tests/test-lib-oidc-clients.sh` grew four assertions, 39/39 green. Each was mutation-tested against the branch:

| Mutation | Caught by |
|---|---|
| sweep moved back inside `_halos_oidc_render` | `..._when_nothing_needs_rendering`, `..._when_hashing_fails` |
| sweep glob widened to `.*` (unlinks the merge lock) | `..._from_a_killed_run` + 2 pre-existing tests |
| `mktemp` template drifts away from `.tmp.` | `test_sweep_glob_matches_what_the_renderer_creates` |
| prestart names an Authelia tag directly again | `test_hashing_image_matches_the_compose_pin` |

The first three all passed before this round — the original test planted a filename matching the sweep's own glob and asserted only a count, so it could not see any of them.

Still unverified: the cold first boot itself (empty image cache, no secrets, no stamp). `TimeoutStartSec=600` is bracketed by "more than 90" on one side and nothing on the other; no timing exists for prestart on a Pi with a cold cache.

Note for whoever merges: #215 touches the same unit file. Its earlier `StartLimitIntervalSec=300` would have made this timeout defeat it — a 610 s attempt cycle never fills a 300 s burst window — but #215 now uses `RestartSteps`/`RestartMaxDelaySec` instead, which has no such coupling.

No `VERSION` bump: latest stable is `v0.7.0+1` and `VERSION` is already 0.7.1, so the cycle is open.

### DIFF
--- a/assets/lib-oidc-clients.sh
+++ b/assets/lib-oidc-clients.sh
@@ -206,10 +206,13 @@ $(echo -e "${redirect_uris}" | sed '/^$/d')
     fi
 
     local tmp_file
-    tmp_file=$(mktemp "${AUTHELIA_OIDC_FILE}.XXXXXX") || return 1
+    tmp_file=$(mktemp "${AUTHELIA_OIDC_FILE}.tmp.XXXXXX") || return 1
     # The temp file holds the HMAC secret and the signing key, and lives in the
-    # directory bind-mounted as Authelia's /config.
-    trap 'rm -f "$tmp_file"' RETURN
+    # directory bind-mounted as Authelia's /config. RETURN alone would leave it
+    # behind whenever the shell is killed rather than returning — a shutdown
+    # mid-merge, a start timeout — so the signals are trapped too. SIGKILL and
+    # power loss still escape; halos_oidc_merge_clients sweeps for those.
+    trap 'rm -f "$tmp_file"' RETURN INT TERM HUP
     chmod 600 "$tmp_file"
 
     # Writes are checked explicitly: this function runs as an `if` condition, so
@@ -289,6 +292,14 @@ halos_oidc_merge_clients() {
     local stamp_file="${AUTHELIA_OIDC_FILE}.stamp"
 
     _halos_oidc_lock
+
+    # Sweep temp renders a previous run could not clean up — killed by SIGKILL,
+    # or by power loss. They are mode 0600 but hold the HMAC secret and the
+    # signing key, in the directory bind-mounted as Authelia's /config. This
+    # runs under the lock and before the digest short-circuit below, so it
+    # covers every path out of the merge, including the ones that never render.
+    # The .tmp. infix keeps the pattern off .lock and .stamp.
+    rm -f "${AUTHELIA_OIDC_FILE}".tmp.*
 
     local input_digest output_digest
     input_digest="$(_halos_oidc_input_digest)"
--- a/debian/halos-core-containers.service
+++ b/debian/halos-core-containers.service
@@ -16,6 +16,12 @@ EnvironmentFile=-/run/container-apps/halos-core-containers/runtime.env
 EnvironmentFile=-/run/halos/domain.env
 ExecStart=docker compose up
 ExecStop=docker compose down
+# Type=simple counts the unit as started once ExecStart forks, so this bounds
+# prestart.sh. On a first boot prestart generates a 4096-bit OIDC signing key
+# and runs Authelia's hashing container, which on a cold image cache means a
+# registry fetch first — none of that fits systemd's 90 s default, and the
+# timeout kills prestart mid-render.
+TimeoutStartSec=600
 Restart=on-failure
 RestartSec=10
 StandardOutput=journal
--- a/docs/SSO_ARCHITECTURE.md
+++ b/docs/SSO_ARCHITECTURE.md
@@ -58,7 +58,7 @@ Traefik is the central routing component. It receives all incoming HTTP/HTTPS re
 
 ### Authelia Identity Provider
 
-**Image**: `authelia/authelia:4.39`
+**Image**: pinned in `docker-compose.yml` (and mirrored by `HALOS_OIDC_AUTHELIA_IMAGE` in `assets/lib-oidc-clients.sh`, which is what prestart and the reload tool hash with)
 **Network**: `halos-proxy-network` (external)
 **Ports**: None exposed (accessed via Traefik)
 
--- a/prestart.sh
+++ b/prestart.sh
@@ -444,11 +444,23 @@ echo "REDIS_PASSWORD=${REDIS_PASSWORD}" >> "${RUNTIME_ENV}"
 if [ ! -f "${AUTHELIA_DATA}/users_database.yml" ]; then
     echo "Creating initial admin user..."
     DEFAULT_PASSWORD="halos"
-    INITIAL_HASH=$(docker run --rm authelia/authelia:4.39 authelia crypto hash generate argon2 \
-        --password "${DEFAULT_PASSWORD}" 2>/dev/null | grep 'Digest:' | sed 's/Digest: //')
+    # Plaintext through the environment, not argv: /proc/<pid>/cmdline is
+    # world-readable, and this is the pattern _halos_oidc_hash_secret uses for
+    # the same CLI. Docker's own output is kept: without it an unreachable
+    # registry, a rate-limited pull and a stopped daemon all reduce to one
+    # generic line, with the whole stack down and nothing to attribute it to.
+    if ! HASH_OUTPUT=$(HALOS_ADMIN_PW="${DEFAULT_PASSWORD}" docker run --rm \
+        -e HALOS_ADMIN_PW "${HALOS_OIDC_AUTHELIA_IMAGE}" \
+        sh -c 'authelia crypto hash generate argon2 --password "$HALOS_ADMIN_PW"' 2>&1); then
+        echo "ERROR: could not generate the initial password hash" >&2
+        printf 'docker: %s\n' "${HASH_OUTPUT}" >&2
+        exit 1
+    fi
+    INITIAL_HASH=$(printf '%s\n' "${HASH_OUTPUT}" | grep 'Digest:' | sed 's/Digest: //')
 
     if [ -z "${INITIAL_HASH}" ]; then
-        echo "ERROR: Failed to generate password hash"
+        echo "ERROR: password hashing produced no digest" >&2
+        printf 'docker: %s\n' "${HASH_OUTPUT}" >&2
         exit 1
     fi
 
--- a/tests/test-lib-oidc-clients.sh
+++ b/tests/test-lib-oidc-clients.sh
@@ -238,6 +238,98 @@ test_merge_renders_clients() {
     assert_not_contains "$out" '${HALOS_DOMAIN}' "placeholder left unexpanded"
 }
 
+# Name a leftover the way _halos_oidc_render's mktemp template does, so the
+# fixture tracks the renderer rather than the sweep's own glob.
+plant_stale_render() {
+    local root="$1"
+    local stale="$(authelia_dir "$root")/oidc-clients.yml.tmp.AbC123"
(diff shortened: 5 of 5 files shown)


mairas:
# Code review — PR #214

**Scope:** `523031d..1a9a8f5`, 3 commits, 4 files.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on); security (the change moves secret-bearing temp files); reliability (first-boot and upgrade paths).

## P2 — Moderate

|---|---|---|---|---|---|
| 1 | `assets/lib-oidc-clients.sh:212` | **The sweep is unreachable on the paths that matter.** It sits inside `_halos_oidc_render`, *after* the `render_failed` return at :197, and `halos_oidc_merge_clients` short-circuits on a matching stamp before rendering at all. So: a render that cannot hash (registry unreachable — the exact #210 scenario) returns early and leaves the file; and on a settled device every unchanged merge skips the sweep entirely. A reviewer demonstrated both with probes against this branch. Moving the `rm -f` to just after `_halos_oidc_lock` in `halos_oidc_merge_clients` covers every path, and makes the comment's "callers hold the merge lock" self-evident instead of an invariant asserted from inside a callee. | correctness, security, reliability, testing, maintainability | 0.95 | **must fix** |
| 2 | `debian/halos-core-containers.service:24` | **Cross-PR with #215.** #215 adds `StartLimitIntervalSec=300`/`StartLimitBurst=6` to this unit. With `TimeoutStartSec=600` and `RestartSec=10` an attempt cycle is ~610 s, so no two attempts fall inside the 300 s window, the burst counter resets every time, and a hanging start loops forever in `activating` — the state #215 exists to eliminate. Its `NRestarts=6 → failed` measurement used a fast `exit 1`, which does not probe this. The two numbers have to be sized together: `StartLimitIntervalSec ≥ StartLimitBurst × (TimeoutStartSec + RestartSec)`. | correctness, reliability, testing | 0.85 | **must fix** |
| 3 | `tests/test-lib-oidc-clients.sh:241` | The new test plants a filename matching the *sweep's glob*, not one the renderer produces, and asserts only a count of files matching that same glob. Two mutations pass 35/35: changing the `mktemp` template to `.tmp.XXXXXX` (sweep no longer matches any real leaked file), and widening the sweep to `rm -f "${AUTHELIA_OIDC_FILE}".*` (which unlinks the merge `.lock` while the flock is held, so two renders can proceed concurrently). Assert the specific file is gone **and** that `.lock` and `.stamp` survive. | testing, security | 0.90 | **must fix** |
| 4 | `assets/lib-oidc-clients.sh:212` | The `.??????` glob claims every 6-character suffix beside `oidc-clients.yml`. It is safe today only because `.lock` is 4 and `.stamp` is 5 — an accident nobody wrote down. `.backup`, `.sha256`, `.digest`, `.render` are all 6. Deleting the stamp is fail-safe; deleting the lock is not. `mktemp "${AUTHELIA_OIDC_FILE}.tmp.XXXXXX"` swept by `.tmp.*` makes the namespace explicit. | security, maintainability | 0.72 | **fix** |
| 5 | `assets/lib-oidc-clients.sh:218` | The stated bug is that a `RETURN` trap does not fire on a signal — and the fix does not add that coverage, it cleans up on the *next* render. `trap 'rm -f "$tmp_file"' RETURN INT TERM HUP` removes the file at kill time regardless of what any later run does; keep the sweep as the SIGKILL/power-loss fallback rather than the only mechanism. | security | 0.75 | **fix** |
| 6 | (PR body) | `Fixes #210` auto-closes an issue whose own "Suggested direction" names cold-boot verification as part of the fix, and the PR body states that verification has not happened. Either run it before merge, or drop the closing keyword and keep #210 open. | project-standards | 0.72 | decide |

## P3 — Low

|---|---|---|---|---|---|
| 7 | `docs/SSO_ARCHITECTURE.md:61` | Still reads **Image**: `authelia/authelia:4.39` — the stale tag whose presence in prestart *is* this PR's bug. The PR's verification claim ("grep now reports one tag across compose, the library and prestart") is true for those three and misses the fourth copy. Verified on the branch. | correctness, testing, maintainability | 0.90 | **fix here** |
| 8 | `prestart.sh:448` | `2>/dev/null` on the line this PR edits discards the reason the hash failed, so a DNS failure, a rate-limited pull and a stopped daemon all reduce to "ERROR: Failed to generate password hash" with the whole stack down — the "points nowhere" symptom #208/#210/#212 are all about. `_halos_oidc_hash_secret` in the library this line now sources does the opposite: captures `2>&1` and prints `docker: %s`. | correctness, reliability, project-standards | 0.87 | **fix here** |
| 9 | `prestart.sh:447` | The plaintext goes on argv while the library's helper deliberately routes it through the environment, with the reason in a comment: `/proc/<pid>/cmdline` is world-readable. Harmless while the value is the fixed literal `halos`, and a live leak the moment that default is ever randomized (see #219). | security | 0.70 | fix |
| 10 | `prestart.sh:447` | Second hashing invocation left beside the library helper that already wraps this exact CLI with stderr capture, digest validation and env-passed plaintext. AGENTS.md's own contract: "two copies drift, and the drift stays invisible until a login breaks". The tag half was reconciled; the invocation was not. | project-standards | 0.60 | consider |
| 11 | `tests/test-packaging-contract.sh` | Nothing pins the one-tag invariant this PR restores by hand. A grep asserting a single distinct `authelia/authelia:` tag across compose + library, and no such literal in prestart at all, is the same shape as the checks already there. | testing, maintainability | 0.85 | fix |
| 12 | `debian/halos-core-containers.service:24` | The 10-minute window now also bounds `deb-systemd-invoke restart` during `apt upgrade`, which runs without `--no-block` — so a wedged prestart can hold the dpkg lock for 10 minutes. `debian/rules` already flags this hazard for the reload oneshot and works around it with `--no-start`; the main unit has no such guard. | correctness | 0.70 | note |
| 13 | `assets/lib-oidc-clients.sh:208` | The comment names the two moments the leak *cannot* happen in. `mktemp` runs after the whole hashing loop, so the temp file exists only between :215 and :246 — a chmod, an awk and a heredoc. A start-timeout kill lands in the docker loop, where no temp file exists yet. The sweep is still right; the recorded reason points the next reader at the wrong window. | maintainability, testing | 0.70 | fix wording |
| 14 | (commits) | Two subjects exceed the 50-char limit (64 and 56). | project-standards | 0.85 | amend |

## Coverage

- Nothing suppressed by the 0.60 gate; all six reviewers returned.
- The tag substitution itself checks out: `4.39.19` has the same `authelia crypto hash generate` CLI and `Digest:` output shape, and pinning makes the generator and the verifier the same build. `HALOS_OIDC_AUTHELIA_IMAGE` is genuinely in scope at the use site (library sourced at `prestart.sh:52`, constant assigned unconditionally, nothing reassigns it).
- `TimeoutStartSec=600` is unmeasured — no timing exists for a cold first boot's prestart on a Pi, so 600 is bracketed by "more than 90" on one side and nothing on the other.
- `TimeoutStartSec` bounds prestart only. `Type=simple` marks the unit started once `docker compose up` forks, so the five remaining image pulls happen outside any timeout, with the unit already `active`.
- Pre-existing security findings surfaced here, now filed: #217 (Homarr's whole-tree mount reads Authelia's signing key — which undercuts the "0600 in Authelia's config dir" reasoning in the new comment), #218 (first-boot secrets created 0644, env file never chmod'd), #219 (shared default admin password).

## Verdict

**Not ready.** Finding 1 is the blocker: the sweep does not run on the failure path it was written for, so the change does not do what the PR says. 3 and 4 mean the test would not notice either that or a glob that eats the merge lock. 2 has to be settled jointly with #215. 7 and 8 are small and belong in this PR.

mairas:
## Review findings addressed

| # | Finding | Resolution |
|---|---|---|
| 1 | P2 — sweep unreachable on the paths that matter | Moved to the top of `halos_oidc_merge_clients`, under the lock and before the digest short-circuit. Two new tests cover the paths it used to skip; both fail if it moves back |
| 2 | P2 — cross-PR with #215's start limit | Resolved in #215, which replaced the limit with `RestartSteps`/`RestartMaxDelaySec` — no rate-limit window left to interact with |
| 3 | P2 — test asserts a count of files matching the sweep's own glob | Now asserts the specific file is gone, that `.lock` and `.stamp` survive, and that the render still lands. Mutation-checked: widening the glob to `.*` fails 3 tests |
| 4 | P2 — `.??????` claims any 6-char suffix | Temp renders get a `.tmp.` infix; the sweep matches `.tmp.*`. A new test pins the glob against the `mktemp` template, since a hand-named fixture cannot see them drift |
| 5 | P2 — sweep does not fix the trap, only cleans up later | `trap ... RETURN INT TERM HUP`. The sweep stays as the SIGKILL / power-loss fallback, which is what its comment now says |
| 6 | P2 — `Fixes #210` closes an issue whose verification has not run | Closing keyword dropped. `gh pr view 214 --json closingIssuesReferences` returns `[]`; #210 stays open for the cold boot |
| 7 | P3 — `docs/SSO_ARCHITECTURE.md` still says 4.39 | Now points at the compose pin instead of restating a tag, so there is no fourth copy to drift |
| 8 | P3 — `2>/dev/null` discards why the hash failed | Captures and prints docker's output on both failure paths |
| 9 | P3 — plaintext on argv | Through the environment now, matching `_halos_oidc_hash_secret` |
| 11 | P3 — nothing pins the one-tag invariant | `test_hashing_image_matches_the_compose_pin`: the library constant must equal compose's tag, and prestart must not name a tag at all |
| 13 | P3 — comment names the wrong window | Rewritten: the leak is any signal between `mktemp` and `mv` |
| 14 | P3 — commit subjects over 50 chars | All four now ≤50 |
| 10 | P3 — second hashing invocation beside the library helper | **Not done.** The helper hard-codes pbkdf2 for client secrets while this path needs argon2 for `users_database.yml`, so sharing it means a parameter and a wider change than this PR. The drift that actually bit (#210) is now pinned by a test |
| 12 | P3 — 10-min window can hold the dpkg lock | **Not fixed**, noted. `debian/rules` uses `--no-start` for the reload oneshot for this reason; doing the same for the main unit changes upgrade behaviour and wants its own PR |

Pre-existing security findings from this review are filed as #217, #218 and #219 rather than folded in.

PR #215, opened 2026-08-10
Do not let a failed input unit silently cancel the core stack

### LINKED ISSUE #212: A failed oneshot dependency leaves the core stack down with no retry and an empty journal
`halos-core-containers.service` hard-requires two oneshots:

```
Requires=docker.service halos-manage-certs.service halos-resolve-domain.service
```

If either oneshot exits non-zero at boot, systemd cancels the stack's start job with result `dependency`. `Restart=on-failure` on the stack does not cover that — it applies to a unit whose own start was attempted, not to a start job that was cancelled — so the stack sits inactive with an empty `journalctl -u halos-core-containers.service -b`.

That is the same silent signature as #208: no web stack, and nothing in the stack's own journal saying why. Both oneshots are `set -e` shell scripts doing real work early in boot (CA and leaf cert provisioning via openssl; hostname resolution ordered only `After=local-fs.target`, i.e. potentially before the network settles), so a transient failure is not exotic.

Nothing recovers it either. `halos-manage-certs.timer` re-runs the certs oneshot at `OnBootSec=15min`, but the timer has `Unit=halos-manage-certs.service` and no edge back to the stack, so a successful late run leaves the stack down until someone starts it by hand or reboots.

## Suggested direction

- Bounded `Restart=on-failure` + `RestartSec` on the two oneshots so a transient failure self-heals during boot, and/or an edge from the certs timer back to the stack so a later successful run brings it up.
- `OnFailure=` on the oneshots so a stack blocked this way surfaces somewhere other than an empty journal.

Found during the review of #209; not caused by it.

mairas:
A failed `Requires=` dependency does not fail the core stack — it deletes its start job. The unit never runs `ExecStart`, ends `inactive` with `Result=success`, and stays out of `systemctl --failed` while Traefik, Authelia and Homarr are all absent. Same signature as #208 and, like #208, points nowhere.

Fixes #212

## What was measured

Scratch units in `/run/systemd/system` on a HALPI2 (systemd 257), no packaged file touched:

| Experiment | Result |
|---|---|
| consumer `Requires=` a oneshot that fails | consumer `inactive`, `Result=success`, `ExecMainStartTimestamp` empty — never ran |
| same, but the oneshot has `Restart=on-failure` and succeeds on attempt 3 | oneshot reached success; consumer still never ran |
| consumer `Wants=` + `After=` a oneshot that fails | consumer `active`, started 1.8 ms after the dependency finished, failed dependency visible in `systemctl --failed` |

Row two is why adding retries to the dependency is not the fix on its own: once a start job is cancelled, nothing re-queues it.

## Changes

**`Wants=` for the two oneshots, `Requires=` only for docker.** Ordering unchanged. docker is the runtime; the oneshots produce inputs.

**prestart refuses an empty `HALOS_DOMAIN`,** because `Wants=` alone would trade a silent absence for something worse: `prestart.sh` runs `set -e` without `set -u`, so a missing domain renders an empty string into the OIDC issuer, Authelia's cookie domain and Traefik's redirects — a stack that looks healthy and whose every login fails. The guard checks the resolver's file, not the ambient variable, so a pre-upgrade `HALOS_DOMAIN` left in `runtime.env` cannot satisfy it.

**Backoff instead of a start limit.** An earlier revision capped restarts so an unbootable stack would reach `failed` and show up in `systemctl --failed`. Two problems, both raised in review: nothing re-queues a start-limit `failed` unit, so any cause lasting more than ~75 s became permanent on a headless device; and the window interacts with #214's `TimeoutStartSec=600` — systemd's rate limiter resets when an attempt arrives after the interval, so a 610 s attempt cycle never fills a 300 s burst and a *hanging* start would loop forever anyway. `RestartSteps=6` / `RestartMaxDelaySec=300` retries indefinitely at 10 s doubling to 5 min: self-healing, no 10 s spin, no coupling to the timeout.

**The resolver retries itself,** so the guard's recovery claim is true. An auto-restart of the consumer does not re-pull its dependencies — the resolver unit documents that — so without its own `Restart=` a failed resolve would leave the stack refusing forever.

**The adoption sentinel is guaranteed before compose.** `docker-compose.yml` bind-mounts it as a *file* and names the cert manager as the guarantor — a guarantee that held only while the stack required that unit. With `Wants=`, a failed cert run would let Docker create a directory there; `lib-ca.sh` repairs the host path later while the running sidecar keeps the removed inode, so adoption is never recorded and the CN-refresh gate can orphan a trust anchor the operator already installed. prestart now makes the same call the cert manager does, which preserves an existing sentinel verbatim.

Three sibling units still explained their own design in terms of the removed `Requires=`; all updated. `tests/test-packaging-contract.sh` pins the new dependency shape (its existing loops glob `debian/<pkg>.*.{service,timer,path}` and never see the main unit, so the check names it). `docs/solutions/` gets the measurement table.

## Verification on hardware

Transient failure, the case that matters — a drop-in override made the resolver fail twice then succeed, no packaged file touched:

```
11:54:19 prestart.sh: HALOS_DOMAIN_UNSET: halos-resolve-domain.service has not published /run/halos/domain.env
11:54:29 x-flaky-resolver: attempt 3, running the real resolver
11:54:29 halos-resolve-domain: HALOS_DOMAIN=halosdev.local
11:54:29 prestart.sh: Domain: halosdev.local
11:54:30 docker: Attaching to authelia, authelia-valkey, autoheal, ca-download, homarr, traefik
```

`NRestarts=1`, `ActiveState=active`, 302 from `https://localhost/`. No operator involved. Harness removed and rebooted clean afterwards: stack up, adoption sentinel still a regular file reading `pending` (preserved, not rewritten), only the unrelated `configure-ublox-marine` failure remains.

## What this does not fix

A failed `halos-manage-certs` now lets the stack start with whatever TLS material exists, and recovery depends on its timer — `OnBootSec=15min` **plus `RandomizedDelaySec=1h`**, so 15–75 minutes. Nobody has observed what Traefik serves in that window. Neither oneshot sets `TimeoutStartSec`, so a *hung* dependency still blocks the stack indefinitely — a different failure from the one fixed here.

No `VERSION` bump: latest stable is `v0.7.0+1` and `VERSION` is already 0.7.1.

### DIFF
--- a/debian/halos-core-containers.halos-manage-certs.service
+++ b/debian/halos-core-containers.halos-manage-certs.service
@@ -8,11 +8,12 @@ Before=halos-core-containers.service
 # Type=oneshot WITHOUT RemainAfterExit so the timer's `start` action
 # actually re-runs ExecStart on each fire. systemd's `start` on a
 # RemainAfterExit=yes oneshot in active(exited) state is a documented
-# no-op, which would silently break periodic renewal. The Requires= +
-# After= chain on halos-core-containers.service still gates correctly:
-# a successfully-exited oneshot satisfies Requires= even after it goes
-# naturally inactive (verified empirically — only failed exits drop
-# the consumer).
+# no-op, which would silently break periodic renewal. The ordering
+# against halos-core-containers.service still holds: a successfully-exited
+# oneshot satisfies After= even after it goes naturally inactive. The stack
+# only Wants= this unit, though, so a failed run no longer holds the stack
+# back — it starts with whatever TLS material exists and the timer below is
+# what repairs it (#212).
 Type=oneshot
 ExecStart=/usr/lib/halos-core-containers/halos-manage-certs
 StandardOutput=journal
--- a/debian/halos-core-containers.halos-manage-certs.timer
+++ b/debian/halos-core-containers.halos-manage-certs.timer
@@ -11,14 +11,16 @@ OnBootSec=15min
 # Re-check every 24h. Well below HALOS_CA_LEAF_RENEW_THRESHOLD_DAYS (60d),
 # so even fortnightly-rebooted devices get multiple chances to renew.
 OnUnitActiveSec=24h
-# Spread fleet-wide hits over a 1h window to avoid synchronized renewal
-# storms.
+# Spread the hits from every device with this package over a 1h window to
+# avoid synchronized renewal storms. Note this delays the first post-boot
+# check to 15-75 min, which is also how long a boot-time cert failure goes
+# unrepaired.
 RandomizedDelaySec=1h
 # Deliberately NOT Persistent=true: Raspberry Pi has no RTC, so first
 # boot starts with a fake-hwclock timestamp (often 1970). With Persistent
 # set, the post-NTP clock jump from 1970 → 2026 would trigger a catch-up
 # fire storm of "missed" runs. Boot-time renewal still happens because
-# halos-core-containers.service Requires= halos-manage-certs.service —
+# halos-core-containers.service pulls this unit in via Wants= —
 # losing Persistent only means a device powered off across a scheduled
 # fire waits for the next OnUnitActiveSec=24h tick (≤24h post-boot) to
 # re-check, which is well below the 60-day renewal threshold.
--- a/debian/halos-core-containers.halos-resolve-domain.service
+++ b/debian/halos-core-containers.halos-resolve-domain.service
@@ -7,14 +7,22 @@ Before=halos-core-containers.service
 [Service]
 # Type=oneshot WITHOUT RemainAfterExit so an explicit consumer (re)start
 # re-runs the resolver and republishes /run/halos/domain.env, picking up admin
-# edits to /etc/halos/hostnames.conf. A consumer's Requires= pulls in this
-# (inactive) oneshot when the consumer is started; the successful exit
-# satisfies Requires= even after the unit goes naturally inactive. (A systemd
-# Restart=on-failure auto-restart of the consumer does not re-pull Requires=
-# deps, so it does not re-run this — domain.env simply retains its last value.)
+# edits to /etc/halos/hostnames.conf. A consumer's Wants= pulls in this
+# (inactive) oneshot when the consumer is started, and a successfully-exited
+# oneshot still satisfies the ordering after it goes naturally inactive. (A
+# systemd auto-restart of the consumer does not re-pull its dependencies, so it
+# does not re-run this — domain.env simply retains its last value.)
 # Same reasoning as halos-manage-certs.service.
 Type=oneshot
 ExecStart=/usr/lib/halos-core-containers/halos-resolve-domain
+# The consumer refuses to start without the domain this publishes, and an
+# auto-restart of the consumer will not re-run this unit. Retry here instead, so
+# a transient failure heals on its own: the resolver republishes, and the
+# consumer's next backed-off retry finds the file.
+Restart=on-failure
+RestartSec=30
+RestartSteps=4
+RestartMaxDelaySec=300
 StandardOutput=journal
 StandardError=journal
 
--- a/debian/halos-core-containers.service
+++ b/debian/halos-core-containers.service
@@ -1,7 +1,18 @@
 [Unit]
 Description=HaLOS Core Containers (Traefik, Authelia, Homarr)
 After=docker.service halos-manage-certs.service halos-resolve-domain.service
-Requires=docker.service halos-manage-certs.service halos-resolve-domain.service
+# docker is the runtime -- without it there is nothing to run, so it stays a
+# hard requirement. The two oneshots produce inputs instead, and Requires= on
+# them is a trap: a failed dependency cancels this unit's start job, so the unit
+# never runs ExecStart and ends inactive with Result=success, invisible to
+# `systemctl --failed`, while the whole web stack is absent. Nothing re-queues
+# it either -- measured: a dependency with Restart=on-failure that retries to
+# success does not bring the cancelled consumer back (#212). Wants= keeps the
+# ordering, lets the stack start, and leaves the failure visible on the unit
+# that actually failed; prestart refuses to run without the domain the resolver
+# publishes, so a genuinely missing input still fails loudly here.
+Requires=docker.service
+Wants=halos-manage-certs.service halos-resolve-domain.service
 
 [Service]
 Type=simple
@@ -17,7 +28,16 @@ EnvironmentFile=-/run/halos/domain.env
 ExecStart=docker compose up
 ExecStop=docker compose down
 Restart=on-failure
+# Back the retries off rather than capping them. A start limit would put the
+# unit in `failed` and leave it there -- nothing here re-queues it, so a cause
+# that clears later (the network arrives, the docker daemon finishes
+# restarting) would need a human, on a device that is usually headless. Retry
+# forever instead, but stop hammering: 10 s, doubling to a 5 min ceiling, so a
+# stack that cannot start logs its reason a few times an hour rather than every
+# 10 s, and comes up on its own the moment the cause goes away.
 RestartSec=10
+RestartSteps=6
+RestartMaxDelaySec=300
 StandardOutput=journal
 StandardError=journal
 
--- a/docs/solutions/2026-08-10-requires-on-a-failable-unit-cancels-the-consumer.md
+++ b/docs/solutions/2026-08-10-requires-on-a-failable-unit-cancels-the-consumer.md
@@ -1,0 +1,80 @@
+---
+title: "Requires= on a unit that can fail cancels the consumer's start job"
+date: 2026-08-10
+repo: halos-core-containers
+pr: https://github.com/halos-org/halos-core-containers/pull/215
+tags: [systemd, requires, wants, oneshot, restart, backoff, boot, gotcha, knowledge]
+---
+
+# Context
+
+`halos-core-containers.service` hard-required two `Type=oneshot` units that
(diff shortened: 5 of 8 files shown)


mairas:
# Code review — PR #215

**Scope:** `523031d..983cb1f`, 1 commit, 2 files.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on); reliability (boot/shutdown unit semantics); adversarial (a wrong fix here breaks SSO silently).

## P1 — Critical

|---|---|---|---|---|---|
| 1 | `debian/halos-core-containers.service:21` | The start limit converts every failure lasting more than ~75 s into a **permanent** outage. Before this PR the stack retried every 10 s forever, so a cause that cleared later (network arrived, docker daemon finished restarting) self-healed. Now the unit reaches `failed` and nothing re-queues it — no `OnFailure=`, no timer, no path edge. The PR argues the visibility half of the trade and not the permanence half. | adversarial, reliability, correctness | 0.92 | **must fix** |

## P2 — Moderate

|---|---|---|---|---|---|
| 2 | `debian/halos-core-containers.service:24` (with #214) | **Cross-PR:** #214 adds `TimeoutStartSec=600` to this same unit. systemd's rate limiter resets its counter when an attempt arrives more than `interval` after the window opened, so with a 610 s attempt cycle no two attempts ever land inside the 300 s window, the burst never fills, and a *hanging* start retries forever in `activating` — the exact state this PR exists to remove. The `NRestarts=6 → failed` measurement was taken with a fast `exit 1`, the one shape the limit still bounds. | correctness(214), reliability(214), testing(214), project-standards | 0.85 | **must fix** |
| 3 | `prestart.sh:65` | The guard comment claims "Restart=on-failure retries, and the retry succeeds as soon as the resolver publishes a domain." Nothing publishes it. `halos-resolve-domain.service`'s own comment records the measurement: "A systemd Restart=on-failure auto-restart of the consumer does not re-pull Requires= deps, so it does not re-run this." All six retries re-read the same absent `/run/halos/domain.env`. The PR body says this correctly; the comment is what survives in the tree. | maintainability, correctness, reliability, adversarial | 0.88 | **must fix** |
| 4 | `debian/halos-core-containers.halos-manage-certs.service:11`, `.timer:21`, `.halos-resolve-domain.service:10` | Three sibling units still explain their own design in terms of the `Requires=` this PR removes — and one is load-bearing: the timer justifies dropping `Persistent=true` because "Boot-time renewal still happens because halos-core-containers.service Requires= halos-manage-certs.service". Verified: all three comments present on the branch. | correctness, testing, project-standards, maintainability | 0.90 | **must fix** |
| 5 | `docker-compose.yml:222` | With `Wants=`, a certs run that fails before `halos_ca_adoption_init` lets compose create a **directory** at the `ca-download` adoption bind-mount source. Compose's own comment names the removed guarantee: "The cert-manager (halos-manage-certs.service, ordered Before= this stack) creates the file so Docker finds a regular file to bind here." `lib-ca.sh` repairs the host side later, but the running container keeps the orphaned inode, so adoption is never recorded and the CN-refresh gate can orphan a trust anchor the operator already installed. | reliability, adversarial, correctness | 0.80 | **must fix** |
| 6 | `debian/halos-core-containers.service:22` | Six starts per 300 s is reachable by *documented operator workflows*: `README.md:53`, `docs/HOSTNAMES.md:14` and `assets/hostnames.conf:32` all prescribe `systemctl restart halos-core-containers.service` after editing the hostname list, and `./run deploy` restarts on every deploy. The seventh restart is refused with the stack already torn down, and recovery needs `systemctl reset-failed`, which none of those docs mention. | adversarial | 0.75 | **must fix** |
| 7 | `tests/` | Nothing pins any of the three changes. `tests/test-packaging-contract.sh` is the established home (a sibling PR just added the #208 invariant there) but its glob is `debian/${PKG}.*.{service,timer,path}`, which **skips the main unit** — a check has to name the file explicitly. | testing, reliability, project-standards, maintainability | 0.95 | **must fix** |

## P3 — Low

|---|---|---|---|---|---|
| 8 | `prestart.sh:67` | The guard tests the ambient variable, not the producer's output. `/run` is tmpfs so a stale `domain.env` cannot survive a reboot — but `runtime.env` survives an in-place upgrade and older versions wrote `HALOS_DOMAIN` into it, so on an upgrade restart with a failing resolver the guard passes on a pre-upgrade value. Checking `/run/halos/domain.env` directly closes it. | adversarial | 0.60 | fix or note |
| 9 | `debian/halos-core-containers.service:21` | The bound only catches *fast* failures. Six starts must fit in 300 s, so a cycle averaging over 50 s (first-boot `genrsa 4096`, compose blocking on Authelia's healthcheck) rolls the window instead of tripping it. Same root cause as #2. | adversarial | 0.62 | fix with #2 |
| 10 | `debian/…halos-manage-certs.service`, `.halos-resolve-domain.service` | Neither oneshot sets `TimeoutStartSec`, and `Type=oneshot` disables the start timeout by default — so a *hung* dependency still blocks the stack forever with no journal line. `halos-oidc-clients-reload.service` already documents and guards this exact hazard. Pre-existing; this PR fixes the failed case and leaves the hung case. | reliability | 0.70 | follow-up |
| 11 | `debian/halos-core-containers.service:16` | `container-packaging-tools`' `service.j2` — the generated twin of this unit — already uses `Requires=docker.service` + `Wants=halos-resolve-domain.service` with `StartLimitIntervalSec=600`/`StartLimitBurst=5`, and its comment carries the operational note this one omits: a start-limit `failed` unit does not self-heal, `reset-failed` then `start`, and an upgrade will not clear it. Two budgets for one mechanism on one device. | maintainability | 0.75 | align |
| 12 | `docs/solutions/` | The three hardware measurements are the reason the fix takes this shape and live only in the PR body. #209 shipped its structurally identical finding as a `docs/solutions/` entry the same day; `(#208)` in a unit comment is backed by a document, `(#212)` by nothing. | maintainability, project-standards | 0.78 | fix |
| 13 | `debian/…halos-manage-certs.timer:14` | "Spread fleet-wide hits" — "fleet" for "all deployed devices" is banned in this codebase; in a marine product it reads as vessel-fleet management. Pre-existing, same unit family. | project-standards | 0.78 | fly-by |

## Coverage

- Nothing suppressed by the 0.60 gate; all six reviewers returned.
- Untested and now reachable: the stack starting with `halos-manage-certs` failed. Nobody has observed what Traefik serves with `/certs` empty, or whether the timer's later `touch` of `tls-default.yml` recovers it *from that starting point*. The timer's first fire is `OnBootSec=15min` **plus `RandomizedDelaySec=1h`** — 15–75 minutes, which the PR body's "15min" understates.
- Untested: upgrade with the unit already rate-limited (`deb-systemd-invoke restart` is refused and the `|| true` swallows it), docker-daemon restart under the new limit, and the unclean-reboot path where `restart: unless-stopped` restores the previous boot's containers while the unit reports `failed`.
- Pre-existing security findings surfaced here, now filed: #217, #218, #219.

## Verdict

**Not ready.** Findings 1 and 2 mean the unit as drafted can be *worse* than what it replaces: permanent instead of self-healing for fast failures, and still-infinite for slow ones once #214 lands. Findings 3–5 are wrong-or-stale statements and a real regression in the certs path. Suggested order: settle the start-limit/timeout arithmetic across both PRs (1, 2, 9), fix the comment and the sibling comments (3, 4), guard the adoption sentinel in prestart (5), then the contract test (7).

mairas:
## Review findings addressed

| # | Finding | Resolution |
|---|---|---|
| 1 | P1 — start limit makes failure permanent | **Dropped the limit.** `RestartSteps=6` / `RestartMaxDelaySec=300` retries forever at 10 s doubling to 5 min. Self-healing and no 10 s spin. Verified the directives parse on the target: systemd 257 reports `RestartSteps=6`, `RestartMaxDelayUSec=5min` |
| 2 | P2 — cross-PR with #214's `TimeoutStartSec=600` | Dissolved by the same change: backoff has no rate-limit window to interact with |
| 3 | P2 — comment claims a retry that cannot succeed | **Made it true** rather than rewording it. `halos-resolve-domain.service` now has its own `Restart=on-failure` with backoff, so it republishes and the stack's next retry picks it up |
| 4 | P2 — three sibling comments still assert `Requires=` | All three updated (certs service, certs timer, resolver) |
| 5 | P2 — adoption sentinel can become a directory | prestart calls `halos_ca_adoption_init` before compose, preserving an existing sentinel verbatim. `tests/test-prestart-loader-contract.sh` extended for the third loader — it caught the new symbol on the first CI run, which is what it exists for |
| 6 | P2 — operator restart lockout | Gone with the start limit |
| 7 | P2 — nothing pins the dependency shape | `tests/test-packaging-contract.sh` check 4, naming the main unit explicitly since the existing loops skip it. Fails if `Requires=` returns |
| 9 | P3 — start-limit only bounds fast failures | Gone with the start limit |
| 12 | P3 — no `docs/solutions/` entry | Added, with the experiment table and the start-limit reasoning |
| 13 | P3 — "fleet-wide" | Reworded; also notes the 15–75 min first-fire window |
| 8 | P3 — stale `runtime.env` satisfies the guard | Guard now checks `/run/halos/domain.env` itself |
| 10 | P3 — hung oneshot still blocks | **Not fixed.** Different failure from the one this PR is about; neither oneshot sets `TimeoutStartSec`. Worth its own issue |
| 11 | P3 — diverges from `service.j2`'s 600/5 | Moot — that budget was a start limit; this unit no longer has one. The generated template still does, which is a real question for `container-packaging-tools` |

The self-healing path is now verified end to end on hardware, which the earlier revision only asserted: a drop-in made the resolver fail twice, it retried itself, and the stack came up 10 s later with no operator. Details in the updated description.

PR #216, opened 2026-08-10
Describe this repo's actual changelog scheme in the hook and AGENTS.md

### LINKED ISSUE #213: Changelog pre-commit hook points at a dch step ./run bumpversion does not perform
The pre-commit hook that blocks `debian/changelog` edits tells the author to run a command that does not do what the message says:

```
.github/scripts/check-changelog-edits.sh:30-33
  Solution: Use './run bumpversion [patch|minor|major]' which:
    1. Updates VERSION file
    2. Uses dch to update debian/changelog with correct dates
    3. Commits the changes
```

In this repo `./run bumpversion` runs `bump2version` and nothing else, and `.bumpversion.cfg` lists a single target (`[bumpversion:file:VERSION]`). No `dch`, no changelog edit. The changelog that ships is generated at build time by `shared-workflows/.github/workflows/build-release.yml` via `.github/scripts/generate-changelog.sh`; the tracked `debian/changelog` has read `halos-core-containers (0.2.1-1)` since January.

The block itself is correct — hand-edits should not happen here — but an author who trips it is sent to a command that leaves the changelog untouched, and is left unsure whether they missed a step.

The workspace `AGENTS.md` says "Per-repo changelog schemes differ (committed vs CI-generated `debian/changelog`); see each repo's AGENTS.md", and this repo's AGENTS.md has no changelog section to see.

## Suggested direction

- Reword the hook message to state this repo's actual scheme: the changelog is CI-generated, do not hand-edit, and `./run bumpversion` only moves `VERSION`.
- Add a one-line changelog note to the Version Management section of this repo's `AGENTS.md`, which is where the workspace doc points readers.

Found during the review of #209.

mairas:
The pre-commit hook that blocks `debian/changelog` edits told authors to fix it with a command that does not touch the changelog.

Fixes #213

`./run bumpversion` here runs `bump2version` against a config with one target (`[bumpversion:file:VERSION]`) — no `dch`, no changelog edit. The entry a release ships is written at build time by `.github/scripts/generate-changelog.sh`; the tracked file has read `halos-core-containers (0.2.1-1)` since January.

## Changes

- The hook's "Why" and "Solution" text now describes the actual scheme, and gives the command for the case the author is actually in (`git restore --staged --worktree debian/changelog`) rather than prose.
- **`lefthook.yml`'s `fail_text` too.** lefthook prints it *after* the script's output, and it still read "Use './run bumpversion' instead." — so the corrected explanation was followed immediately by the uncorrected one, which is the last line an author reads. It now matches the sibling hook's form: "See error above for details."
- **The claim is scoped.** The first draft said an edit here "changes nothing a device ever sees". That is false for local builds: `tools/build-all.sh` regenerates only `if [ ! -f debian/changelog ]`, so `./run build` stamps its `.deb` from the tracked placeholder — the mechanism behind `docs/solutions/2026-05-31-deploy-downgrade-evicts-dependent-packages.md`, whose own mitigation is a hand edit. Both the hook text and the AGENTS.md section now say "a released package" and point at that document, and the bypass note names the case where overriding is right.
- AGENTS.md gains the Changelog section the workspace doc defers to.

## Verification

Exercised through a real blocked `git commit` with hooks installed — the path the first round missed by running the script directly, which never renders `fail_text`. Full output ends with the corrected line, exit 1. Clean tree exits 0. `bash -n` clean. The staged edit was reverted; the changelog is untouched by this branch.

## Follow-ups filed

- halos-org/halos#137 — the workspace `AGENTS.md` states the `dch` rule unconditionally; it holds for four repos and not for four others.
- #220 — untrack `debian/changelog` so local builds stop carrying a stale version, which would retire the hook and both blocks of documentation; plus pinning the tooling claims this text now makes.

No `VERSION` bump: a CI script and a docs file, neither package-affecting.

### DIFF
--- a/.github/scripts/check-changelog-edits.sh
+++ b/.github/scripts/check-changelog-edits.sh
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
 #
 # Pre-commit hook to prevent direct debian/changelog edits.
-# Changelog updates must go through ./run bumpversion which uses dch
-# to ensure proper RFC 2822 date formatting.
+# The changelog a *release* ships is generated at build time by
+# .github/scripts/generate-changelog.sh, so a hand-edit here never reaches a
+# released .deb. It does still version local ./run build packages, which is how
+# a dev build came to deploy as a downgrade (docs/solutions/2026-05-31-...).
 #
 # Bypass: SKIP_CHANGELOG_CHECK=1 git commit ...
 
@@ -10,7 +12,7 @@ set -o errexit
 set -o pipefail
 set -o nounset
 
-# Allow bypass for ./run bumpversion (via env var)
+# Manual override.
 if [[ "${SKIP_CHANGELOG_CHECK:-}" == "1" ]]; then
     exit 0
 fi
@@ -24,15 +26,21 @@ if [[ -n "$CHANGELOG_FILES" ]]; then
     echo "Staged changelog files:"
     echo "$CHANGELOG_FILES" | sed 's/^/  /'
     echo ""
-    echo "Why: Manual changelog edits often have RFC 2822 date formatting errors"
-    echo "(wrong weekday for date). The dch tool handles this correctly."
+    echo "Why: the changelog a release ships is written at build time by"
+    echo ".github/scripts/generate-changelog.sh, so editing the tracked file"
+    echo "changes nothing a released package carries. It does set the version of"
+    echo "local './run build' .debs, which is how a dev build once deployed as a"
+    echo "downgrade and evicted its dependents."
     echo ""
-    echo "Solution: Use './run bumpversion [patch|minor|major]' which:"
-    echo "  1. Updates VERSION file"
-    echo "  2. Uses dch to update debian/changelog with correct dates"
-    echo "  3. Commits the changes"
+    echo "Solution: drop the edit."
+    echo "  git restore --staged --worktree debian/changelog"
     echo ""
-    echo "If you need to bypass this check (e.g., fixing a changelog):"
+    echo "To open a new release cycle instead, run"
+    echo "'./run bumpversion [patch|minor|major]', which bumps VERSION and"
+    echo "commits that change -- it does not touch the changelog."
+    echo ""
+    echo "To override anyway (e.g. deliberately raising a local build's version"
+    echo "so it installs over a released one):"
     echo "  SKIP_CHANGELOG_CHECK=1 git commit ..."
     echo ""
     exit 1
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -35,6 +35,20 @@ The `VERSION` file is a **meta-version for git tags only**. It does NOT control 
 - Controls when new releases are triggered
 - Bump this when you want a new release bundle
 
+### Changelog
+
+`debian/changelog` is **CI-generated**, not maintained by hand. The entry a
+release ships is written at build time by
+`.github/scripts/generate-changelog.sh`, and a pre-commit hook rejects edits to
+the tracked file. `./run bumpversion` bumps `VERSION` and commits it — it does
+not run `dch`, unlike the repos the workspace changelog policy describes.
+
+The tracked file is not inert, though: `tools/build-all.sh` regenerates it only
+when it is absent, so a local `./run build` stamps its `.deb` from the committed
+placeholder. That is why a dev build installs as a downgrade and can evict its
+dependents — see
+[docs/solutions/2026-05-31-deploy-downgrade-evicts-dependent-packages.md](docs/solutions/2026-05-31-deploy-downgrade-evicts-dependent-packages.md).
+
 ### App Package Versions
 
 Individual apps in `apps/` have their own versions in `metadata.yaml`. These are independent of the VERSION file.
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -10,4 +10,4 @@ pre-commit:
 
     check-changelog:
       run: .github/scripts/check-changelog-edits.sh
-      fail_text: "Direct debian/changelog edits are not allowed. Use './run bumpversion' instead."
+      fail_text: "Direct debian/changelog edits are not allowed. See error above for details."


mairas:
# Code review — PR #216

**Scope:** `523031d..e894c6e`, 1 commit, 2 files.
**Reviewers:** correctness, testing, maintainability, project-standards.

Both P2 findings say the same thing in different registers: a PR whose premise is *the old text was factually wrong* ships text that is also wrong, and leaves the sentence the author actually acts on untouched.

## P2 — Moderate

|---|---|---|---|---|---|
| 1 | `lefthook.yml:13` | **The fix is undone one line later in the same terminal output.** lefthook prints `fail_text` after the failing command's stdout, and it still reads `"Direct debian/changelog edits are not allowed. Use './run bumpversion' instead."` — verbatim the instruction #213 was filed about. The author now reads the corrected explanation, then the uncorrected one, and acts on the last thing they read. One reviewer reproduced the combined output with lefthook 2.1.9 in a scratch repo. The sibling hook on `lefthook.yml:9` already uses the non-duplicating form: "See error above for details." Verified present on the branch. | correctness, project-standards, maintainability, testing | 1.00 | **must fix** |
| 2 | `.github/scripts/check-changelog-edits.sh:31` | **"an edit here changes nothing a device ever sees" is false.** `tools/build-all.sh:25` regenerates the changelog only `if [ ! -f debian/changelog ]`, and the file is tracked, so every local `./run build` stamps the .deb from the committed `0.2.1-1` — and `./run deploy` puts that .deb on a device. This repo's own `docs/solutions/2026-05-31-deploy-downgrade-evicts-dependent-packages.md` records the incident that caused, and its mitigation is a hand edit ("first bump its changelog above the installed version") — the thing the new text tells the author to drop. `run:59` carries a HAZARD comment saying the same. Verified. `AGENTS.md`'s "stale by design" has the same gap. | correctness, project-standards, testing, maintainability | 0.90 | **must fix** |

## P3 — Low

|---|---|---|---|---|---|
| 3 | `check-changelog-edits.sh:36` | The retained bypass line still says "(e.g., fixing a changelog)" — the one activity the new text has just declared pointless, and now the only sanctioned escape hatch. After #2 is fixed it has a real referent again (the local-build downgrade case); until then it contradicts the paragraph above it. | maintainability | 0.78 | fix with #2 |
| 4 | `check-changelog-edits.sh:14` | `# Allow bypass for ./run bumpversion (via env var)` — bumpversion never sets `SKIP_CHANGELOG_CHECK` and never stages a changelog. `git grep` finds the variable only inside this script. The PR rewrote every other stale statement in the file and left the one directly above the code it describes. | correctness, maintainability | 0.85 | fix |
| 5 | `check-changelog-edits.sh:34` | "moves VERSION only" drops a true part of the old message: `.bumpversion.cfg` has `commit = True`, so it also commits. An author following the instruction mid-work gets an unexpected commit. | correctness | 0.70 | fix |
| 6 | `check-changelog-edits.sh:33` | "drop the change" is the only remedy in the message with no command, and it is the case the author is actually in. `git restore --staged --worktree debian/changelog` is two commands they will otherwise half-do. | maintainability | 0.62 | fix |
| 7 | `/Users/…/halos/AGENTS.md:58` (workspace) | The workspace Changelog Policy still states the dch rule unconditionally. It holds for four repos and not for this one, `halos-cockpit-config`, `halos-metapackages` or `halos-marine-containers`. The new AGENTS.md sentence points at the divergence and leaves it standing, so an agent reading top-down still gets the wrong rule first. Out of scope here — file it on `halos-org/halos` with non-closing wording. | project-standards | 0.68 | follow-up |
| 8 | `tests/test-packaging-contract.sh` | Nothing pins the tooling facts the new text names: that `.bumpversion.cfg` has no changelog target, that `generate-changelog.sh` exists, that `build-all.sh` still calls it. Each can silently become false and the hook resumes printing wrong instructions — the exact drift class that produced #213. Three greps in the file that already exists for cross-file invariants. | testing | 0.82 | fix |
| 9 | `debian/changelog` | The deeper option nobody took: untrack the file. `build-all.sh` already generates from `VERSION` when it is absent, which would give local builds `0.7.1-1` instead of the `0.2.1-1` that caused the eviction incident — and would leave the hook and both blocks of documentation with nothing to guard. Blessing the placeholder as "stale by design" makes it permanent instead. | maintainability | 0.60 | follow-up |

## Coverage

- Nothing suppressed by the 0.60 gate; all four reviewers returned.
- Script behaviour is confirmed correct and unchanged: all four paths exercised against the branch (staged changelog → 1, staged non-changelog → 0, empty index → 0, `SKIP_CHANGELOG_CHECK=1` → 0), `bash -n` clean, `nounset` safe. No CI job or caller parses the output — `git grep` finds only `lefthook.yml:12` — so rewording is safe.
- The central claim about CI is accurate: `build-release.yml` prefers the repo-local `generate-changelog.sh` and calls it before the build, and the script does `cat > debian/changelog`. The error is one of scope, not direction.
- The no-VERSION-bump call is right, on firmer ground than the PR states: `VERSION` (0.7.1) already differs from the latest stable tag (`v0.7.0+1`), so the cycle is open regardless of package-affectingness.
- Testing gap worth naming: the hook was verified by running the script directly, which never renders `fail_text`. One real blocked `git commit` with hooks installed would have surfaced finding 1.

## Verdict

**Not ready** — and the reason is instructive: my verification ran the script instead of the commit, so it could not see the half of the output that still carries the defect. Fix 1 and 2, fold in 3–6 while the file is open, then this is a small, correct change.

mairas:
## Review findings addressed

| # | Finding | Resolution |
|---|---|---|
| 1 | P2 — `lefthook.yml` fail_text still prints the wrong remedy | Now "See error above for details.", matching the sibling hook on line 9. Verified through a real blocked `git commit` this time — the path the first round missed by running the script directly |
| 2 | P2 — "changes nothing a device ever sees" is false for local builds | Scoped to "a released package" in both the hook text and AGENTS.md, with a pointer to `docs/solutions/2026-05-31-...`, which is the incident this glossed over |
| 3 | P3 — bypass rationale contradicted the new text | Rewritten to the case that survives: deliberately raising a local build's version so it installs over a released one |
| 4 | P3 — stale header comment about `./run bumpversion` | Now just "Manual override" |
| 5 | P3 — "moves VERSION only" drops that it also commits | "bumps VERSION and commits that change — it does not touch the changelog" |
| 6 | P3 — "drop the change" gives no command | `git restore --staged --worktree debian/changelog` |
| 7 | P3 — workspace policy still states the dch rule unconditionally | Filed as halos-org/halos#137 rather than widening this PR |
| 8 | P3 — nothing pins the tooling claims | Folded into #220 |
| 9 | P3 — untracking the changelog would remove the hazard entirely | Filed as #220. This PR documents the current state; that issue changes it |

The verdict's own point stands and is the reason finding 1 existed: running the script instead of the commit could not see half the output.

